PR #37, opened 2017-10-04
Categories' sidebar

patrykpiston:


src/pages/Category.vue
```diff
@@ -61,7 +61,8 @@ export default {
       category: {},
       filters: { // filters should be set by category, and should be synchronized with magento
         color: ['red', 'blue'],
```
pkarw: Sizes and colors must be mapped to some database-related values. For example "color" in DB is: 169, 180 etc. The same with size.

Can you provide data model for that? separate key / value or label / caption - you name it :) - pairs would be great @patrykpiston

pkarw: I'll add a mechanism to fill out these dictionaries from database

pkarw: This is key to merge this into master :-)

patrykpiston: Sure, once I get some time today / tomorrow I'll get back to it and come up with something :) @pkarw

src/pages/Category.vue
```diff
@@ -60,8 +60,14 @@ export default {
       products: {},
       category: {},
       filters: { // filters should be set by category, and should be synchronized with magento
-        color: ['red', 'blue'],
-        size: ['XS', 'S', 'M', 'L', 'XL'],
+        color: [{'id': 165, 'label': 'red'}, {'id': 166, 'label': 'blue'}],
+        size: [
```
patrykpiston: I've put the key-value pairs here as these are just test data. Or should they be placed in resources/models folder while leaving these fields empty to be filled later?

pkarw: Seems just fine for me. I'll add some db-logic to load the appropriate attribute options in here. It's fine for me

PR #87, opened 2017-10-11
Microcart styling + Filters actions

patrykpiston:
Regarding #5  
I'm short on time but I managed to bind elements with actions. They print out "Switch Filter" for now, but it just depends on how we want to manage filtering. If we want to store active filters in the store and reset it on every exit / entrance to the category page or if we want to do it any other way.


pkarw:
Cool, I'll figure out how to connect this to filter logic.
I think that we'll store the filter values in Category page data not Vuex store because it's category related more than global state

Issue #97, opened 2017-10-12
Change the cursor for add to cart button from default to hand


pkarw:
<img width="484" alt="zrzut ekranu 2017-10-12 06 06 08" src="https://user-images.githubusercontent.com/211899/31478846-81b615ac-af13-11e7-9d9e-79b760c60d93.png">

PR #125, opened 2017-10-17
[#97] Change the cursor for add to cart button from default to hand

mercs600:
closes: #97


filrak:
Thank you very much :)

Issue #44, opened 2017-10-06
Use vuex ui-store instead of global events from event-bus in UI interactions


filrak:
Now we have 3 events for toggling overlay, microcart and sidebar. It would be much better if we move all ui-related common events to one store. There is ui-store.js atm providing overlay, microcart and sidebar interactions but it's not used.

Definition of done: sidebar, overlay and microcart ui interactions are fired from ui-store.js not from global event bus

mercs600:
@patrykpiston are you working on it ? I could refactore evens to ui-store

patrykpiston:
Sure Adam, I could focus on rewriting classes to scss and later RWD :)

mercs600:
I think we can close it.

PR #131, opened 2017-10-18
[#44] Use vuex ui-store instead of global events

mercs600:
Additionally I moved core component functionality from theme to source components

### DIFF
--- a/src/store/modules/ui-store.js
+++ b/src/store/modules/ui-store.js
@@ -1,0 +1,30 @@
+export default {
+  // TO-DO: Make ui-stores inheritance in themes from default ui-store like pages/components
+  state: {
+    sidebar: false,
+    microcart: false,
+    overlay: false
+  },
+  getters: {
+    ui (state) {
+      return {
+        sidebar: state.sidebar,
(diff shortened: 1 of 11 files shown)


src/store/modules/ui-store.js
```diff
@@ -0,0 +1,30 @@
...
+    ui (state) {
+      return {
+        sidebar: state.sidebar,
+        microcart: state.microcart,
+        overlay: state.overlay
+      }
+    }
+  },
+  mutations: {
+    setOverlay (state, action) {
```
filrak: I'm not sure about connecting UI elements like Microcart & Overlay in core.
 Core ui-store should provide the most basic interactions. 
Also i think it should be possible to inherit from this store in `themes/` folder and use theme's ui-store to meet theme's specific requirements (location should be `themes/{theme_name}/store/ui-store.js` so it'll be clear that it's inherited from core `store` folder  )

Can you do this?

mercs600: Sure, so how you want to inherit stores ? by store.registerModule in specific theme ?

Issue #48, opened 2017-10-07
Rewrite css atomic rules to sass mixins


filrak:
The example is in _margin.scss

PR #135, opened 2017-10-19
CSS refactor to SCSS

patrykpiston:
text.css is still to be refactored.

Refers to: #48


src/themes/default/css/border.scss
```diff
@@ -0,0 +1,55 @@
...
+
+$border-px: 1;
+
+// Display
+
+$border-display: none;
+
+// Colors
+
+$border-colors: (
```
filrak: Colors should be imported from colors.scss or you can move all global variables to vars.scss

src/themes/default/css/color.scss
```diff
@@ -0,0 +1,38 @@
+// IMPORTANT: Add new css rules ON DEMAND only when you need them! 
+// Try to use multiples of 5 when using pixels
+
+// Background colors
+
+$background-colors: (
```
filrak: sama as borders, all colors should be referenced from one place

Issue #79, opened 2017-10-11
Sample data for "Get inspired" section on the homepage


karolbzik:
I will prepare a set of data for the section, it bases on photos, usernames and links from Instagram.

karolbzik:
For this section I used the sample data from Magento2, the design is shared in the open source Figma file here - https://www.figma.com/file/VKyqbHFI55TKIKcQlFLiVpVF/Vue-Storefront---Open-Source.

<img width="1440" alt="vue storefront - homepage with the sample data from magento2" src="https://user-images.githubusercontent.com/18116406/31516182-f0db04cc-af97-11e7-9402-23e51c1b1835.png">

Please use "@Instagram" as a sample name for an account on Instagram and link it directly to the website https://www.instagram.com/.

karolbzik:
Here you can find images exported from Figma:
<img width="360" alt="inspiration 6" src="https://user-images.githubusercontent.com/18116406/31516367-a5a63ade-af98-11e7-82c5-380157d59975.png">
<img width="360" alt="inspiration 5" src="https://user-images.githubusercontent.com/18116406/31516369-a620918a-af98-11e7-9bc7-2ed289ee0b09.png">
<img width="360" alt="inspiration 4" src="https://user-images.githubusercontent.com/18116406/31516371-a65efd26-af98-11e7-8f8f-0b6eba7bbebf.png">
<img width="360" alt="inspiration 3 mouseover" src="https://user-images.githubusercontent.com/18116406/31516374-a6912670-af98-11e7-96ea-e51f4a727ade.png">
<img width="360" alt="inspiration 2" src="https://user-images.githubusercontent.com/18116406/31516375-a6c32436-af98-11e7-97bd-c798e7feb514.png">
<img width="359" alt="inspiration 1" src="https://user-images.githubusercontent.com/18116406/31516376-a73642c2-af98-11e7-8c7b-642da5ff50f6.png">

pkarw:
@patrykpiston please fix this PR and let's close this one!

karolbzik:
Padding and alignment in this section need to be improved after changes in the RWD.

@patrykpiston please fix this according to the desired stage from the design.

Current stage:
![16](https://user-images.githubusercontent.com/18116406/32464374-15675b04-c340-11e7-8edc-b254f9e2fc87.png)

Desired stage from the design:
[...]

PR #147, opened 2017-10-23
Created the 'Get inspired' section, for now with slider images

patrykpiston:
Refers to: #79


src/themes/default/store/inspirations.js
```diff
@@ -0,0 +1,63 @@
+// initial state
```
filrak: 1. Is it the same data format as the one provided by Instagram API? If not please change it so it can mock the Instagram API more accurate.
2. Data stores not related to UI should be in `src/store` folder. 
You can pull out the data itself from vuex store and put it in `resource` folder as json file.

src/themes/default/components/theme/blocks/Inspirations/InspirationTile.vue
```diff
@@ -1,36 +1,48 @@
...
     <div class="inspiration-tile">
-        <router-link :to="{ name: 'product', params: { id: product.id, slug: product.slug }}">
-        <div class="product-image bg-lightgray">
-            <img v-bind:src="thumbnail" class="product-thumbnail"/>
-        </div>
-        <p class="mb0">{{ product.name }}</p>
-        {{ product.price | price }}
-        </router-link>
-        
+        <a :href="inspirationBlock.account_link">
```
filrak: I think this component should be reusable and not related with instagram feed block. It's basically an image with overlay which can be used in many ways.

src/themes/default/store/inspirations.js
patrykpiston: @filrak  

1. This and naming the block with anything related to Instagram would be limiting, I think we can fetch images from different social media, piping them through an aggregator of sorts and then saving them in the common store. In this way we could share images from Twitter, IG, FB and more. What do you think?  

2. I'll get to it ASAP

src/themes/default/components/theme/blocks/Inspirations/InspirationTile.vue
patrykpiston: Next on my ASAP list :)

src/themes/default/store/inspirations.js
filrak: 1. What i ment was just mocking it properly. The json from `resource` folder should look like data retrieved from Instagram API. Of course you can use this json in store as you wrote (new one).

patrykpiston:
I had to resolve the conflicts locally, merged it with current version from Divante repo and fixed the conflicts + the cool bags block

Issue #155, opened 2017-10-24
Product Page, Filters are multiplying every time you visit a product


filrak:
After a few rounds homepage -> product it looks like this 

![zrzut ekranu z 2017-10-24 12-47-53](https://user-images.githubusercontent.com/15185752/31939142-c4d51d60-b8b9-11e7-8ab7-39083f323653.png)
@patrykpiston can you fix this?

PR #159, opened 2017-10-24
[#138][#155] Simplified router logic for components with/without mixing and filter/breadcrumbs multiplication fixed

pkarw:
Regarding https://github.com/DivanteLtd/vue-storefront/pull/149. I've simplified the logic to support all use cases


mercs600:
@pkarw I checked it, looks good. I can entry on component without mixins.

pkarw:
Cool. So I'm merging this in.

PR #165, opened 2017-10-24
SSR support for Homepage (#146)

pkarw:
Closes: #146

Without "Inspirations" support - but it's intended (let's say to show that SSR can be used not for all components on the page)

To be honest SSR has downsides as well - the size of JavaScript and HTML is noticeable. We have to figure out how to compress the data outputed as SSR prefetched stores


pkarw:
Important notice: Please clear storage in your browser, as the Homepage stub used to be cached

Issue #157, opened 2017-10-24
Update on microcart doesn’t update the cart model qty 


pkarw:
Label: Easy first pick

mercs600:
@pkarw could you explain a bit more the issue ? I noticed another issue with updating. Described in #168

pkarw:
After your PR it works BUT please add return key support for accepting new value; now - clicking the "Edit" is counter-intuitive

pkarw:
@mercs600

PR #168, opened 2017-10-25
[#157] Microcart / parse updated qty to integer instead of string

mercs600:
I think this may be related with #157. 

When you update qty of product on microcart the number was parse to string. It caused problem for example with counting totals. You can guess what is the result when you add number + string ;-) 

We have 2 solutions. Parsing to Int when qty is updating in  updateQuantity method or just use vue directive. I picked the second one. https://vuejs.org/v2/guide/forms.html#number

We should think about validation and parsing when we are working with v-model and forms.


pkarw:
Seems like a solution! :)

Issue #161, opened 2017-10-24
Logo should link to Home


pkarw:
![zrzut ekranu 2017-10-24 21 23 06](https://user-images.githubusercontent.com/211899/31963602-8b88875c-b901-11e7-8b0a-5e4c707c1364.png)

madonzy:
@pkarw I've added this functionality in https://github.com/DivanteLtd/vue-storefront/pull/176

PR #176, opened 2017-10-26
#161. Set home page url as default when click to logo

madonzy:
Set redirection to the home page when clicking on the logo.


filrak:
Thanks!

Issue #186, opened 2017-10-27
Use the same product image on category and product page - or prefetch product image


pkarw:
The main goal here is: speed + offline support

mercs600:
@pkarw do you mean the same image size ?

pkarw:
Yes, because the resizing proxy is used here to generate the image file. The goal is to have the product image in browser cache when entering product page

PR #211, opened 2017-10-31
[#186] Use the same product image on category/product

mercs600:
Quick fix for categories - using the same image size what product page. 
To achieve prefetching we have to prepare meta tags to handle it.

### DIFF
--- a/src/components/core/ProductTile.vue
+++ b/src/components/core/ProductTile.vue
@@ -14,7 +14,7 @@ export default {
   props: ['product'],
   computed: {
     thumbnail () {
-      return thumbnail(this.product.image, 570, 569)
+      return thumbnail(this.product.image, 310, 300)
     }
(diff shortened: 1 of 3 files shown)


pkarw:
This will slow down products list significantly. Better way is to use ok product page the smaller image (the same as of category page is using) and then dynamically load the original size

Doing so:
- we have worse quality but we have something at least when site is offline (because small images were downloaded along with product list)
- page rendering is fast
- we have high quality image after 1-2s of entering product page

So we have to refactor it to have opponent workflow that you introduced :)

Could you change it?

pkarw:
@mercs600 will you be able to provide the changes till end of the week? We were asked to record the video demonstration to show it on Meet Magento Argentina (16th of November) would be great to have it till then because I would like to show how offline works

PR #262, opened 2017-11-07
Update slider.json

karolbzik:


pkarw:
Carol and the png files? are they uploaded by some other commit?

PR #288, opened 2017-11-11
[#186] Use lazy-load api to change thumbnail on original image

mercs600:
[#186] Use lazy-load api to change thumbnail on original image when user is online.


filrak:
Thanks ;)

Issue #134, opened 2017-10-19
Add static page with sidebar menu


karolbzik:
The design:
<img width="1440" alt="static page with sidebar menu" src="https://user-images.githubusercontent.com/18116406/31780507-3631eb86-b4f6-11e7-92a1-f30cba4f244a.png">

The sample content for this page was taken from a raw Magento2 static page.

The design is also available in the open source graphic files on Figma - https://www.figma.com/file/VKyqbHFI55TKIKcQlFLiVpVF/Vue-Storefront-Open-Source.

pkarw:
@mercs600 maybe this is good fit for you what do you think?

mercs600:
@pkarw I can handle it, but after #44 because I'm working on it now. 

What is scope of this task ? Only markup ? Do we have some api for getting cms content ?

pkarw:
@merc600 Great. Right now we assume that cms content lays in static .md files under /resource. To not complicate things. But we’re open to ideas here?

In the future: I was thinking about integration with Magento CMS, pimcore or WordPress. There are also some futuristic API based cmses, but I think at this point we can’t complicate the tech stack

pkarw:
Please add the links to the footer and we can close this one @mercs600 https://www.dropbox.com/s/scycunny7hz7e0v/Zrzut%20ekranu%202017-11-13%2016.23.18.png?dl=0

mercs600:
https://github.com/DivanteLtd/vue-storefront/pull/293 
@pkarw do we have kidswear category ?

PR #289, opened 2017-11-11
[#134] Add static page with sidebar

mercs600:
1.  I changed styles of static page. 
2. I added possibility to require static page by route props. This is the only one solution which I found. Another solution may be additional route-view with additional routes inside the static page component. 
zrzut ekranu 2017-11-11 o 15 27 26


filrak:
ad 2. - There was a task for it - thanks!

PR #294, opened 2017-11-13
[DO NOT MERGE BEFORE DEMO][NEED TESTING] Tax support beta (very simple rule processing without checking the Co…

pkarw:
…untry/zip-code)


kpodemski:
Isn't an idea of having calculations on "front-end" bad? or i'm missing something?

pkarw:
Good point we’re considering this for a while. Basically in many countries (EU - cross border, US) taxes must be calculated based on current billing address (residence)z this is the reason the logic has been implemented on front end. Second was of course that Vue storefront must work while offline.  To make it safe we should add checksums validation for data provided to the client and then back to avoid any modifications.  That was your point (that’s not safe to have full taxes logic client side)?

kpodemski:
It's not that safe, it's not that accurate, many challenges, you can have many products in cart, some without tax, some with it, you need to handle vouchers and taxing properly, we have VAT MOSS and many regulations behind it. You see, there's a lot of it, a lot of complexity. 

I agree tho that if we want to keep it offline-able and quick we need to go for some compromises. I'll try to dig deeper into this topic and think about possible solutions.

pkarw:
Cool! Would be great to have your feedback on this. This is one of the points where the middleware approach (which makes us totally separated from the eCommerce CMS) has it downsides :/ 

Probably we need open discussion on that matter

pkarw:
Anyway @kpodemski what do you think about adding Prestashop support?

kpodemski:
This is why i'm here, currently i'm at a scout mode, demo is insanely fast, i see a lot of potential in this project

pkarw:
Cool! I'm excited about it. You definitively should take a look at https://github.com/DivanteLtd/mage2nosql which is our Magento2 adapter. It probably needs kind of refactoring because code gets quite complicated during the experiments ;P But it's working and support adapter model for other platforms. Probably can be a good start to add PS support. But to be honest you can as well create your own synchronizer from scratch as separate app etc. The only thing frontend needs is the data that fix current ES schema. In case of any questions just ask, you can use our Slack as well.

Fingers crossed!

pkarw:
TO ALL: please remember to update `vue-storefront-api` and run `npm run migrate` OR `npm run restore` to load all the tax rate rules

Issue #312, opened 2017-11-20
Add default user's country selection based on browser language


pkarw:
https://stackoverflow.com/questions/11826149/is-it-possible-to-determine-the-user-locales-country-as-set-in-the-os-from

pkarw:
https://github.com/DivanteLtd/vue-storefront/pull/311#pullrequestreview-77812774

pkarw:
@Aekal maybe just add the mapping feature we discussed within the PR as for now and when no country is mapped use default from config ?

karolbzik:
@pkarw @Aekal - shouldn't we first add the language versioning (based on language and countries, like here:
![zrzut ekranu 2017-12-05 o 21 55 56](https://user-images.githubusercontent.com/18116406/33630329-41512280-da07-11e7-8e5b-78c84c1f8801.png)

We can add this in the form of pop-up, available through sidebar menu, my account (preferences) and link in the footer. I could add the designs. What do you think?

moezzie:
I just submitted PR #1398 implementing this feature for the checkout and my account pages.

PR #314, opened 2017-11-21
[#312] Add default country selection based on browser language

Aekal:
Default selection is now based on browser language


Aekal:
Sorry, I didn't test it well. Right now I'm not sure that using `navigator.language` is a good idea. 
If user has EN language we actually don't know to what country it should lead - GB, US, maybe another one. User can also lives e.g. in DE and can have other browser language selected (which I think is frequent), e.g. EN and he'll be assigned to wrong country.
So even with mapping to fixed list it'll bring many mistakes.

For now we can close this PR add think about other posibillities.

Issue #198, opened 2017-10-29
Add the login and register pop-ups


karolbzik:
The use scenario: as a customer, I can log in or register from any place in the eCommerce, by clicking the link I will see a pop-up with 2 options - login or register an account.

The design:
<img width="1440" alt="log in pop-up" src="https://user-images.githubusercontent.com/18116406/32146732-7e0fc98a-bcdc-11e7-9a72-db0d395af97e.png">
<img width="1440" alt="register pop-up" src="https://user-images.githubusercontent.com/18116406/32146731-7dff8052-bcdc-11e7-9ffc-49f7ed19b965.png">

The file with the design - https://www.figma.com/file/VKyqbHFI55TKIKcQlFLiVpVF/Vue-Storefront-Open-Source.

I suggest starting with the development of the static pop-ups, preparing it for RWD.
Then we will work on the API so the connection with the logic behind the login/registration could be possible.

pkarw:
https://magento.stackexchange.com/questions/150581/magento-2-how-to-call-rest-api-to-register-a-new-customer

pkarw:
I'm going to take care of this adding login/register api methods to vue-storefront-api as a proxy to Magento api calls

pkarw:
This is done @karolbzik

PR #319, opened 2017-11-27
[#198] Add login and register popups

Aekal:
I created static login and register popups with RWD. 
There are still things to be done, like validation, some logic and connection with backend api, but I think it's a good starting point for next improvements.

For now the popup is temporarily called by clicking on wishlist icon, I wasn't sure where the event should be called.

zrzut ekranu 2017-11-27 o 11 06 08

zrzut ekranu 2017-11-27 o 11 06 19

zrzut ekranu 2017-11-27 o 10 49 22


pkarw:
Cool! Can you add some basic validation using vuelidate (like on Checkout - please take a look there). Merging in for now

Aekal:
Ok, I'll do it later

Issue #217, opened 2017-11-02
Add support for html special chars in titles 


pkarw:
https://www.dropbox.com/s/pkimasuhxzbuyvx/Zrzut%20ekranu%202017-11-02%2020.00.08.png?dl=0

PR #327, opened 2017-11-28
#217: Add support for html special chars in titles

talalus:
https://github.com/DivanteLtd/vue-storefront/issues/217


talalus:
Yes would be better, I though we can trust the service

kacperknapik:
from https://lodash.com/docs/4.17.4#unescape:
> this method converts the HTML entities &amp;, &lt;, &gt;, &quot;, and &#39; in string to their corresponding characters. No other HTML entities are unescaped.

this might not be enough imo

they say this (https://github.com/mathiasbynens/he) should be what we want

Issue #284, opened 2017-11-10
Show the picture of currently selected variant on Product page


pkarw:
It's rather simple `Product.vue:15` when the filterChanged is called. All we need to do is we need to change the images.src after variant is changed to `this.$store.state.product.product_selected_variant`.image  - or rather thumbnail of this image

pkarw:
Thanks to @kacperknapik it's fixed! I'm closing the issue

PR #369, opened 2017-12-05
[#284] Fix product variant images

kacperknapik:


src/lib/filters.js
```diff
@@ -9,7 +9,7 @@ import he from 'he'
  * @param {Int} height
  */
 export function thumbnail (relativeUrl, width, height) {
-  return config.images.baseUrl + parseInt(width) + '/' + parseInt(height) + '/resize' + relativeUrl
+  if (relativeUrl) return config.images.baseUrl + parseInt(width) + '/' + parseInt(height) + '/resize' + relativeUrl
 }
```
pkarw: And what if not? :) it should return sth - for example placeholder URL from the config or just plain string, but please not nothing; it violates LSP (Liskov Substitiution Principle)

src/store/modules/product.js
```diff
@@ -185,30 +180,23 @@ const actions = {
...
-            return false
-          }
-        } else {
-          return false // by some reason configurable_children doesn't have such attribute
-        }
-      }
-
-      return match
+      return Object.keys(configuration).every((configPropeprty) => {
+        return parseInt(child[configPropeprty]) === parseInt(configuration[configPropeprty].id)
```
pkarw: configpropeprty - misspell

src/themes/default/pages/Product.vue
```diff
@@ -15,7 +15,7 @@
           </div>
           <div class="col-md-5">
 
-            <h1 class="mb25 c-black"> {{ product.name | htmlDecode }} </h1>
+            <h1 class="mb25 c-black"> {{ configured_product.name | htmlDecode }} </h1>
```
pkarw: Not sure if this is best option; I mean - probably the expected behavior is that the Product Title is not changing on the product page when selecting variants; maybe let's have product.name here - intentionally

kacperknapik: `product.name` will also be dynamically changed, since we have a `configure` action in store. we should probably store the unchanged product somewhere.

Issue #345, opened 2017-12-01
Product variants as separate products


kacperknapik:
So there is this idea, we could use on the product page.

For the sake of good SEO and easier and correct filtering, we could separate products' variants. Every variant would be a single product. (NOTE: still somehow connected to each other)

Eg.: product 't-shirt foo' has 3 colors, so we create 3 separate products:
't-shirt foo red',
'.. blue'
and '.. green'.

Thanks to that, the filtering is easy, we can choose 'red' color in filters and we get the red t-shirt (and not every product which have 'red' as one of the variants).

pkarw:
OK, cool. We already have all the data requried in product entity (below).

So all we need is to pass the variant_id or sku via URL. To do so we need to change `themes/default/index.js`:

```diff
    { name: 'product', path: '/p/:id/:slug', component: Product },
  +  { name: 'product', path: '/p/:id/:slug/:selected_variant_id', component: Product },
```

Then let's  modify `Product.vue:44` and `store/modules/product.js:110` adding `selected_variant_id` and pass the attribute here and THEN :) modify `store/modules/product.js:125` to make the distinction - selecting not the default variant but this passed via attribute.

Doing so we would be able to index these urls

```json
 "configurable_children": [
            {
              "price": 52,
              "name": "Chaz Kangeroo Hoodie-XS-Black",
              "sku": "MH01-XS-Black",
              "custom_attributes": [
                {
                  "value": "0",
                  "attribute_code": "required_options"
                },
                {
                  "value": "0",
                  "attribute_code": "has_options"
                },
                {
                  "value": "2",
                  "attribute_code": "tax_class_id"
                },
                {
                  "value": [
                    "15",
                    "36",
                    "2"
                  ],
                  "attribute_code": "category_ids"
                },
                {
                  "value": "167",
                  "attribute_code": "size"
                },
                {
                  "value": "49",
                  "attribute_code": "color"
                },
                {
                  "value": "/m/h/mh01-black_main.jpg",
                  "attribute_code": "image"
                },
                {
                  "value": "/m/h/mh01-black_main.jpg",
                  "attribute_code": "small_image"
                },
                {
                  "value": "/m/h/mh01-black_main.jpg",
                  "attribute_code": "thumbnail"
                },
                {
                  "value": "chaz-kangeroo-hoodie-xs-black",
                  "attribute_code": "url_key"
                },
                {
                  "value": "0",
                  "attribute_code": "msrp_display_actual_price_type"
[...]

PR #382, opened 2017-12-07
[#345] Add variants as separate products functionality

kacperknapik:


src/themes/default/components/core/ProductTile.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="product align-center p15">
-    <router-link :to="{ name: 'product', params: { id: product.id, slug: product.slug }}">
+    <router-link :to="{ name: 'product', params: { id: product.id, slug: product.slug, sku: product.configurable_children[0].sku || null }}">
```
pkarw: @kacperknapik it can be misleading the route param should be named like "variant_sku" not SKU because, the key attribute to load product is "id" and product .. has also "sku"; it's misleading, should be named more domain specific

src/components/core/ProductTile.vue
```diff
@@ -21,7 +21,10 @@ export default {
   props: ['product'],
   computed: {
     thumbnail () {
-      return thumbnail(this.product.image, 310, 300)
+      // get first configurable child image
+      // NOTE: might not work when product has no children
+      // todo: play with the image based on category page filters - eg. when 'red' color is chosen, image is going to be 'red'
+      return thumbnail(this.product.configurable_children[0].image, 310, 300)
```
pkarw: It's only an example how ProductTile can look like - OK; but we should provide an condition here; if there are no configurable_children let's use product.image? Why not? @kacperknapik

PR #383, opened 2017-12-07
[-] Fix event bus

kacperknapik:
Fix event bus by replacing `this.$emit...` with `this.$bus.$emit...`


mercs600:
Sorry, I guess I missed it, thanks for fix

Issue #298, opened 2017-11-15
Fix the offline message on the checkout


pkarw:
Each time the user enters the checkout new event is binded and it's the reason why we do have  4 messages instead of one (after I entered the checkout for the 4th time)
![zrzut ekranu 2017-11-15 08 42 24](https://user-images.githubusercontent.com/211899/32824375-0728e47a-c9e1-11e7-8719-72e87c3cb7f9.png)

pkarw:
Fixed by https://github.com/DivanteLtd/vue-storefront/pull/387

PR #387, opened 2017-12-08
Fix the offline message on the checkout

akbarik:
Added destroyed to Checkout page in order to remove network.status listener. This fixes the issue 'Fix the offline message on the checkout #298'

### DIFF
--- a/src/themes/default/pages/Checkout.vue
+++ b/src/themes/default/pages/Checkout.vue
@@ -108,6 +108,9 @@ export default {
       this.activateSection(section)
     })
   },
+  destroyed () {
+    this.$bus.$off('network.status')
+  },
   computed: {
     isValid () {
       let isValid = true


pkarw:
Thx, merging it in. I'm just wondering if the other events there - assigned within "created()" aren't also bound and executed more than once, could you please check it?

akbarik:
Thanks! 
I checked other events in created() of Checkout page. Basically, one other event, namely checkout.placeOrder, is causing exactly the same problem, since it eventually calls checkConnection method too. And so when you press Place an order on Checkout page, same message appears several times again. That definetely needs to be fixed. Sorry, I missed this.
As for other events in created(), they are ok, since they are not causing any trouble, just assigning same values to data, which is safe. But I think it's still a good idea to remove those events on destroyed(), it will be consistent with the above mentioned events and overall will be a good practice.

pkarw:
Ok! Its better to fix them all at once and for all :)

Issue #295, opened 2017-11-14
Add onboarding pop-up on the demo


karolbzik:
As we discussed with @pkarw - we will add the onboarding pop-up on the live demo, visible with the first visit and after clicking the info button on the right bottom corner of the screen.

pkarw:
@mercs600 can you take this one? The designs are on FIGMA

mercs600:
@karolbzik "after clicking the info button on the right bottom corner of the screen." Where can I find the info button which you mentioned ?

karolbzik:
@mercs600 just give me 2 days - I will send you the design for this issue ;)

pkarw:
@mercs600 could you add the final onboarding popup using the popup feature you've recently added to close this one?

karolbzik:
@mercs600 here it is:
<img width="1440" alt="ondoarding widget" src="https://user-images.githubusercontent.com/18116406/33956727-c5174d72-e03f-11e7-9174-65d6d23b3435.png">

In Figma file (https://www.figma.com/file/VKyqbHFI55TKIKcQlFLiVpVF/Vue-Storefront-Open-Source) under the name "Ondoarding widget".

PR #397, opened 2017-12-11
[#295] Add core modal component

mercs600:
I added simple global modal component for all purposes. We can develop it later. But right now: 

zrzut ekranu 2017-12-11 o 22 20 55

Example markup - to get modal design according to figma: 
```html
<modal name="modal-example" class="if-you-need" :width="500">
   <p slot="header">Modal example</p>
   <div slot="content">
      Lorem Ipsum is simply dummy text of the printing and typesetting industry. 
   </div>
</modal>
```

Width value is optional. As default I set 945 according to figma 

If you want to design your own modal just don't use slots:
```html
<modal name="example">
  <div> Lorem Ipsum is simply dummy text of the printing and typesetting industry. </div>
</modal>
```

You can open, hide or toggle by simple events. 
```html
<button @click="$bus.$emit('modal.toggle', 'modal-example')">Example</button>
<button @click="$bus.$emit('modal.show', 'example')">Example</button>
<button @click="$bus.$emit('modal.hide', 'example')">Example</button>
```

PS. this is only first part of #295, I have to add suitable modal + content now 
PS2 if everything goes well, we can rewrite our current modal (auth, subscription) 
PS3 one thing I see to handle: jumping of page when overlay appears -> caused by overflow on #app, we should consider overflow on body only.

### DIFF
--- a/build/webpack.base.config.js
+++ b/build/webpack.base.config.js
@@ -29,7 +29,8 @@ module.exports = {
       theme_pages: path.resolve(__dirname, '../src/themes/default/pages'),
       theme_components: path.resolve(__dirname, '../src/themes/default/components'),
       'theme/components': path.resolve(__dirname, '../src/themes/default/components'),
-      'theme/pages': path.resolve(__dirname, '../src/themes/default/pages')
+      'theme/pages': path.resolve(__dirname, '../src/themes/default/pages'),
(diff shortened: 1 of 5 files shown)


pkarw:
Cool! @mercs600 please add the read me as a doc under /docs :)

Issue #338, opened 2017-11-29
Add AJAX loader to Login and Register screens


pkarw:
Particulary the registration process can take a while, would be great to have a loader for this operation i belive

filrak:
http://tobiasahlin.com/spinkit/ second one looks good ;) 

The task includes creating global loader object with props that allows to show/hide him

pkarw:
Approved :)

karolbzik:
Won't it be a problem to add a progress bar, in the top of the page?
![homepage with sample data with loading bar](https://user-images.githubusercontent.com/18116406/33551621-40b0e5d8-d8f2-11e7-86ac-ff335d163a36.png)
In the design I used #333 color and 2px height.

pkarw:
@collymore how does the progress look like here?

BlitZ-UK:
#410

pkarw:
Closed with #410 and #412

PR #410, opened 2017-12-13
#338 - Ajax Loader

BlitZ-UK:
Usage: 
Activate using `this.$store.commit('ui/setLoader', true)`

### DIFF
--- a/src/components/core/Loader.vue
+++ b/src/components/core/Loader.vue
@@ -25,6 +25,10 @@ export default {
     hide () {
       this.$store.commit('ui/setLoader', false)
     }
+  },
+  mounted: function () {
+    this.$bus.$on('notification-progress-start', this.show)
+    this.$bus.$on('notification-progress-stop', this.hide)
   }
 }
 </script>
--- a/src/components/core/Loader.vue
+++ b/src/components/core/Loader.vue
@@ -2,10 +2,20 @@
 
   <div class="loader-container" v-if="isVisible">
 
-      <div class="spinner">
-      <div class="double-bounce1"></div>
-      <div class="double-bounce2"></div>
-  </div>
+      <div class="loader-inner-container">
+
+          <div class="spinner">
+              <div class="double-bounce1"></div>
+              <div class="double-bounce2"></div>
+          </div>
+
+          <div class="loader-message-container" v-if="message">
+              {{ message }}
+          </div>
+
+      </div>
+
+
   </div>
 
 </template>
@@ -15,11 +25,17 @@ import { mapState } from 'vuex'
 
 export default {
   name: 'loader',
+  data: function () {
+    return {
+      message: null
+    }
+  },
   computed: mapState({
     isVisible: state => state.ui.loader
   }),
   methods: {
-    show () {
+    show (message = null) {
+      this.message = message
       this.$store.commit('ui/setLoader', true)
     },
     hide () {
@@ -27,7 +43,9 @@ export default {
     }
   },
   mounted: function () {
-    this.$bus.$on('notification-progress-start', this.show)
+    this.$bus.$on('notification-progress-start', (message) => {
+      this.show(message)
+    })
     this.$bus.$on('notification-progress-stop', this.hide)
   }
 }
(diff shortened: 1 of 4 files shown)


pkarw:
Could you please provide the event for loaders (it's a kind of convention right now to handle all kind of notifications using events for example `EventBus.$emit('notification')` etc.

So would be great to have event like: `EventBus.$emit('notification-progress-start')` and `EventBus.$emit('notification-progress-stop')` for this.

Then, we can add `EventBus.$emit('notification-progress-step', { progress: 50, of: 100, message: 'Loading user profile ...' })` in the next step

@BlitZ-UK @collymore

pkarw:
After adding events support - could you please add the `notification-progress-start` and`notification-progress-stop` around:
- all fetch() calls you can find in the app
- ... for sure please add it to: store/modules/user.js - login, password reset etc,
- entering Product.vue and Category.vue pages - you probably need to emit notifiation-progress-stop inside `fetchData()` methods 

for example:

```js
import EventBus from 'src/event-bus'
  /**
   * Send password reset link for specific e-mail
   */
  resetPassword (context, { email }) {
    console.log({ email: email })
    EventBus.$emit('notification-progress-start') 

    return fetch(config.users.endpoint + '/resetPassword', {
      method: 'POST',
      mode: 'cors',
      headers: {
        'Accept': 'application/json, text/plain, */*',
        'Content-Type': 'application/json'
      },
      body: JSON.stringify({ email: email })
    }).then((response) => {  
    EventBus.$emit('notification-progress-stop')
      const contentType = response.headers.get('content-type')
      if (contentType && contentType.includes('application/json')) {
        return response.json()
      } else {
        console.error('Error with response - bad content-type!')
      }

      return response
    })
  },
```

filrak:
@BlitZ-UK  `notification-progress-start` and `notification-progress-stop` with message under loading bar as `notification-progress-start` event param (e.g. ` $bus.$emit('notification-progress-start', "Loading user profile, please wait")`) would be awesome. You can also do it in Vuex store like overlay, both ways are correct ;)

BlitZ-UK:
Loader can now contain a message using the format mentioned above :)

vue-storefront-loading

pkarw:
Thanks guys, it's cool. But how does it regard the UI design from https://github.com/DivanteLtd/vue-storefront/issues/338 - a top bar. Could you implement the second version of loader regarding the UI designs ?:)

Very simple, linear loader like on Youtube etc. 

Would be great to choose the loader by specific parameter of notification / ui store.

If we won't add it like on design, our designer @karolbzik will be back with some issues I belive ;)

BlitZ-UK:
@pkarw At the moment the loader is activated via .start/.stop commands, how will we translate this into an animated progress bar ? Until it has percentage base or progress steps it should wait until then. I chose one of the loaders that was from spinkit/from #338 too.

I'd wait off until we wrap the ajax requests around this loader (or alternatively rather than manually going through, why don't we use something like Pace and automatically catch all ajax requests and display the top loader that way (see http://github.hubspot.com/pace/docs/welcome/)).

Thanks

pkarw:
@BlitZ-UK point taken :) Let's leave it as it is.
My initial idea was to use sth like http://github.hubspot.com/pace/docs/welcome/ - it's cool.

Can you integrate the http://github.hubspot.com/pace/docs/welcome/ and check how it works with our resource loading model? I belive it will be jsut fine. 

Sorry for this long comments topic I didn't mean to use a negotiation-nibble tactic for this feature ;P

pkarw:
Plus! Could you add the documentation about how to use the - let say 'manual progress bar' - under: https://github.com/DivanteLtd/vue-storefront/tree/master/doc ?

pkarw:
I'm merging this in, however it's still great to have this linear-loader (pace) implemented, could you take care of this guys?

BlitZ-UK:
Yeah we'll take care of this and integrate Pace (top loader) with all Ajax requests, we've recently done this with one of our other Vue projects, it takes a little patch that we wrote but after that Pace works fine with it. I'll get the docs updated with how to use the manual loader for now too,

Thanks !

PR #462, opened 2017-12-27
Logging in or registering at Checkout

akbarik:
Added possibility to log in or register at Checkout page by filling in password in personal details. 
The design was taken from Figma project. 
According to issue https://github.com/DivanteLtd/vue-storefront/issues/221
Here are screenshots of how it looks now:
screenshot from 2017-12-27 20-39-27
screenshot from 2017-12-27 20-39-44


src/client-entry.js
```diff
@@ -208,3 +208,6 @@ function checkiIsOnline () {
 window.addEventListener('online', checkiIsOnline)
 window.addEventListener('offline', checkiIsOnline)
 
+EventBus.$on('user.loggedin', (receivedData) => {
```
pkarw: Please change the event name to be compliant with the convention: module-after|before-eventname so in that case it should be: user-after-loggedin

src/store/modules/user.js
```diff
@@ -136,6 +137,12 @@ const actions = {
         .then((resp) => {
           if (resp.code === 200) {
             context.commit(types.USER_INFO_LOADED, resp.result) // this also stores the current user to localForage
+
+            EventBus.$emit('user.loggedin', {
```
pkarw: Please correct the event name (as suggested in the previous comment)

src/themes/default/components/core/blocks/Checkout/OrderReview.vue
```diff
@@ -61,7 +67,41 @@ export default {
...
+            type: 'error',
+            message: result.result,
+            action1: { label: 'OK', action: 'close' }
+          })
+        } else {
+          this.$bus.$emit('notification', {
+            type: 'success',
+            message: 'You are logged in!',
+            action1: { label: 'OK', action: 'close' }
+          })
```
pkarw: We should add the user.id to the order object - it's not implemented yet to pass the non-guest orders to the Magento backend yet; i belive we should add next section to the order object like `user` with `id` property. The whole section shouldn't be required - please modify the models/order.schema.json accordingly (and also update the `vue-storefront-api` copy of the schema)

src/themes/default/pages/Checkout.vue
```diff
@@ -123,11 +123,20 @@ export default {
     isValid () {
       let isValid = true
       for (let child of this.$children) {
-        console.log(child)
+        console.log('Child: ', child)
```
pkarw: Please remove the verbose logging here as it's probably not required beyond the dev. process

akbarik:
Done

PR #466, opened 2017-12-29
Filters duplication on Product page

akbarik:
Fixed bug with more than one click on Product tile and showing duplicate filters on Product page (Issue https://github.com/DivanteLtd/vue-storefront/issues/461)

### DIFF
--- a/src/themes/default/components/core/ProductTile.vue
+++ b/src/themes/default/components/core/ProductTile.vue
@@ -1,19 +1,21 @@
 <template>
   <div class="product align-center p15">
-    <router-link :to="{ name: product.type_id + '-product', params: { parentSku: product.parentSku ? product.parentSku : product.sku, slug: product.slug, childSku: product.sku }}">
-      <div class="product-image">
-        <transition name="fade" appear>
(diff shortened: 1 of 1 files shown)


mercs600:
Generally I think we should fix this bug with duplicated filters on product page instead of prevent double clicking on product link. I was trying reproduce it, but I couldn't. @akbarik are you able to send me some video how it looks like ?

akbarik:
The duplication because of the way router-link works and SSR. When you click on product tile, a route specified in router-link is pushed to global $router object. So the more you click the more duplicate routes get pushed there. After that each one of them is executed by launching prefetch promises defined in asyncData methods of Product component. So there the duplication occurs. If you try to put console.log in asyncData method of Product page, you'll see that after clicking on product tile several times, it outputs several debugs in console. 
I haven't tested @click.once, I'll do that right now and get back with the result.

akbarik:
Unfortunately, couldn't get @click.once to work.
I agree with you guys that the solution is not elegant, and I didn't want do to it this way either. It's just that this was the only working solution, at leat of that I knew :) 
 Please help me out here, if you know how to make it better.

PR #473, opened 2017-12-31
Clean focus on click 

mercs600:
1. We should not remove the default focus from the buttons, because then we lose accessibility.
Specially when we don't design our focus state. So for situations like these I prepared special directive to clean focus state only when user click on button by mouse. When user is clicking by tab key - focus is visible. 

Basic usage: 
```html
<button v-focus-clean>Button</button>
```
When user is clicking on button by mouse I execute blur event to lose focus. When user is clicking by tab behavior is like default. 

But what if our animation is working on focus state ? We can't execute blur event, because we need focus. So then we can delete just visually part of focus by: 
```html
<button v-focus-clean="{class: 'no-outline'}">
```
Then when user is clicking by mouse we are applying special class to remove visually part of focus. 

2. Referring to non-native buttons I'm wondering why they are non-native ? Why we can't just use button ? Right now I added tabindex attribute to add focus possibility on div, but from my perspective we should keep just button element to keep all functionalities like submit on enter keypress etc. @filrak what do you think ? 

3. We didn't have directives yet. I added it here src/themes/default/components/theme/directives But it is good place ? @filrak what do you think ?

### DIFF
--- a/src/themes/default/components/theme/directives/focusClean.js
+++ b/src/themes/default/components/theme/directives/focusClean.js
@@ -1,0 +1,32 @@
+import Vue from 'vue'
+
+export default Vue.directive('focus-clean', {
+  inserted: function (el, binding) {
+    let mouseDown = false
+    let css = binding.value && binding.value.class ? binding.value.class : false
+    let timeout = binding.value && binding.value.delay ? binding.value.delay : 1000
+
+    el.addEventListener('mousedown', () => {
+      mouseDown = true
+    })
+
+    el.addEventListener('mouseup', (event) => {
+      mouseDown = false
+
+      css && setTimeout(() => {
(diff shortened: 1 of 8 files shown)


pkarw:
@filrak should accept this one

src/themes/default/components/theme/directives/focusClean.js
```diff
@@ -0,0 +1,32 @@
...
+      if (mouseDown) {
+        if (css) {
+          event.target.classList.add(css)
+        } else {
+          event.target.blur()
+        }
+      }
+    })
+  }
+})
```
pkarw: For me it makes sense to have the directives there. As for now we do have only one or two directives inside ‘lib/filters.js’

PR #495, opened 2018-01-06
[#8] Add nested categories support

mercs600:
New behavior you can see on video:
https://www.dropbox.com/s/tq9gprjxjhg05t7/nested_categories.mov?dl=0 

PS. We have some problem In categories model, categories with 4 level doesn't have slugs.

### DIFF
--- a/src/themes/default/components/core/blocks/SidebarMenu/SidebarMenu.vue
+++ b/src/themes/default/components/core/blocks/SidebarMenu/SidebarMenu.vue
@@ -1,28 +1,26 @@
 <template>
     <div class="sidebar-menu bg-lightgray" :class="{ active: isOpen }">
-        <div class="row">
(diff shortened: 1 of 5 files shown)


pkarw:
Wow! @mercs600 its extremely cool!
Slugs are assigned within the category vuex store, it seems like the code has some limitation regarding recurrence level. I can take a look at it on

hafeez3000:
I see limitation after 2 levels, need to re-look into the code.  But feature will be super cool after fixing the issue.

pkarw:
Ok, we need the recurrency in here https://github.com/DivanteLtd/vue-storefront/blob/4beccafd98b072ba1ae1b697f6074866de24c75e/src/store/modules/category.js#L149 :)

hafeez3000:
@pkarw: That is correct and now no issues with slug and no problem with nested category navigation

Issue #494, opened 2018-01-06
Adding Google Analytics feature


hafeez3000:
Hi, is something possible to use plugin https://github.com/MatteoGabriele/vue-analytics for implementing Google Analytics?

I found documentation here https://matteogabriele.gitbooks.io/vue-analytics/content/

pkarw:
Seems like a pretty straightforward solution and we should implement it! Thanks @hafeez300!
Are you willing to try to make it work?

pkarw:
@hafeez3000 how are things in here?

pkarw:
@hafeez3000 how are things?

pkarw:
@hafeez3000 how about the improvements to this one?

PR #496, opened 2018-01-07
Adding Google Analytics feature #494

hafeez3000:
1) Added https://github.com/MatteoGabriele/vue-analytics in packages.json
2) Defined Google Analytics UA ID to src/config.example.json
3) Initiated Vue-analytics in src/app.js with route option

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,8 @@
     "vuelidate": "^0.6.1",
     "vuex": "^3.0.0",
     "vuex-router-sync": "^5.0.0",
-    "webpack-plugin-critical": "^1.0.0"
+    "webpack-plugin-critical": "^1.0.0",
+    "vue-analytics": "^5.8.0"
   },
   "devDependencies": {
     "babel-core": "^6.13.2",
--- a/src/app.js
+++ b/src/app.js
@@ -19,6 +19,11 @@ Vue.use(VueLazyload, {
   attempt: 2
 })
 
+Vue.use(VueAnalytics, {
+  id: config.googleanalytics.id,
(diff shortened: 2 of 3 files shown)


pkarw:
@hafeez3000 thanks! Still few things to improve:
- if no google account Id provided, the Vue analytics shouldn’t have been instantiated - let users to decide by setting this option is
- I would propose to add support for this library as a plugin / check the /extensions/mailchimp
- we should support eCommerce events by binding specific event (probably need to be added - event after order is made: order-after-placed) - then bind to such event in your plugin and use https://github.com/MatteoGabriele/vue-analytics/blob/master/docs/ecommerce.md 

The necessity of moving this to the plugin is implied by the assumption of having external services(mailchimp, ga) in plugins, not in the core

Please also check how can we support offline google analytics? This is PWA so we cannot assume that user has internet connection all the time

For reference https://www.npmjs.com/package/sw-offline-google-analytics

https://developers.google.com/web/updates/2016/07/offline-google-analytics

As it says- using this solution you should have out of the box offline support because the mentioned library redo all the requests made to GA automatically 

But please test if it’s fine

Aftter you made the changes mentioned above I’ll be able to merge this feature in

GA is crucial so we have to have it done the right way :)

hafeez3000:
@pkarw: Thank you for your feedback, completely agree with your comments, however this may need sometime to implement.

pkarw:
No worries, please take your time!

pkarw:
Closing this one  because of https://github.com/DivanteLtd/vue-storefront/pull/578

Issue #485, opened 2018-01-03
Adjustments on the category page on mobile


karolbzik:
Apply adjustments and improvements to make the view closer to the design ("Category on mobile" in https://www.figma.com/file/VKyqbHFI55TKIKcQlFLiVpVF/Vue-Storefront-Open-Source).

The comparison, left - current stage, right - the design.
![1](https://user-images.githubusercontent.com/18116406/34533060-c8727116-f0b8-11e7-9cd5-b620c1e13951.png)

The most important adjustments:
- Adjustments in sizes and colors of fonts;
- Add gray background to the product's photo;
- Adjustments in margins and paddings;
- Add transparency to the photo of the product (like in the design, 20%);
- Add mouseover effect (changes in colors of fonts and background, changes in the transparency of the product's photo - like in the design) ;
![zrzut ekranu 2018-01-03 o 19 14 42](https://user-images.githubusercontent.com/18116406/34533429-70acc97a-f0ba-11e7-81c9-a9ae65beb343.png);
- Removal of the top for filters active in mobile;
![image](https://user-images.githubusercontent.com/18116406/34533343-26392e2e-f0ba-11e7-8854-eaa9ee2f2b57.png).

The most important improvements (nice to have):
- Add two columns to the list of the products;
- Add "Sort by" feature;
- Add the number of items;
- Add "go to the top" button;
- Add "Search" button" and "Favorites" to the top (improvement of UX of these features).

Some adjustments may be related to https://github.com/DivanteLtd/vue-storefront/issues/456.

pkarw:
@Aekal good work seems like th task is ready to be closed

PR #498, opened 2018-01-08
[#485] Adjustments on the category page on mobile

Aekal:
I did some things from #485:
* Adjustments in sizes and colors of fonts
* Added gray background to products photos with opacity and mouseover effect
* Moved filters above the top bar
* Added products counter
* Added two columns on product list
  
zrzut ekranu 2018-01-08 o 22 46 55


pkarw:
Thanks @Aekal, Good job! Can you Just mark the fixed bulletpoints in the original issue?

Issue #500, opened 2018-01-10
[BUG] Horizontal scroll on mobile shows empty space


rossmc:
On mobile if the user scrolls/swipes right they will see an empty space to the right of the content.  This is very obvious on the main slider or a main product image but occurs everywhere.

![horizontal-scroll-issue2](https://user-images.githubusercontent.com/2452991/34764946-ee159e96-f5e7-11e7-8dc4-bbc99415ab46.jpg)

We can fix this by adding the following CSS
```
html, 
body {
    overflow-x: hidden;
}
```

I suggest we create a new file to add this global css to, e.g. `src/themes/default/css/layout.scss`.  Although the [documentation](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/Working%20with%20stylesheets%20(CSS).md#extending-scss-files) says we should not create any new files in this directory?

pkarw:
@rossmc good point; please proceed

filrak:
@rossmc Good idea, feel free to fix it! The styles modifying `body` element are in `App.vue`. now but it's a good idea to move them to some external file like `layout.scss` you mentioned ;)

pakiam:
@rossmc very bad practice, cuz its a hack. 
Problem is more deep. In grid.

rossmc:
@pakiam really? Any specific case where it would stand out as bad practice?

I like it as it's very clear in what it's doing and see where it's coming from. Can't think of many cases where you'd want horizontal scrolling on the body element.

Although I get your point, it's good not to have the underlying causes in the first place.  Though I don't think you should edit the [flexbox-grid.css](https://github.com/DivanteLtd/vue-storefront/blob/master/src/themes/default/css/flexbox-grid.css) directly in [your PR](https://github.com/DivanteLtd/vue-storefront/pull/501), it's from [flexboxgrid.com](http://flexboxgrid.com/), and incase it was ever replaced with an updated version, we'd loose your changes.

rossmc:
@pakiam interestingly there are some issues related to this reported on [flexbox-grid's github](https://github.com/kristoferjoseph/flexboxgrid/issues?utf8=%E2%9C%93&q=is%3Aissue+horizontal), I found these issues/comments interesting:
- [Issue 177 - Core Dev's Comment](https://github.com/kristoferjoseph/flexboxgrid/issues/177#issuecomment-239568734)
- [Issue 51](https://github.com/kristoferjoseph/flexboxgrid/issues/51)

[...]

PR #556, opened 2018-01-19
[#500] / [#539PR] Bugfix to horizontal scrollbar and router behavior

pakiam:
1. Remove flexbox-grid.css 
2. Add flexboxgrid2 as a dev dependency 
3. Add flexboxgrid in App.vue 
4. Change col-md-12 to col-xs-12 because of lack of styles for col-md-* for phone media queries on Home page and its components 
5. Add adaptivity for New Luma collection Carousel 
6. Remove padding-y for .tile-container on mobile media query


build/config.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"elasticsearch":{"httpAuth":"","host":"localhost:8080/api/catalog","index":"vue_storefront_catalog"},"cart":{"synchronize":true,"create_endpoint":"http://localhost:8080/api/cart/create?token={{token}}","updateitem_endpoint":"http://localhost:8080/api/cart/update?token={{token}}&cartId={{cartId}}","deleteitem_endpoint":"http://localhost:8080/api/cart/delete?token={{token}}&cartId={{cartId}}","pull_endpoint":"http://localhost:8080/api/cart/pull?token={{token}}&cartId={{cartId}}"},"orders":{"endpoint":"http://localhost:8080/api/order"},"users":{"endpoint":"http://localhost:8080/api/user"},"stock":{"endpoint":"http://localhost:8080/api/stock"},"images":{"baseUrl":"https://demo.vuestorefront.io/img/"},"install":{"is_local_backend":true,"backend_dir":"../vue-storefront-api"},"demomode":true,"tax":{"defaultCountry":"PL","defaultRegion":"","calculateServerSide":false},"i18n":{"defaultCountry":"US","defaultLanguage":"EN"},"mailchimp":{"endpoint":"http://localhost:8080/api/ext/mailchimp-subscribe/subscribe"},"registeredExtensions":["custom_extension","mailchimp-subscribe"]}
```
pkarw: Please add ignore and remove this file from this PR as it is automatically generated config dump

src/themes/default/App.vue
```diff
@@ -71,7 +71,8 @@ export default {
 </script>
 
 <!-- CSS DOCS: https://github.com/DivanteLtd/vue-storefront/blob/master/doc/Working%20with%20stylesheets%20(CSS).md -->
-<style src="./css/flexbox-grid.css"></style>
+<style src="../../../node_modules/flexboxgrid2/flexboxgrid2.min.css"></style>
```
pkarw: We should have used the web pack alias in that case or other way to keep the css unaware of physical path structure @pakiam @filrak @mercs600

pakiam: @pkarw added to vue-loader.config option IncludePath: '/node_modules/'

build/config.json
pakiam: @pkarw Add this file to .gitignore, Is this that you mean ?

pkarw: Yes, thank you

mercs600:
@pakiam thanks for this improvement, finally we will be able to avoid horizontal scroll. I see some small issues/differences to fix before we merge it. Homepage works good, but if we want replace grid library, we must to take care about all views. 

1. Something wrong with icons in mobile version of header. I think we have to adjust some col-* class. 
zrzut ekranu 2018-01-20 o 16 04 22
2. Rows/cols of grid on product page should be fixed: 
zrzut ekranu 2018-01-20 o 16 05 43
3. Product list on category page. On md view we have 2 products in row, not 3. 
zrzut ekranu 2018-01-20 o 16 08 27
4. Links in footer
zrzut ekranu 2018-01-20 o 16 13 19

If I will find some more I will let you know.

pkarw:
@mercs600 thanks, please do test all the pages @pakiam and @mercs600 before we merge this to the core, as we're releasing new version on 29th of January - so, the date is very close and the changes quite radical as I see :) @filrak

mercs600:
@pakiam are we able to add some webpack alias to node_modules ? Right now It might be a bit confusing, when some user see just 
```html
<style src="flexboxgrid2" lang="scss"></style>
```

@filrak what do you think ?

mercs600:
@pakiam I have some issues. I switched on your latest commit and I can't see your fixes. Please look on video below: 
https://www.dropbox.com/s/0t7s562ckcwahcl/grid-merge.mov?dl=0

pakiam:
@mercs600 yes, I see. Problem with header comes while I was merging my branch with master.
Thx for helping. Also update npm's to update flexbox2grid to latest version.

src/themes/default/components/core/blocks/Header/Header.vue
```diff
@@ -3,23 +3,23 @@
     <header class="brdr-bottom bg-white brdr-c-alto"  :class="{ 'is-visible': navVisible }">
       <div class="container">
         <div class="row between-xs middle-xs px15" v-if="!isCheckout">
-          <div class="col-sm-4 col-xs-2 middle-xs">
+          <div class="col-xs-2 col-sm-2 middle-xs">
```
mercs600: @pakiam are you sure to declare the same column on each resolution ? I mean when you declared 2 columns on xs, you don't have to repeat it on sm if you don't need any change

src/themes/default/components/core/blocks/Footer/Footer.vue
```diff
@@ -170,6 +170,9 @@ $gray-secondary: map-get($colors, gray-secondary);
       margin-left: 20px;
     }
   }
+  a {
```
mercs600: Please use .no-underline class instead of this

mercs600:
@pakiam Thanks I see some fixes right now, but we have some issues still. I mean 2 (scroll on mobile product page and 3 - 3 products in row instead of 2. Could you check this please ?

src/themes/default/components/core/blocks/Footer/Footer.vue
pakiam: I dont understand... It should be underlined :) In previous commits i deleted this accidentally, cuz thought it's a bug of double underlining on hover.
https://github.com/DivanteLtd/vue-storefront/blob/master/src/themes/default/components/core/blocks/Footer/Footer.vue#L172

mercs600: ahh sorry man ;-)

pakiam: image
Cant see  2.
Same for 3.
image

src/themes/default/components/core/blocks/Header/Header.vue
pakiam: @mercs600 Well, I think it`s helps read code easily, plus you can count cols faster. If you want I can remove repeating. But when you watch at this you can 100% sure which columns disappear on -md- for example, and which stays on -sm- without thinking.
PS This is how Header looks like on master branch 
image

src/themes/default/components/core/blocks/Footer/Footer.vue
mercs600: Right now product description works good for me, I don’t know why I couldn’t see your previous changes. 
According to point 3, please compare your local version to test version. Problem is on some resolutions about 1000px width. On our current version we have 2 products in row, on your version is there 3 products in row. Please compare also similiar products on product page on lower resolutions. This is big change so we have a lot of change requests :-)

src/themes/default/components/core/blocks/Header/Header.vue
mercs600: @pakiam I now what you mean, but in my opinion and from code perspective this is totally redundant. Grid frameworks are being builded mobile first, so when you see only col-xs-2 you should know at the same time you have 2 col on other resolution. Please remove repeating ;-)

filrak:
@mercs600 i think webpack aliases shouldn't be theme-related

src/themes/default/components/core/ProductListing.vue
```diff
@@ -11,7 +11,7 @@ import { coreComponent } from 'lib/themes'
 import ProductTile from './ProductTile.vue'
 
 export default {
-  props: ['products', 'columns'],
```
mercs600: We shouldn't remove this props, because we are using this component on other pages like home and product pages with different settings.

src/themes/default/components/core/blocks/Header/Header.vue
```diff
@@ -172,20 +172,33 @@ export default {
...
-    .col-xs-2:first-of-type {
-        padding-left: 0;
+    @media (max-width: 374px) {
+      padding-right: 15px;  
     }
-    .col-xs-2:last-of-type {
-        padding-right: 15px;
+  }
+
+  .col-xs-2:first-of-type {
```
mercs600: We should not styling by grid classes. We should add some additional class.

pakiam: https://github.com/DivanteLtd/vue-storefront/blob/master/src/themes/default/components/core/blocks/Header/Header.vue#L183
I just fix this to fit it in new grid.
Should I rewrite this normally in this PR?

pakiam: @mercs600 anyway, i added some classes

src/themes/default/components/core/ProductListing.vue
pakiam: @mercs600 the strange thing is that mixins (also i cant understand why not extends) merging props from "base component" which is core/ProductListing and i dont see any reason to doubleing them.
And the second one, check my little refactoring of this component(base)

mercs600:
@pakiam why did you close this PR ?

PR #600, opened 2018-01-27
prevented scrolling to top, when overlay is active

PanMisza:
It's not enough to set `overflow: hidden;` on `<div id ="app">` - we must set it on body or html to prevent scrolling.

### DIFF
--- a/src/components/core/Overlay.vue
+++ b/src/components/core/Overlay.vue
@@ -10,6 +10,12 @@ export default {
     onClick () {
       this.$store.commit('ui/setOverlay', false)
     }
+  },
+  beforeCreate () {
+    document.documentElement.classList.add('no-scroll')
+  },
+  destroyed () {
+    document.documentElement.classList.remove('no-scroll')
   }
 }
 </script>
--- a/src/components/core/blocks/Auth/SignUp.vue
+++ b/src/components/core/blocks/Auth/SignUp.vue
@@ -18,6 +18,9 @@ export default {
     Register,
     ForgotPass
   },
+  mounted () {
+    this.$store.commit('ui/setSignUp', true)
+  },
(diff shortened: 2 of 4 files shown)


pkarw:
Thanks!

szafran89:
@PanMisza Is this issue related only with `Sign Up` modal?

On stage and my local I can't see this issue but I think it can be if `Sign Up` modal is too high and is outside of screen (depends on resolution). Basically there is no need to set `overflow` for document element. All what we need is to use our `Modal (core/Modal.vue) component` in `SignUp` component. IMO it should resolve the issue.

PanMisza:
It fixes scrolling to top, when you open cart or something else that uses overlay. 
CSS changes in SignUp make this modal scrollable in landscape mode on mobile devices.

szafran89:
hmm weird, could you tell me which browser/system you use? Any video/gif would be helpful. @filrak @mercs600 did you spot the issue?

PanMisza:
I'm on stable Chrome 64 (Mac). The same problem is in Firefox, Chrome on Android, Safari...

1. Scrolling bug:
[before](https://i.imgur.com/4DRAmGt.mp4) --> [after](https://i.imgur.com/7zltKx2.mp4)

2. `SignUp.vue` modal bug:
[before](https://i.imgur.com/hqZ2zq2.png) (scrolls with page, without black overlay if opened from sidebar) --> [after](https://i.imgur.com/0svXm95.png) (only modal is scrollable)

szafran89:
Thanks for demo.

I get it now. You right there isn't another way to prevent scrolling.
For me code looks ok. 

One improvement. It would be fine to replace current SignUp modal with our `Modal` component but it can be in next PR.

PR #601, opened 2018-01-28
Fix click and hover area on account icon

Aekal:
For now the click and hover area on account icon is limited to the icon inside a button. So, if you don't click exactly in the icon the login popup won't show up.
zrzut ekranu 2018-01-28 o 02 10 28
I've made a small fix to make the whole button clickable (same as other buttons).
I also removed bugged `setTimeout` on mouseout.

### DIFF
--- a/src/themes/default/components/core/blocks/Header/AccountIcon.vue
+++ b/src/themes/default/components/core/blocks/Header/AccountIcon.vue
@@ -1,12 +1,17 @@
 <template>
-  <div class="account-icon inline-flex" @click="openMyAccount">
-    <div class="dropdown" @mouseover="hoverIcon">
-      <i class="material-icons md-18" @click="gotoAccount">account_circle</i>
-      <div v-if="currentUser" :class="showDropdown ? 'dropdown-content show-dropdown' : 'dropdown-content'">
+  <div
+    class="account-icon inline-flex"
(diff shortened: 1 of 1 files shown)


akbarik:
@Aekal, thanks for fixing this icon. Now it really works much better. That setTimeout was like a workaround and I didn't like it either :) So, great job!
One note, though.You've put a call to gotoAccount in outer div, which now allows user to get into his My Account page just by clicking on the icon itselt. But in a dropdown there's a separate My Account link for that. So I believe it's better to leave a call to gotoAccount in a router-link, rather than in outer div. What if user doesn't want to open his My Account page, but just wants to logout?

Aekal:
@akbarik I think we need a call to this method on button, because on mobile we should be able to click and move to My Account page (dropdown doesn't appear there). 
The `openMyAccount` method works only on mobile so this action is not duplicated on desktop.
On desktop the dropdown appears on mouseover and user can simply click on logout link and he won't be redirected to My Account (which works fine I think :))

A small issue may appear when you load a page on mobile and resize to desktop size (using devtools) and then `openMyAccount` method will still work. But also in this case everything works fine, there will be only one extra click event to open My Account page, but it doesn't break anything.

akbarik:
Oh, ok, makes sense :) thx!

PR #603, opened 2018-01-29
Add hover and focus styles to filter buttons

Aekal:
- Added hover, focus and aria-label to filter buttons for support keyboard navigation and accessibility
- Some code refactoring
zrzut ekranu 2018-01-29 o 12 55 22
zrzut ekranu 2018-01-29 o 12 55 12
zrzut ekranu 2018-01-29 o 12 55 33


filrak:
Good job, thanks ;)

PR #619, opened 2018-02-03
Moved back button to far left

collymore:
Just a small update to move the slide out menu's back button to the far left. It feels more intuitive and conventional being on the left rather then in the center.


filrak:
@collymore  Seems like a good idea but please provide a screenshot of how it looks after your changes ( https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)

collymore:
It now looks like this on the left, 
screen shot 2018-02-03 at 12 35 54

It previously sat in the centre of the menu
screen shot 2018-02-03 at 12 36 45

filrak:
@collymore awesome, thanks!

PR #631, opened 2018-02-04
Feature/612

szafran89:
I've decided to create 3 directories:
- `mixins` to store global mixins,
- `filters` to store global filters,
- `helpers` to store global methods which are used in multiple places.

@filrak what do you think about it?


filrak:
Good idea! Can you do the same with plugins (https://github.com/DivanteLtd/vue-storefront/issues/607)? Plus in case of mixins i don't think its neccesarry since we are using our core components as a mixins and they are stored just in `src/components`. The rest is very good idea :)

szafran89:
I've resolved conflicts and added short desc for documentation. Is too late to handle plugins so if you can please merge it and we'll care about plugins in new PR.

PR #632, opened 2018-02-04
Bugfix/footer links

szafran89:
I spotted there are duplicated links in code. I've removed and fixed styles for small devices. Below screens before and after.

Before
screenshot at lut 05 00-42-09

screenshot at lut 05 00-37-28

After
screenshot at lut 05 00-41-16
screenshot at lut 05 00-42-42


src/themes/default/components/core/blocks/Footer/Footer.vue
```diff
@@ -149,69 +159,81 @@ export default {
...
+    &:focus,
+    &:active {
+      .icon {
+        fill: $gray-secondary;
+      }
     }
   }
 
   .links {
+    padding-left: 0;
```
filrak: use atomic class instead please (we use them to have minimal css footprint without repeating things like 'padding-left: 0')

szafran89: Sure, fixed.

szafran89:
@filrak you can check again ;)

filrak:
Great, I'll merge it but i thought that maybe instead of using this media queries we can justa dd some global classes to handle mobile views? Will you take care of this(and make an issue)?

szafran89:
Conflicts resolved. If you can please create new issue for global classes and describe exactly what we need to do ;)

Issue #614, opened 2018-02-01
Move NewsletterPopup to a modal and remove it from vuex store


filrak:
Its connected with https://github.com/DivanteLtd/vue-storefront/issues/608
We should remove  the newsletter popup state from vuex ui-store (and find all occurences of this action dispatch) and use [Modal component](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/components/modal.md) for it

Please update ui-store docs when you're done with it -> https://github.com/DivanteLtd/vue-storefront/blob/master/doc/Working%20with%20UI%20Store%20(interface%20state).md

PR #635, opened 2018-02-05
[#614] Move newsletter to core modal, remove ui/newsletter store

mercs600:
I removed also core component, because there was only store handler.


filrak:
thanks! ;)

PR #646, opened 2018-02-10
Fix i18n import in catalog theme (newsletter component)

qiqqq:
When catalog theme is set, webpack commands crashes with output:
`Module not found: Error: Can't resolve './lib/i18n' `
This change fix this problem


pkarw:
Thx!

Issue #640, opened 2018-02-07
Add components for form inputs


Aekal:
Right now we have a mess in code in places where we use many inputs (Checkout, MyAccount, Login/Register). Also the styles are often duplicated.

We should create reusable components to replace existing inputs and checkboxes.
For input component there should be an option to optionally add validation, icon (show/hide password) and so on.

PanMisza:
@Aekal, You can assign me to this.

Aekal:
Sure, I suggest to do this after merging my changes in #637 to prevent merge conflicts (I've refactored 
 templates/styles code).

Additionally, it'll be nice to remove `#my_account` and `#checkout` styles in `pages/Checkout.vue` and `pages/MyAccount.vue` and change styles scope there to `<style lang="scss" scoped>`. We have duplicated styles for inputs because of this.

Aekal:
@pkarw Could you assign PanMisza to this issue? I don't see him on the list.

filrak:
@Aekal It's merged now ;)

filrak:
@PanMisza it would be also wise to make validation errors a component (i've started it long ago, there is a `validationError.vue` file). Instead of adding multiple div's (or other actions) for different validation errors we can do it in one components and just pass the error messages here.

Aekal:
Maybe we could even mix `FormInput` and `ValidationError` components to one? Then just pass validation as optional prop to `FormInput` component? It may be too complicated but it'll be nice to try.

filrak:
@Aekal very good idea!
 @PanMisza will you take care of it?

PanMisza:
@filrak 
Yes, at the beginning of next week :)

filrak:
@PanMisza great, thanks ;)

pkarw:
@PanMisza please do accept the github invitation to allow me to assign the task to you please :)

pkarw:
@PanMisza we do have release on 1st March; could you handle it before?

PanMisza:
@pkarw 
Sorry for delay... I'm currently changing my flat and have no time for this.
I can finish this task this weekend.

filrak:
@PanMisza Great ;)

pkarw:
@PanMisza how the progress looks like for this one ?:)

pkarw:
@PanMisza any progress on that?

pkarw:
@Aekal any chance that you'll take care of this one maybe next week? It doesn't seem like any body is working on this :)

[...]

Issue #627, opened 2018-02-04
Add the $t() support for i18n for remaining components


pkarw:
I've already added the support for translations (http://kazupon.github.io/vue-i18n/en/api.html#methods) to Category, Product, Home components and all the instances of 'notification' event - system wide according to #618 

We need to add some translation helper calls like:
```
<p>Filters</p>
```
should be changed to:
```js
<p> {{ $t('Filters') }}</p>
```

If there are any remaining, locale specific messages in the code we should change them for example from:

```js
      EventBus.$emit('notification', {
        type: 'success',
        message: `Product ${htmlDecode(product.name)} has been added to the compare!`,
        action1: {label: 'OK', action: 'close'}
      })
```
to

```js
import i18n from 'lib/i18n'

      EventBus.$emit('notification', {
        type: 'success',
        message: i18n.t('Product {productName} has been added to the compare!', { productName: htmlDecode(product.name) }),
        action1: {label: 'OK', action: 'close'}
      })
```

Aekal:
We still need translations for all inputs with validation and checkboxes with label. 
@PanMisza could you apply these translations in #640 ?

pkarw:
I've just fixed the forms

PR #656, opened 2018-02-13
Feature/627 - Translations

Aekal:
#627 
I added missing translations to all components in default theme except:
- all inputs and checkboxes - it should be done within/after the changes in #640 
- placeholder texts in core components, like `Core Header`, `Core shipping` - they are overwritten in theme components


pkarw:
Wow, nice job!

PR #657, opened 2018-02-13
Move templates and styles from few core components to theme

Aekal:
I moved template and styles from core `Overlay`, `OfflineBadge`, `Loader` and `ThankYouPage` to related components in theme. I left only base logic in core.


filrak:
@Aekal thank  you for taking care of it ;)
One thing - please leave the logic on overlay (it works the same everywhere and overriding it is an edge case) and check if both catalog and default templates are working properly with your changes

Aekal:
I think we should move styles from overlay core to theme, because sometimes we may need to change color etc. I also had to move 
```
  beforeCreate () {
    document.documentElement.classList.add('no-scroll')
  },
  destroyed () {
    document.documentElement.classList.remove('no-scroll')
  }
```
from core to theme because:

- `no-scroll` class should be theme specific 
- `beforeCreate` was causing an error in console when switching to catalog theme.

Everything works fine after these changes on each theme.

Aekal:
@filrak Please take a look at my last changes :)

filrak:
@aekal it's ok but please update overlay component in Vue-catalog. It relies on core template. Ok?

filrak:
Great, thanks :)

PR #664, opened 2018-02-14
Issue: 606. Make assets theme-independent.

qiqqq:
Regarding to changes in PromotedOffers.vue I added background-color classes in order to tiles shows correctly without images. Images are now theme-specific and some components like MainSlider or PromotedOffers use global JSON resource to read URLs to images. Now if there are no images in theme, component looks quiet good with background-color only.

I think we should create another task to make some resources theme-specific (this requires also rebuilding core components that use global resources)


filrak:
@qiqqq can you create this task?

qiqqq:
@filrak There is no need to change URLs to images. I added middleware so each image is serve from path /assets/*.* . Images can be placed both in global /asstes/ directory as in theme /assets/ directory and path stays these same. I don't see also any broken images on my local instance so I don't understand your request.

filrak:
Oh ok i wrongly ubderstood what you wrote, sorry :) merging

Issue #654, opened 2018-02-13
Home Page slider CTA is not linked


pkarw:
We need to Just fix the button links on slider 
![207cdd69-6770-4429-bf16-46f9ee8991e0](https://user-images.githubusercontent.com/211899/36135485-b936c268-108b-11e8-8bd0-a67c246a4e18.png)

adaudenthun:
Hi,

I'm new and i think is a good issue to start ! :)
Can you tell me where this link normally send us ?

Thanks

pkarw:
Cool! You can choose it :) maybe just some category?

adaudenthun:
OK , you can assign me this ticket ? 

Thanks ! :)

pkarw:
Please accept the github invitation I’ve just send you to allow me adding you to the project thanks!

adaudenthun:
it's ok for me

adaudenthun:
#680

adaudenthun:
Cool :)

PR #680, opened 2018-02-17
Home Page slider CTA is not linked #654

adaudenthun:
Add prop link to component buttonOutline for use it on the MainSlider


filrak:
Thank you!

PR #701, opened 2018-02-20
Change behaviour of error message (forms validation)

Jassmin586:
Validation errors in forms are hidden now by default. They show up when user click on button or leave a field. I have added '*' to required fields like in the design.


filrak:
Great UX change! thanks and welcome :)

filrak:
@aekal take a look - you'd working on this part of ui too

PR #714, opened 2018-02-21
Improvement/608

mercs600:
I did some cleanup related with modals. All signup modals using core modal now. If you see somewhere extra custom modals let me know.


mercs600:
@pkarw Ok I fixed it

pkarw:
Ok, just update the docs and it’s done

Issue #740, opened 2018-02-23
[catalog] theme not displaying - wrong mixins path


anqaka:
In two files (`src/themes/catalog/components/theme/Search.vue`, `src/themes/catalog/components/core/blocks/Wishlist/Product.vue`) 
there is a wrong mixin's path: 
`mixins: [coreComponent('core/blocks/...')]`
webpack cannot find a module and front doesn't display

filrak:
Feel free to assign yourself to issues you're willing to do ;)

PR #741, opened 2018-02-23
#740 fix mixins path

anqaka:


filrak:
Thanks! I overlooked it during refactoring...
...and welcome ;)

PR #744, opened 2018-02-23
Issue: 685 Move scripts to separate folder

qiqqq:
I checked all scripts after move and this looks like:

fixconfigs.js - I removed it after talking wit Filip
installer.js - I tested it dozen times in any configuration and everything is ok
server.js - I tested it dozen times also and everything is ok
all.js - Here is some problem with this script but propably it is'nt related with path change, on master branch this problem occurs also. @pkarw, you wrote this script, it is everything ok with it?

I don't see need to create webpack alias so I omited it


pkarw:
Thanks!

pkarw:
the all.js script is for running vue-storefront-api and vue-storefront once installed; i'll test it

PR #751, opened 2018-02-25
Issue: 705 Make separate package.json and gitignore files for themes

qiqqq:
Some informations about this changes:

- I added .gitignore and package.json files to all themes (default, catalog and theme-starter)
- I added webpack alias `theme/node_modules` to import dependencies inside themes directly from themes dependencies
- I was wondering if dependencies used globally, for example bodybuilder, vue or vuelidate should be imported from global node_modules or node_modules inside theme and I decided to make themes dependencies fully separeted, so all dependencies needed in themes are imported from node_modules in themes
- I tested each theme and it looks everything works


pkarw:
This is really great but just one thing:
- theme/node_modules prefix is pretty not intuitive for the theme devs, hthey probably will still try to include modules directly (like in manuals given)

Is there any idea on how to avoid the prefix? Is it possible to resolve aliases as multipath with priorities
I mean: to check by default in theme/node_modules - then in parent node_modules etc

qiqqq:
@pkarw I added themes /node_modules/ directory to place where webpack search for modules. This solution allow us to stay with, default importing of modules. I removed webpack alias and changed paths. I talked with @marlass about this and this is good solution. Now webpack search in global /node_modules/ first and in theme's /node_modules/ second. This allow user both use global modules as install own, directly in theme.

pkarw:
Thanks

Issue #660, opened 2018-02-14
Split Vuex Stores to actions/mutations/getters/helpers files


filrak:
Each module of our Vuex Store should be a separate folder containing `index.js` as a main export file and state + `actions.js`, `mutations.js`, `getters.js`. `helpers.js` (if needed). It will make reasoning about Vuex states easier and allow further refactoring of Stores

pkarw:
@szafran89 we do have release planned for 1st March and I belive we should have this one finished for 1.0RC; is it possible for you?

szafran89:
@pkarw I'll do it for sure.

pkarw:
@szafran89 is the 1st March realistic? :)

szafran89:
@pkarw done ;)

filrak:
@szafran89 gr8 ;)

PR #763, opened 2018-02-26
Feature/660

szafran89:
Issue #660


pkarw:
Looks good! Just one question. I see you separated the helpers to helpers.js for each store accordingly which is for sure good practice.

But then, I don’t see import changes to the new module name. For example the attribute helpers seems like not imported anywhere in the code. Could you expand on this? Maybe it’s well tested and I’ve just misssed something :)

In the code there were imports like: import  attributeLabel from ‘store/attribute’
Shouldn’t we change it to import attributeLabel from ‘store/attribute/helpers’?

The rest is fine.  @szafran89

pkarw:
Ok I see that attribute helpers and order helpers are imported.
But product helpers not And he module contains only the event handlers so probably during the build there is no error for missing symbol which doesn’t mean that it will work correctly (no events registered)

Probably you need to put the event handlers within some helper method like registerEventa to not deal with unused import but have this logic up and running

Please heck the other helpers event handlers if there is no similar issue with them

szafran89:
As we discussed over chat, IMO its ok, there are imports `import { configureProductAsync, doPlatformPricesSync, calculateTaxes } from './helpers'` in `product/actions.js`

PR #768, opened 2018-02-27
Issue: 666 Make resources theme-specific

qiqqq:
Solution with move resources to themes directories is not possible because it requires to each theme have each resource (ig_feed.json, slides.json and promoted_offers.json) even if some component is'nt use in this theme. This is bad idea.

As I discused with Filip, temporaily I moved data from this resources to data directly in components. After 1.0 version we will be rebuild this compontents to other solution.

Additionaly I removed order.json and breadcrumbs.json beacause this files are not used anymore.


pkarw:
@qiqqq the slider data should be placed in the theme version of component in this case I belive because now it’s like having the specific data in the core which is no better than having json file :)

Could you move this?

qiqqq:
@pkarw @filrak I added actions and mutations to **promoted-offers** and **social-tiles** and method to **main-slider**. Initially this data are empty and if someone want to use this in theme components, he can update data through mentioned actions or methods.

filrak:
OK, for now its just fine, we will move it to the separate extension in 1.0 milestone. Thank you!

Issue #707, opened 2018-02-21
Make posibility to extend webpack build in themes


filrak:
There should be posibility to extend the webpack build (at least `webpack.base` for now) from themes since core build will be untouchable.

I recommend using this for merging
https://github.com/survivejs/webpack-merge

We should have posibility to add `webpack.extend.js` file in `theme/themename/` which will be merged with current build.

- [ ] Add posibility to merge `webpack.extend.js`  with `webpack.base.config.js` from theme if the file exists
- [ ] Make sure the extended config works properly
- [ ] Describe the solution in docs [link](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/Working%20with%20webpack.md)

pkarw:
This is very important, agree.

marlass:
I can take on that one.

filrak:
@marlass if you want to do more you can make posibility to extend all builds (make `build-extend` folder in  themes and add 3 webpack files named the same as build)

pkarw:
Can we close this on @filrak?

PR #778, opened 2018-02-28
Issue 707: Allow to extend webpack configuration from themes

marlass:
This should allow customizing of webpack config #707 for each theme without the need to override core configs.

I decided to follow solution that next.js and nuxt are using instead of using webpack-merge. This should give more possibilities and control for theme developers.

Here is a preview with example config described in docs:
https://vue-storefront-extend-webpack.now.sh/

To check the code of the example go to https://vue-storefront-extend-webpack.now.sh/_src

With that example, I tested usage of loaders and plugins in the extended config and both are working correctly. Bundle analyzer report was generated, as well as stats file. Additionally, I loaded some json5 file (theme/default/App.vue) and `console.log` it and you can see the result in provided example. Dependencies were fetched from theme node_modules.

And I hope that the docs are somewhat understandable.

### DIFF
--- a/core/build/dev-server.js
+++ b/core/build/dev-server.js
@@ -1,8 +1,17 @@
 const path = require('path')
 const webpack = require('webpack')
 const MFS = require('memory-fs')
-let clientConfig = require('./webpack.client.config')
-let serverConfig = require('./webpack.server.config')
+
+let baseClientConfig = require('./webpack.client.config')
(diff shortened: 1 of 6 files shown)


filrak:
Just wow! Probably the most prefessional pull request in VS history.

pkarw:
Congratulations @marlass! Really  cool stuff! Welcome to the core dev team :)

PR #822, opened 2018-03-08
Adding support for "Payment Extensions"

MikeSheward:
Adding support for "Payment Extensions", adding first Payment Extension, cash on delivery as an example and updating the accompanying documents. Payment methods are now extensible and dynamic, and can by dynamically added instead of a fixed JSON file with them in.

You may need to clear Application Cache.


pkarw:
@MikeSheward Great job - please take a look at https://github.com/DivanteLtd/vue-storefront/pull/781

Its about synchronization of the payment list with Magento. I belive we need to connect these two Prs into one

@akbarik @MikeSheward could you discuss how to merge it altogether?

kyvaith:
Maybe this could be a more standard way to implement different payment extensions: https://www.chromestatus.com/features/5160285237149696

pkarw:
@kyvaith we definitely should implement the Payment Request API, nevertheless, it's not widely supported yet. It's on the roadmap - not with the highest priority - it would be great to have it as an extension/npm module

kyvaith:
Just to be clear, I'm referring to Payment Handler API not Payment Request API.

Issue #729, opened 2018-02-23
Add separata package.json files for extensions / install script


filrak:
- [x] Add extension-specific dependencys to extensions files
- [x] Add extension install script registering it in config

pkarw:
@qiqqq could you take care of this?

pkarw:
@MikeSheward we should allow users to install extensions as a npm packages - for example: https://github.com/develodesign/vue-storefront-stripe

Should we have a web pack alias that will allow to include the extensions from /src/extensions and /src/extensions/node_modules?

filrak:
@pkarw yes, we should add webpack alias `ext` and move the extension registtration from json to index.js file at` src/extensions`

pkarw:
I belive it's just the https://github.com/DivanteLtd/vue-storefront/blob/8184c8e41888d4bb690642f5c7b4e972f10eaa30/core/lib/extensions.js#L1 to be modified

@MikeSheward will you take care of it Tomorrow? @filrak

filrak:
@MikeSheward please contact with @Igloczek. He is working on moving themes to npm packages. Both tasks should use the same format

Igloczek:
I'll take this over, the plan is to add new CLI command that will handle the process of adding new packages and necessary configuration.

For now, I will just symlink from `node_module` to the destination directory to don't mess with webpack for now and keep all setup scripts in the core.

I'd like to add a npm task like `npm run package` that will have the interactive shell with 3 questions
1. Type of package extension / theme
2. Name of package
3. Destination path of the package

Depends on the user input it will add necessary stuff to the `local.json` config file, add the dependency to the `package.json`, install and then symlink it to the destination directory

On the `local.json` I'd change `registerExtensions` to `extensions` and push all package config inside this object and add `themes` that will do same for themes.

pkarw:
@Igloczek @filrak I've done this with #878

Because of the web pack nature (doesn't support dynamic import / require) I had to modify the extension loader - moving enabled extensions from config/local.json to /extensions/index.js 

Otherwise, it wouldn't have been possible to load extensions installed via npm like - https://github.com/develodesign/vue-storefront-stripe

By the way - I've added dependency to vue-storefront-stripe by default @MikeSheward - good job :)

PR #878, opened 2018-03-16
Added stripe payments dependency, modified npm extension loader #729

pkarw:


pkarw:
@Igloczek @filrak I've done this with #878

Because of the web pack nature (doesn't support dynamic import / require) I had to modify the extension loader - moving enabled extensions from config/local.json to /extensions/index.js

Otherwise, it wouldn't have been possible to load extensions installed via npm like - https://github.com/develodesign/vue-storefront-stripe

By the way - I've added dependency to vue-storefront-stripe by default @MikeSheward - good job :)

pkarw:
Now, the extensions can be installed by:
- creating the extension folder inside `src/extensions`
- installing the ext via `npm install <extensionpackage>`

In both cases, the newly installed extension should be adedd to the `src/extensions/index.js`:
https://github.com/DivanteLtd/vue-storefront/blob/develop/src/extensions/index.js

Please notice: I's been merged to the `develop` branch

PR #879, opened 2018-03-16
Auto refresh

AurelienLavorel:


filrak:
I dont think auto refresh is a good idea. It will add traffic without a reason

pkarw:
@filrak this is just the first request when app is being built by web pack and it takes like 90s this is quite usefull for me :)

pkarw:
@AurelienLavorel could you add very basic but proper html markup around the document - html /head / body? Maybe just load some static template from like: /var/app-loaded.html file

AurelienLavorel:
Yes it's just to avoid spamming F5 on that page, here we can make another task, we'll see title update when page is ready.
@pkarw I updated html

pkarw:
Thanks!

PR #883, opened 2018-03-17
[update] ux improvement for personnal details step

AurelienLavorel:


pkarw:
Hi @AurelienLavorel, could you please attach the screen how does it look like after change?

AurelienLavorel:
Of course.

Before:
before
After:
after

I exchange **login** and **next step** position to respect the global design, validation button is always in right position.

When you press "enter" on e-mail field, next step is loaded.

pkarw:
Cool! Such UX improvements are much appreciated! Thanks for that @AurelienLavorel good job.

pkarw:
@AurelienLavorel could you please move the submit buttons related to the other checkout steps to the right as well to make it more coherent? https://www.dropbox.com/s/e2pgvkytc5cgedk/Screenshot%202018-03-18%2012.55.26.png?dl=0

PR #891, opened 2018-03-18
[add] ability to press back in checkout

AurelienLavorel:
https://github.com/DivanteLtd/vue-storefront/issues/885


pkarw:
It can be accepted as a interim/temporary solution; I belive it’s possible to make it as a vue router routing - something like (route config):

{ name: 'newsletter', path: '/my-account/newsletter', component: MyAccount, props: {activeBlock: 'newsletter'} }

By doing so we can avoid hashes and playing directly with window.history

The question is if it doesn’t require very deep changes to the Checkout logic / structure.

Would be great if you can investigate this and check if it’s feasible (with router changes)

PR #907, opened 2018-03-23
Coding guidelines docs

Igloczek:


doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
+# Vue Storefront Coding Guidelines - DRAFT
+
+Hello Extension / Theme / Core developer! This is just a short guide to align the code style among the core modules and all the extensions and thems.
```
Igloczek: I'm not sure if it's necessary.

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+
+```
+|____store.js
+|____index.js
+|____components
+| |____ComponentExample.vue
+|____router.js
+|____package.json
+|____pages
+| |____PageExample.vue
```
Igloczek: I'd suggest to keep folders on top and then files, both in alphabetical order.

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
+# Vue Storefront Coding Guidelines - DRAFT
+
+Hello Extension / Theme / Core developer! This is just a short guide to align the code style among the core modules and all the extensions and thems.
+
+## General guidelines
+In general we're using kind of mix of: [Airbnb JavaScript Style Guide](https://github.com/airbnb/javascript) and [JavaScript Standard Style](https://github.com/feross/standard/blob/master/RULES.md#javascript-standard-style). There are a eslint and vue-eslint configured to enforce the style. Please do use the same [configs in Your modules](https://github.com/DivanteLtd/vue-storefront/blob/master/.eslintrc.js). 
+
+## ES2015
+We do preffer to use ES2015 constructions in the code - and there is a babel compiler set to allow us to do so. So use the ES modules instead of `require` and feel free to use other fancy structures.
```
Igloczek: I'm not sure if we should recommend using ES6 for custom modules. Currently, we are using [some hacks](https://github.com/DivanteLtd/vue-storefront/blob/develop/src/extensions/index.js) to run Babel on code from `node_modules`, will be nice to get rid of it somewhere in the future.

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+All the component properties should be named using `camelCase` as well.
+All the component data fields should be `camelCased`.
+
+Theme catalog structure is well described in here: [Working with themes](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/themes/Working%20with%20themes.md)
+
+## Extensions
+For the reference of how extension can be structured please take a look at [template](https://github.com/DivanteLtd/vue-storefront/tree/develop/src/extensions/template).
+Check the [Working with extensions](https://github.com/DivanteLtd/vue-storefront/blob/develop/doc/extensions/Working%20with%20extensions.md) for more info.
+
+It's something like this:
```
Igloczek: IMO it's not necessary

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+|____components
+| |____ComponentExample.vue
+|____router.js
+|____package.json
+|____pages
+| |____PageExample.vue
+```
+
+## Events
+We're using [Vue events](https://vuejs.org/v2/guide/events.html) for async. operations handling and exposing some extension points in the code.
```
Igloczek: These part of Vue docs are related to handling typical DOM events like `click` or `change`.

In our case, we should link to https://vuejs.org/v2/guide/components.html#Using-v-on-with-Custom-Events and https://vuejs.org/v2/api/#Instance-Methods-Events + maybe also some docs about creating EventBus. 

But... I'd suggest to get rid of this concept and move everything to Vuex, b/c using two ways of managing app state is not necessary and not easy to understand / work with.

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+|____attribute
+| |____mutations.js
+| |____actions.js
+| |____getters.js
+| |____index.js
+| |____helpers.js
+|____homepage
+| |____index.js
+...
+```
```
Igloczek: I think we should list all store modules and show the general structure separately.

Something like this:

Store modules:
```
- attribute
- homepage
- order
```
Single module structure:
```
- actions.js
- getters.js
- helpers.js
- index.js
- mutations.js
```

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+Check the [Working with Vuex](https://github.com/DivanteLtd/vue-storefront/blob/develop/doc/data/Working%20with%20Vuex.md) for more info.
+
+Please use Vuex for all state/data management and avoid loading local properties/local data inside Your components. This will casue problems with reactivity and SSR.
+
+## Data fetching
+All remote operations should've been placed inside the Vuex actions, using `fetch()` or exisiting [`data/sync`](https://github.com/DivanteLtd/vue-storefront/blob/develop/doc/data/Sync%20Store.md) module.
```
Igloczek: I think we should recommend only one way dealing with external data, with all logic handled by VS core.

Igloczek: Probably we need to take a look at the naming of this part, b/c for me `sync` = `synchronously` and I believe that most ppl will expect something like `fetch` or `get`, b/c we are not synchronizing the whole store with some external resource, but just asking about some particular set of data.

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Offline support
+Vue Storefront should be able to work offline. All key data must be stored using `localForage`. The network requests that are modifying the client's or other data records should be [queued using the `data/sync` module](https://github.com/DivanteLtd/vue-storefront/blob/develop/doc/data/Sync%20Store.md) Please see more on this in [Working with data](https://github.com/DivanteLtd/vue-storefront/blob/0128bad05b6eeb19d7bd5b8cebcc56df28892c84/doc/Working%20with%20data.md).
+
+## SSR Support
```
Igloczek: I'd move this to own page with all about SSR.

doc/Coding Guidelines.md
filrak: I think people will use it anyway as it is heavly used across whole project (even extensions are using es6 export  instead of module.exports)

doc/Coding Guidelines.md
filrak: I think we should just link to this folder instead of writing down the structure. When it changes it will require extra step to be up-to-date in docs ;)

doc/Coding Guidelines.md
filrak: Agree. We should use Vuex not only b/c its easier to maintain but its also easier to understand. Also currently we are depending too much on the events. I tried to change some things yesterday with @qiqqq  ago and it's worse than visiting hell now b/c of multilevel event handling. It would be much cleaner if we will group logic in smart components and keep the app state in Vuex. Some of the events should stay but most of them is redundant

doc/Coding Guidelines.md
filrak: Maybe just link to Working with Vuex and extend this docs?

doc/Coding Guidelines.md
```diff
@@ -0,0 +1,107 @@
...
+        return resolve()
+      }).catch((err) => {
+        console.error(err)
+        return resolve()
+      })
+    })
+  },
+```
+
+## Core vs Theme components
```
filrak: Just link to Working with components

doc/Coding Guidelines.md
pkarw: I think we rather should create a boilerplate / npm module for extension

doc/Coding Guidelines.md
pkarw: Good point; I think we should change sync to fetch - I can do this

pkarw: sync = synchronization :D

PR #909, opened 2018-03-23
Update 'Installing on Linux and MacOS.md'

Tjitse-E:
When installing vue-storefront-api I noticed that docker is not enough. You also need 'docker-compose'.


pkarw:
Thanks!

Issue #957, opened 2018-03-30
Button shift


StasiekDivante:
When you hover over topbar icon on desktop button shifts ~2px to the bottom

![screen shot 2018-03-30 at 11 31 33](https://user-images.githubusercontent.com/37068687/38133985-a72bd9ce-3411-11e8-8801-3d9e117291b8.png)

icons should be aligned to exact height of topbar

<img width="220" alt="screen shot 2018-03-30 at 12 00 57" src="https://user-images.githubusercontent.com/37068687/38134146-2030124a-3412-11e8-802d-b3dd42580bbf.png">

lorenaramonda:
Hi @StasiekDivante can you please more specific about browsers and devices where this happens?

StasiekDivante:
sure, it's safari on macOS High Sierra

lorenaramonda:
ok, thanks. gonna have a look 😉

PR #962, opened 2018-03-30
ref. #957 reset margin from all buttons

lorenaramonda:
Remove margin from all buttons to prevent default browsers margin such as safari


pkarw:
Thanks!

Issue #956, opened 2018-03-30
Slight color change


StasiekDivante:
Change all #333 colors to #4F4F4F

lorenaramonda:
Hi @StasiekDivante just to be sure to operate in the right context, can you make an example in where it should changed, ie a button?
I see that #333 is set as value of the color variable of night-rider, if this color should not be present at all we could consider to remove it from variables list and replacing it where used with the proper color (#4f4f4f)

PR #963, opened 2018-03-30
ref. #956 replace all #333 colors to #4F4F4F

lorenaramonda:
Notice: pay attention that this may disable color change effects on hovering some element, especially those that use primary color variables as now they're the same in both statuses:

  primary: (
    default: map-get($colors, matterhorn),
    hover: map-get($colors, matterhorn)
  ),

matterhorn is #4F4F4F


lorenaramonda:
ops i made some mess on the commit 🙄 gonna PR the correct if I can make i working again

PR #964, opened 2018-03-30
ref. #956 replace all #333 colors to #4F4F4F

lorenaramonda:
Notice: pay attention that this may disable color change effects on hovering some element, especially those that use primary color variables as now they're the same in both statuses:

primary: (
default: map-get($colors, matterhorn),
hover: map-get($colors, matterhorn)
),

matterhorn is #4F4F4F

(this is the right PR, I hope! thanks @pkarw for the patience!)

### DIFF
--- a/src/themes/catalog/components/core/Notification.vue
+++ b/src/themes/catalog/components/core/Notification.vue
@@ -73,6 +73,6 @@ export default {
 }
 
 .notification.info {
-  background: #333333;
+  background: #4F4F4F;
 }
 </style>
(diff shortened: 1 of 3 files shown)


pkarw:
@StasiekDivante could you check the hovers - as Lori they disappear after this change and I’m not sure if that’s expected effect?

StasiekDivante:
@lorenaramonda that's one good point, haven't noticed that

By color replacement I meant replacement in font colors - headers, link etc
Thanks for staying attentive!

pkarw:
@lorenaramonda could you just adjust this change to not cover the elements that have some hover effect?

lorenaramonda:
Yes @pkarw sure. I'll have a check!

lorenaramonda:
Now it should have been replaced where needed. The change also affected the .notification.info class in src/themes/catalog/components/core/Notification.vue. Please, check that this is the desiderata.

pkarw:
Thanks!

Issue #99, opened 2017-10-12
UI for product Galery + Display subsequent simple product photos in gallery on configurable


pkarw:

karolbzik:
@pkarw please be more specific here

pkarw:
For configurable products we do have separate pictures for each variant (for example blue tee or red one etc) - we can display the subsequent pictures in kind of gallery or slider

How it should look like?

We can also synchronize product media (more than 1 picture assigned to the product) - so basically we can asume we have product galleries

Igloczek:
Maybe we can go with a simpler design without the typical gallery [like Nike did in their new store](https://econsultancy.com/blog/69844-why-nike-s-refreshed-product-pages-improve-cx-beat-adidas).

pkarw:
@karolbzik I would love to implement media sync and close the gallery feature; are you willing to provide some designs?

karolbzik:
@pkarw it's ready

Gallery on mobile, visible after clicking zoom icon / photo
<img width="360" alt="3 3 product card with zoomed photo on mobile" src="https://user-images.githubusercontent.com/18116406/38165462-f9bedd8c-3513-11e8-823f-6d42313b7bf8.png">

Gallery on desktop, visible after clicking zoom icon / photo
<img width="1440" alt="3 2 product card with zoomed photo" src="https://user-images.githubusercontent.com/18116406/38165448-d3c71a0e-3513-11e8-92ae-9b7fe396dbad.png">

Zoom button on mobile
<img width="360" alt="3 1 product card on mobile" src="https://user-images.githubusercontent.com/18116406/38165449-d3e972c0-3513-11e8-94d2-801f89ff2bb7.png">

Mouseover on product's photo on desktop
<img width="1440" alt="3 product card mouseover on image" src="https://user-images.githubusercontent.com/18116406/38165450-d40bb66e-3513-11e8-8f27-b04b88c4cb9b.png">

Product card with zoom icon
<img width="1440" alt="3 product card" src="https://user-images.githubusercontent.com/18116406/38165451-d430d76e-3513-11e8-8abf-c6e940c1465c.png">

karolbzik:
All designs are available here - https://www.figma.com/file/VKyqbHFI55TKIKcQlFLiVpVF/Vue-Storefront under the names:
- Product card (including mouse over and mouse away the image)
- Product card on mobile
- Product card with zoomed photo
- Product card with zoomed photo on mobile

Important issues:
[...]

PR #980, opened 2018-04-02
#99 product gallery

nuovecode:
I created the array of gallery img in the [product page](https://github.com/bitbull-team/vue-storefront/blob/feature/99-gallery/core/pages/Product.vue) but I think that this job should be done within vue-storefront-api and it should include the whole gallery.

Zoom opens when double clicked because the click event was intercepted by the swipe. If you have better ideas let me know. 

Is SRR support for product image missing?

### DIFF
--- a/core/components/ProductGallery.vue
+++ b/core/components/ProductGallery.vue
@@ -1,0 +1,53 @@
+<template>
+  <!-- Move in default theme -->
+  <div class="row">
+    <div class="col-xs-2">
+      <div
+        v-for="(images, key) in gallery"
+        :key="key">
+        <transition name="fade" appear>
+          <img v-lazy="images.path" class="mw-100" ref="images.path" @click="navigate(key)">
+        </transition>
+      </div>
+    </div>
+    <div class="col-xs-10">
+      <carousel
+        pagination-active-color="#828282"
+        pagination-color="#fff"
+        :per-page="1"
+        mouse-drag="false"
+        ref="carousel">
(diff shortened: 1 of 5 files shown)


filrak:
Thanks! Can you provide screenshots for default theme version?

pkarw:
Could you please @nuovecode add a special case when the returned gallery array is by any reason empty to display the single default image?

pkarw:
There is one more case to handle; after switching the product variant (https://www.dropbox.com/s/2v793yocqo1gcmn/Screenshot%202018-04-02%2017.27.53.png?dl=0) the current (`this.product.image`) product image is switched to the variant-one - which is great because You can display the image of specific variant and check if it's suitable - color, size etc.

We need to have this behaviour. If you need to have the variant's images on the `gallery` list to implement this - no problem, You can just modify https://github.com/bitbull-team/vue-storefront/pull/1/files - change the "else" section to join the configurable_children images to the media_gallery

pkarw:
Another case is offline support. In ofline mode we should have just the pre-cached image displayed in the gallery or even switch off gallery view?

The offline mode image is the one displayed on the product list. You can get by:
```
import { productThumbnailPath } from 'core/helpers'
const offlineImageUrl = productThumbnailPath(this.product, true)
```

So maybe it's an option to get it as the only one gallery item in case of

```
if (typeof navigator !== 'undefined' && !navigator.onLine) {
```

PR #1007, opened 2018-04-10
Update gallery with new requirements

nuovecode:


pkarw:
3. The animation when switching the photos doesn't look good - product disappears and then suddenly appears back; looks strange can we work on this?

pkarw:
4. It makes no sense to show the yellow color X available sizes (this is the default behaviour - as configurable_children contains all possible configurations). We should limit it somehow by checking the attributes. Moreover - as this is kind of framework You cannot hardcode "color" attribute for this. So You need to have it in the config (please add the variable to default.json - like "products.variantGalleryAttributeName":"color" (sounds stupid, I know, no better idea ;))

**I'm wondering if adding configurable_children to the gallery makes sense. It would be maybe better to just switch the 1st photo in the gallery with the selected variant? In this case the gallery will be always like 3-4 photos not 12-15 like now**

filrak:
Ad 2 - we have vue-offline plugin for this to handle offline states in more elegant way :)

pkarw:
Great job, we're almost there! 

The issues after tests:

1. When switching back from "offline" to "online" by some reason the image gallery doesn't load the content: https://www.dropbox.com/s/vyt9a5fhlusft0h/Screenshot%202018-04-25%2012.46.10.png?dl=0

The offline mode works great (online->offline switching works just fine!)

2. There is no placeholder while pictures are loaded - both on Desktop and Mobile: https://www.dropbox.com/s/gtpd33z16a9qh1q/Screenshot%202018-04-25%2012.44.53.png?dl=0

3. Not sure if it's connected with 2 - but I had an experience that switching picture on the mobile (< > arrows) doesn't always work. Please also center the picure verticaly in the zoom mode on mobile https://www.dropbox.com/s/oob6i3je3tixhxt/Screenshot%202018-04-25%2012.45.33.png?dl=0

PR #1022, opened 2018-04-13
Preparation to release store as a separate package

### LINKED ISSUE #1017: Publish the core/store as separate npm extension using Lerna
Bartek coul You help me with this?

Igloczek:
That's the first step, then I think we should move the code to `packages/@vue-storefront/core` along with other extensions and themes, to make directory structure a bit flatter.

I'm not sure what is wrong with unit tests, but I'm able to get same errors on develop too. I'll investigate it next week.

Closes #1017

### DIFF
--- a/core/app.js
+++ b/core/app.js
@@ -3,7 +3,7 @@ if (!global.$VS) global.$VS = {}
 import _ from 'lodash'
 import Vue from 'vue'
 import App from 'theme/App'
-import store from 'core/store'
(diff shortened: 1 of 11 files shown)


pkarw:
@Igloczek cool thanks! But how to publish the store to the public npm repository as a module?

Igloczek:
Published https://www.npmjs.com/package/@vue-storefront/store

PR #1029, opened 2018-04-16
Fix add config var for attributes referenced by attribute_code rather than label

jimcreate78:
Fix for product attributes with similar names that are broken when referencing by label. This allows using attribute_code rather than the attribute label.


pkarw:
Thanks!

PR #1030, opened 2018-04-16
use main image when no configurable children image

jimcreate78:
We don't have images for our configurable_children so when products load there's no image.  This was added so you can have configurable products with only main images.


pkarw:
Thanks!

Issue #1026, opened 2018-04-14
Add configurable host to json configuration 


Tjitse-E:
`yarn start` now always uses localhost, but in some cases we need to specify a real IP address (on a shared server for example). It would be handy if we could specify this in the default.json/local.json like:

```
{
    "host": "127.0.0.1"
    "elasticsearch": {
      "httpAuth": "",
...
```

PR #1067, opened 2018-04-20
Made host and port configurable in json configuration

Tjitse-E:
Fix for #1026

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -1,6 +1,8 @@
 {
-    "host": "localhost",
-    "port": 3000,
+    "server": {
+      "host": "localhost",
+      "port": 3000
+    },
     "elasticsearch": {
       "httpAuth": "",
       "host": "localhost:8080/api/catalog",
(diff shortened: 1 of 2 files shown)


config/default.json
```diff
@@ -1,4 +1,6 @@
 {
+    "host": "localhost",
```
pkarw: I suggest creating a subsection "server" containing both of these variables ok?

core/scripts/server.js
```diff
@@ -93,7 +94,8 @@ app.get('*', (req, res) => {
     .pipe(res)
 })
 
-const port = process.env.PORT || 3000
+const port = process.env.PORT || config.port
+const host = config.host || 'localhost'
```
pkarw: Please keep it consistent and allow the user to set host via process.env.HOST as well

pkarw:
@Tjitse-E could you add the same feature to the vue-storefront-api?

Issue #897, opened 2018-03-20
Add 'proceed to checkout' action in 'Add to cart' notification


filrak:
Such an action would be a great UX improvement
![image](https://user-images.githubusercontent.com/15185752/37651875-e3794a86-2c39-11e8-9d9f-357e4e9f8b2f.png)

pkarw:
We need extend the `notification` message for the `action2` I belive:

```
          this.$bus.$emit('notification', {
            type: 'error',
            message: i18n.t(result.result),
            action1: { label: 'OK', action: 'close' }
          })

```

Januszpl:
Working on it

PR #1070, opened 2018-04-20
add proceed to checkout action to add to cart notification

Januszpl:
**Ref:** #897
 
**Changes:**
add action2 goToCheckout (and btn to add to cart notification) that redirect to checkout page

**Before:**
before-desktop
before-mobile

**After:**
after-desktop
after-desktop2

### DIFF
--- a/core/components/Notification.vue
+++ b/core/components/Notification.vue
@@ -27,6 +27,11 @@ export default {
       switch (action) {
         case 'close':
           this.notifications.splice(id, 1)
+          break
+        case 'goToCheckout':
+          this.$router.push('/checkout')
+          this.notifications.splice(id, 1)
+          break
       }
     }
   }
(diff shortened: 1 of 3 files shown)


core/components/Notification.vue
```diff
@@ -27,6 +27,11 @@ export default {
       switch (action) {
         case 'close':
           this.notifications.splice(id, 1)
+          break
+        case 'goToCheckout':
+          this.$router.push('/checkout')
+          this.notifications.splice(id, 1)
+          break
```
pkarw: To make the action mechanism more generic please EventBus.$emit, the eventName should be set to “action”. By doing so, extension developers will be able to handle their own custom cart actions. 

Then the handler of goToAction should’ve been handled by some event listener bound in the Microcart.vue in the core or the default theme

We shouldn’t have placed any specific logic inside general component of Notification; this components shouldn’t have known that we have some checkout in the app etc just display the general notifications and emit general events

The event name should be: “cart-after-goToCheckout”

pkarw:
OK, I'll fix it

PR #1072, opened 2018-04-21
Fixes #954: Better heading spacing for headers.

### LINKED ISSUE #954: Spacing problem
right now header have same margin on top and bottom

screen shot 2018-03-30 at 11 47 55

Spacing should be 35px on top and 10px at the bottom on mobile

screen shot 2018-03-30 at 11 48 05

and 40px top 25px on bottom on desktop

screen shot 2018-03-30 at 11 49 40

bobbyshaw:
Took a quick look at #954. I read up on atomic CSS but it seems that because this is loaded from markdown files through StaticContent element, I can't do this through the use of classes on elements. Is that correct?

So I went with the route of doing this in the _text.scss file as my best guess. The original ticket wasn't clear, maybe i'll want to restrict this to h3's within static-content class? And maybe I want to consider other header elements. 

Anyway, I don't expect this PR to be approved without some changes but it'll start a conversation at least!

screen shot 2018-04-21 at 13 55 37
screen shot 2018-04-21 at 13 55 46

### DIFF
--- a/src/themes/default/css/base/_text.scss
+++ b/src/themes/default/css/base/_text.scss
@@ -33,6 +33,13 @@ h2,
 h3,
 .h3 {
   font-size: 24px;
+  margin-top: 35px;
+  margin-bottom: 10px;
+
+  @media  (min-width: 767px) {
+      margin-top: 40px;
+      margin-bottom: 25px;
+  }
 }
 h4,
 .h4 {


pkarw:
Cool, we need @filrak and @StasiekDivante to come on in on this. Guys what do you say?

filrak:
@bobbyshaw thank you very much for your PR ;) I think @StasiekDivante intention was to make it available only for static content. In that case modifying  `Static.vue` file would probably be the best idea. If the `scoped` attribute is a problem just remove it and wrap whole `Static.vue` (the top-level div) and it's SCSS in one class (e.g. static) so it will affect only this part of the page. Modifying the global h3 can be risky ;)

pkarw:
@bobbyshaw could you adjust this fix to what @filrak recommended and we'll merge this in? :) Thanks!

Issue #912, opened 2018-03-24
Fix checkout form in mobile


noindie:
I suggest you to fix this behaviour in mobile form: when you forget to fill in a field and click on the button to proceed there usually is a javascript that brings you back to the "incomplete" field that is required.

![vuestorefronycheckout](https://user-images.githubusercontent.com/10724669/37865738-378d1c0c-2f81-11e8-96ed-69f08b5884e9.png)

ciao

bobbyshaw:
I've started work on this if you wish to assign to me.

ElijahKotyluk:
Is this issue available?

PR #1073, opened 2018-04-21
Work on #912. Add validation to personal details form.

bobbyshaw:
To achieve this we remove the disabling of the button and instead on click, check the form validation and add a validation error message if needed. This follows the user experience that's currently on the account registration page.

Again, not written vue before so let me know if this is the right sort of approach to solve this problem. If it is, or I get advice on a better way of approaching it, I can implement the other checkout steps.

screen shot 2018-04-21 at 15 32 14
screen shot 2018-04-21 at 15 32 21

### DIFF
--- a/core/components/blocks/Checkout/PersonalDetails.vue
+++ b/core/components/blocks/Checkout/PersonalDetails.vue
@@ -6,6 +6,7 @@
 
 <script>
 import { mapState } from 'vuex'
+import i18n from 'core/lib/i18n'
 
 export default {
   name: 'PersonalDetails',
@@ -33,9 +34,19 @@ export default {
   methods: {
     sendDataToCheckout () {
       if (this.createAccount) {
+        if (this.$v.$invalid) {
+          this.$v.$touch()
+          this.validationError()
+          return
+        }
         this.personalDetails.password = this.password
         this.personalDetails.createAccount = true
       } else {
+        if (this.$v.personalDetails.$invalid) {
+          this.$v.personalDetails.$touch()
+          this.validationError()
+          return
+        }
         this.personalDetails.createAccount = false
       }
       this.$bus.$emit('checkout-after-personalDetails', this.personalDetails, this.$v)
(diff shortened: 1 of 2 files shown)


pkarw:
Seems like a way to go. The missing part is setting the focus to the first field to be fixed (+ scrolling to it)

StasiekDivante:
Hey @bobbyshaw, I tottaly agree, this is proper way of handling forms, lets implement this solution

pkarw:
@bobbyshaw is it likely to finalize this task till Monday as I'm planing the release od 1.0RC3 for Monday

bobbyshaw:
@pkarw Ha, sorry, I was taking a rest after Imagine! Waiting for all sections of checkout to have this functionality makes sense. I can look into implementing that as well as setting field focus on the first invalid field and scrolling to it. Any pointers on where that's already done in the codebase or should I implement manually?

Also, does anyone more familiar with writing components have a way of doing this without duplicating the same code in each component or is that *the right way* with componentisation?

bobbyshaw:
Cool, that looks like the sort of thing I want.

filrak:
@bobbyshaw if you'll have time for more complex solution probably the best way is to create whole form as a separate core component with all validation logic inside and add input field components (base-input, base-checkbox etc) via slots. With this approach you can create a function to handle button click which will valiate all fields inside the form components (https://vuejs.org/v2/guide/components-slots.html). @pkarw @bobbyshaw let me know what do you think about this approach

pkarw:
You’re right @filrak, however it sounds like a massive task (including all the tests required). So maybe it’s more like the 2.0? And now we just will do the first step? I don’t know, just thinking aloud :)

filrak:
@pkarw it depends on how much time @bobbyshaw is willing to spend on it but even minor improvements are appreciated :)

@bobbyshaw what do you think?

pkarw:
@bobbyshaw any chances to finish this one ? :)

bobbyshaw:
I should have time over the next couple of weeks to evaluate what needs to be done and give feedback on what I'll be able to do and by when if that's ok? No hard feelings if you need to unassign the task from me.

pkarw:
OK! I'll leave it open - if anyone will be able to do it in the meantime then fine ;) Otherwise please feel free to continue :)

PR #1131, opened 2018-05-11
Dry Run option for server merge

pkarw:


pkarw:
OK I also made it. There is new config val:

* {{cart.server_merge_by_default}} \- set it to false to disable automatic cart merge
* then if set to false please add Event handler to the     {{EventBus.$emit('servercart-after-diff', \{ diffLog: diffLog, serverItems: serverItems, clientItems: clientItems, dryRun: event.dry_run \}) // send the difflog}} 
* You'll get server cart, client cart \+ diff log: [https://www.dropbox.com/s/i4vr46v9qvp2ys7/Screenshot 2018-05-11 11.32.53.png?dl=0|https://www.dropbox.com/s/i4vr46v9qvp2ys7/Screenshot%202018-05-11%2011.32.53.png?dl=0]
* All You need to do in the Popup is:
* when user selects to use local cart \- just do nothing
* *please note* in the event handler for 'servercart-after-diff' \- *show the popup only* when *payload.dryRun = true*
* when user selects to user server cart \- call:     {{rootStore.dispatch('cart/serverPull', \{ forceClientState: false, dryRun: false \}, \{ root: true \})}} 
* this call will in return also fire '{{servercart-after-diff}} but with payload.dryRun = false in this case \- so ignore this event regarding the popup

PR #1146, opened 2018-05-16
Webpack 4 upgrade

DavidRouyer:
* Upgrade to Webpack 4.8.x
* Update to ESLint 4.19.x (new recommended rules)
* **[BREAKING]** Removed corePage() and coreComponent()
* Added eslint-plugin-vue-storefront to migrate corePage() and coreComponent() (Contains a fixer to do it automatically)


DavidRouyer:
Can anyone take a look?

core/build/webpack.prod.client.config.js
```diff
@@ -6,6 +6,7 @@ const themeRoot = require('./theme-path')
 const extendedConfig = require(path.join(themeRoot, '/webpack.config.js'))
 
 module.exports = extendedConfig(baseClientConfig, {
+  mode: 'production',
```
marlass: What's the use case for this?

core/build/webpack.prod.server.config.js
```diff
@@ -6,6 +6,7 @@ const themeRoot = require('./theme-path')
 const extendedConfig = require(path.join(themeRoot, '/webpack.config.js'))
 
 module.exports = extendedConfig(baseServerConfig, {
+  mode: 'production',
```
marlass: Same here.

core/build/webpack.prod.client.config.js
DavidRouyer: The production mode can be configured in the config file or by passing the flag in the command (see [here](https://github.com/DivanteLtd/vue-storefront/pull/1146/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R33) and [here](https://github.com/DivanteLtd/vue-storefront/pull/1146/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R34)). I prefer setting the mode in the config files (dev and prod) and remove them in the command, so that the behavior is the same for both configuration (and no need to remember to set the mode when running webpack)

marlass: I think appending mode in here won't include it in the final configuration, because it is only an object to use by users to extend config in their themes, so they can append different things in dev and production mode. Appending it here won't solve this problem. I this case I would stick with flag in command.

DavidRouyer:
@marlass I have addressed your feedback

marlass:
@DavidRouyer Thank you. I will test this PR more thoroughly on Monday and will let you know if everything is ok and if this PR can be safely merged.

Issue #892, opened 2018-03-18
Add all required install steps to Travis


pkarw:
http://elliot.land/post/using-docker-compose-on-travis-ci

Its the core/scripts/installer.js procedure but not in interactive mode to be copied here. Doing so we will be able to run some HTTP tests to check if there are no critical runtime errors to the core after each build

pkarw:
The installer script probably just need to have an —silent mode to install the app using defaults

marlass:
I can work on that issue.

PR #1154, opened 2018-05-18
Run installer on travis

marlass:
Used commander.js to easily handle extra flags and arguments to the installer.
Extracted default config from questions. Running with the flag `--default-config` will use those defaults instead of interactive mode.

Solves issue: #892 

Also, I installed Cypress and set up the first test that go through catalog, product page, and checkout to be sure that core functionality works.


pkarw:
Is it ready for CR and merging it in?

marlass:
Yes.

pkarw:
@marlass could You please write a short doc on the e2e tests - how to run them and extend them + create few github issues with the default scenarios for somebody to just write the tests fro them?

PR #1155, opened 2018-05-18
[SCSS] Adding missing variable in _border.scss

Aylay:
- Adding missing `-#{$i}` in name class in src/themes/default/css/layout/_border.scss
- Updating every vue file with `brdr-bottom-1` instead of `brdr-bottom`


Igloczek:
Thanks for the contribution!

PR #1168, opened 2018-05-18
Move core components to js files

Aekal:
Now we'll have only js files in core, I updated them to latest merges.

Todo later:
- Move core pages to js files in the same way
- Check existing core files and remove unused / fill with content files like logo.js, tooltip.js and few more, because they're now empty


DavidRouyer:
Thank you guys, I'll have to do my [PR](https://github.com/DivanteLtd/vue-storefront/pull/1146) again :sweat_smile:
Can you remove the corePage() method in the same way?

PR #1170, opened 2018-05-18
Remove corePage() method

DavidRouyer:
Follow-up of https://github.com/DivanteLtd/vue-storefront/pull/1168

### DIFF
--- a/core/lib/themes.js
+++ b/core/lib/themes.js
@@ -35,10 +35,6 @@ export function filters () {
   return _.merge(themeFilters, coreFilters)
 }
 
-export function corePage (path) {
-  return require('core/pages/' + path + '.vue') // using webpack path alias  - core/pages = core/pages
-}
-
 export function extendStore (coreStore, extendStore) {
   return _.merge(coreStore, extendStore)
 }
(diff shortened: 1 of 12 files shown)


pkarw:
Hi David, Bartek and Damian from snow dog during today’s Hackathon refactored the core components to *.js files because of performance and to make the core cleaner (there were no templates so using *.vue files makes no sense). So it’s good chance to make the ends meet - I mean your change with removing core component and their change which is already merged in :) so May I ask You to pull the code from develop and modify your changes accordingly?

Also would be great to modify the eslint rule to use the *.js components instead of *.vue from core. It’s a cosmetic change because the structure is the same @igloczek

DavidRouyer:
@pkarw I saw their PR, they removed coreComponent() but not corePage(), that's why I created this PR :smiley: I'm merging again the latest changes made in develop!

pkarw:
Oh :) and how about the eslint rule to use this new naming strategy for the core components? It would solve the issue with porting the custom themes. Could you take care of it?

pkarw:
I mean - changing the —fix behaviour

DavidRouyer:
Sure, I can do it in another PR!

pkarw:
Thanks, merged in

PR #1180, opened 2018-05-19
[WIP] Pages to js files

Aekal:


filrak:
I don't think Vue.component() is required to export. Simple js objects can also be treated as mixins

pkarw:
@filrak at this point it’s required as many core components uses for example this.$store, $forceUpdate

To have just plain js classes there would be additional effort for refactoring such a falls required - IMO not doable before 1.0

PR #1184, opened 2018-05-20
[WIP] Core module proposal with cart example

filrak:
This is a new experimental API proposal (under development).

The idea is to group similar functionalities into modules. Also I'll try to split features into smaller standalone features (like add to cart, remove from cart, items in cart, microcart [maybe we should split this one into smaller ones too] etc). it's still meant to be used as mixins.

The implementation of this particular module will (for sure) change . The thing that will not change is the API for using core modules

The proposed API for using it will be as follows.

1. You can import all parts of the module and inject:
````js
import Cart from 'core/modules/cart'
...
mixins: [Cart]
````
2....or just the ones that you need (this particular set would be useful for Product page)
````js
import { addToCart, removeFromCart } from 'core/modules/cart'
...
mixins: [addToCart, removeFromCart]
````


filrak:
Just to clarify - the core pages stays the same. For now we have all functions grouped to components which is more like visual visual grouping rather than functionality grouping. With modules approach you can have the theme structure that suits your needs best and use just the features you need. Another advantage is more abstrakctiin, lighter coupling of functions and having everything related to one functionality in one place.

mercs600:
@pkarw I'm totally agree with @filrak. I prefer this approach also. You can couple selected functions to achieve functionalities which you exactly want for your theme. Currently we are loading complete modules like cart. Users may not know ( be unaware) about all functionalities from selected modules. Sometimes you don't need all functions from selected module. Other benefit, smaller pieces of code are easier to refactor or unit testing.  From my perspective this is good direction.

pkarw:
Guys how it’s related to vuex modules then?

filrak:
@pkarw it's not related in any intended way, but vuex modules are also grouping functions in a similar way (in most cases) so there is a high chance that most of them will have their twin in core modules

filrak:
One thought more - I think  it will be better to use methods and data as objects related to modules (so we will use module.cart.add() instead of addToCart() in out templates). With this approach we can instantly see what methods and data properties are coming from which modules and which methods are coming from modules, which from the component itself. It will make debugging and reasoning about the application much easier. It will also prevent some naming conflicts. Ofc we can force it

Issue #1194, opened 2018-05-22
Images don't refresh after filter change 


mercs600:
Ok I was working on #1104 and I noticed strange behaviour of images loading after filter change. You can see it on video below.  https://www.dropbox.com/s/usl63n623k9vvdi/v-lazy-bug.mov?dl=0 
1. Open some category
2. Change the color filter 
3. Some pictures are not refreshed. 
I think this bug is related with v-lazy component plugin. When we changed directive v-lazy to v-lazy-container this plugin don't recognize changes on data-src. https://github.com/DivanteLtd/vue-storefront/commit/30ad04658334ddbc9063d94e2b3c706eb15756e6#diff-b9fff0db01c53261f63f7aee2c4e8459L33 

I have to fix it before, because I want to do some filters refactor related with #1104

mercs600:
I think we can close it now ;-)

PR #1195, opened 2018-05-22
[#1194] Product thumbnail refresh after filter change

mercs600:
fixes issue described in #1194 

@Igloczek check it please

### DIFF
--- a/core/components/ProductTile.js
+++ b/core/components/ProductTile.js
@@ -4,11 +4,6 @@ import imgPlaceholder from 'core/directives/imgPlaceholder'
 export default {
   name: 'ProductTile',
   props: {
-    instant: {
-      type: Boolean,
-      required: false,
-      default: () => false
-    },
     labelsActive: {
       type: Boolean,
       requred: false,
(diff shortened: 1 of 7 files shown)


pkarw:
Its great you fixed it! My fix for that was to use instant flag to disallow lazy loading for first X pictures (it’s in the develop); but would be better to have lazy loading for all objects

Igloczek:
I wrote lots of words in this box, but then found a solution and had to remove everything :D

Now we are using mix of `v-lazy` and `v-lazy-container`, which seems to be redundant and kidna confusing solution.

But we can do it better :)
`v-lazy` [can take object as a value](https://github.com/hilongjw/vue-lazyload#basic), so if we will create `computed` that returns object with all necessary URLs, we can reduce code to this:

```js
imgObj () {
  return {
    src: thumbnail,
    loading: placeholder,
    error: placeholder
  }
}
```

```vue
<div v-if="!instant">
  <img
    :src="placeholder"
    :alt="product.name"
    v-lazy="imgObj"
    height="300"
    width="310"
  >
</div>
```

pkarw:
Seems quite good. Does it work with the filter changes as it's expected?

Igloczek:
I didn't try it.

pkarw:
Could You, please? :D @Igloczek

Igloczek:
RODO vs Free time
500:0

pkarw:
:D OK, I'll test it later then or @mercs600 please do If you could

mercs600:
@Igloczek @pkarw it's done

pkarw:
@mercs600 please just fix the lint error and I'm merging it in:
https://travis-ci.org/DivanteLtd/vue-storefront/jobs/383323726

mercs600:
@pkarw I missed it. It's done. Right now I noticed some other bugs on catalog theme, so I will fix it.

PR #1250, opened 2018-06-02
Webpack config improvements

DavidRouyer:


pkarw:
@DavidRouyer could You please check the build error - https://travis-ci.org/DivanteLtd/vue-storefront/jobs/387045204
?

DavidRouyer:
It seems the config.json file is not generated in core\build :/

pkarw:
Ok, I’ll check it and merge in after you fix the issue with generating build/config.json (however I don’t see at first glance how these changes could broke it :)). For sure these changes are a way to go

pkarw:
ElasticSearch is used just in the core/lib/search.js and we’re using very basic stuff - just sending out the query/getting back the results. As we’re using bodybuilder to create elastic query (https://bodybuilder.js.org/docs/) You can probably just remove the ElasticSearch and user fetch() instead

filrak:
@DavidRouyer how the things are going with this task? ;)

filrak:
@DavidRouyer can you change the target branch to develop?

Issue #1269, opened 2018-06-08
Check core components


Aekal:
We should check existing core components and move some logic from/to theme. 
At the moment we have some empty files in core. It'll be nice to check each file and decide which logic should be placed in core and which in theme.
We can also clean up some code, remove unnecessary files etc.

pkarw:
@aekal Please remove {{ multistoreEnabled }} snippet from filter which causes displaying ... true 
![0cd7e098-3d22-48a3-addd-dd8c7d58b3a3](https://user-images.githubusercontent.com/211899/41174892-a1878302-6b5b-11e8-833a-e4407aaff7c1.jpeg)

Aekal:
@pkarw I forgot to remove one line for debugging purposes :P

pkarw:
Good job!

PR #1270, opened 2018-06-08
Feature/1269

Aekal:
#1269 
Removed some empty core files, moved base logic from theme components to core


pkarw:
Thanks! This makes a lot of sense!

PR #1272, opened 2018-06-08
Feature/1269 Fix

### LINKED ISSUE #1269: Check core components

Aekal:
#1269


pkarw:
No problem, thanks :)

PR #1280, opened 2018-06-11
Add autoprefixer to prefix css for older browsers and postcss flexbox fixes plugin

marlass:
Some browser incompatibilities issues should be solved with just adding prefixes and use plugin for automatic flexbox fixes. Autoprefixer and postcss plugin should take care of that. In package.json you can adjust browserlist config to targeted browsers. I just settled with modified defaults for production build and simple set for newest browsers for development.


filrak:
<3

PR #1297, opened 2018-06-14
Fix for _filterChildrenByStockitem and setCurrent

pkarw:


pkarw:
Please make sure, that You have the `products.filterUnavailableVariants=true`, otherwise this code has no effect at all.

PR #1325, opened 2018-06-20
[WIP] Cms integration

filrak:
CMS integration for Vue Storefront.
I'm starting with abstract boilerplate for generic use which will be combined with extensions for each platform (eg contentful, wordpress etc).

The extensions are developed on separate repos

### DIFF
--- a/core/api/content/getContent.js
+++ b/core/api/content/getContent.js
@@ -1,0 +1,1 @@
+export const getContent = {}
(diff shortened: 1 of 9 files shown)


pkarw:
I got the idea, for me it's just fine. Maybe we should cache the blocks in the indexedDb to not query the server each time? if there will be external API call for getting the blocks as I assume is the case here - it won't be cached by default (only Elastic calls are cached by default). Please consider it.

Some checks are failing - please fix it before the merge.

filrak:
@DavidRouyer I'll adjust my PR to current magento cms integration which is using es :)

PR #1328, opened 2018-06-21
1142 add cms data extension

anqaka:


filrak:
@anqaka please just fix the conflicts and we are ready to merge

src/extensions/cms/store.js
```diff
@@ -0,0 +1,59 @@
...
+    return state.cmsPages.find(item => item.id === id)
+  }
+}
+
+// actions
+const actions = {
+  loadCms (context, {id, type}) {
+    let url = (config.cms.endpoint)
+      .replace('{{type}}', type)
+      .replace(`{{cmsId}}`, id)
```
szafran89: IMO there is no need to use template strings. You can just use single quotes `'{{cmsId}}'`

anqaka: corrected

anqaka:
@filrak I made one component to display page and block, we have different endpoints but I use `type` prop to deal with it. Conflicts fixed. We need Magento module as a composer package to get the parsed content from Magento API, it will be released probably on Monday.

filrak:
I'll wait with merge until we can test with our demo instance with Magento module ;)

anqaka:
@filrak the Magento modul is available as a composer package here: https://github.com/SnowdogApps/magento2-cms-api please check if everything works.

pkarw:
Module installed on the demo-magento2.vuestorefront.io; @filrak please test this extension - it should work now

src/extensions/cms/store.js
```diff
@@ -0,0 +1,59 @@
...
+import fetch from 'isomorphic-fetch'
+
+const state = {
+  cmsPages: [],
+  cmsBlock: []
+}
+
+const getters = {
+  getBlock: (state) => (id) => {
+    return state.cmsBlock.find(item => item.id === id)
```
pkarw: Could You add the in-browser cache for the static content to improve the performance? we can use kind of localForage sync like we have in the shopping cart https://github.com/DivanteLtd/vue-storefront/blob/6be64960d9dc5e728fcdda75068f80d8dbc8436a/core/store/index.js#L59

Or, explicitly call `setItem` and `getItem` on some newly created localForage collection (please add the localForage collection in here: https://github.com/DivanteLtd/vue-storefront/blob/cf49b4c85bb562a55e475d1d7ee33c3e36ab8dfd/core/store/index.js#L143

src/extensions/index.js
```diff
@@ -8,5 +8,6 @@ export default [
   require('@vue-storefront/extension-payment-backend-methods/index.js'),
   require('@vue-storefront/extension-payment-cash-on-delivery/index.js'),
   require('@vue-storefront/extension-template/index.js'),
-  require('vsf-payment-stripe/index.js')
+  require('vsf-payment-stripe/index.js'),
+  require('src/extensions/cms/index.js')
```
pkarw: @anqaka if we like to register this extension by default then we should have the `config.cms` section added to `default.json`

src/themes/default/pages/Home.vue
```diff
@@ -26,6 +26,13 @@
       <tile-links />
     </section>
     <Onboard/>
+
+    <cms-data :id="5" :type="'Block'" />
```
pkarw: Can we add the second way to reference blocks + pages - by url_key?

src/extensions/index.js
pkarw: If we add it to the default.json, then we should modify the installer script to populate this endpoint config as well:
https://github.com/DivanteLtd/vue-storefront/blob/cf49b4c85bb562a55e475d1d7ee33c3e36ab8dfd/core/scripts/installer.js#L371

src/extensions/cms/store.js
anqaka: added

src/extensions/index.js
anqaka: moved to default.json and add to installer

pkarw:
OK, last thing! Could You please add some `.catch()` for fetching the content and add `console.error('You need to install a custom Magento module from Snow.dog to make the CMS magick happen. Please go to https://github.com/SnowdogApps/magento2-cms-api and follow the instructions')` OK? :)

danrcoull:
> OK, last thing! Could You please add some `.catch()` for fetching the content and add `console.error('You need to install a custom Magento module from Snow.dog to make the CMS magick happen. Please go to https://github.com/SnowdogApps/magento2-cms-api and follow the instructions')` OK? :)

A issue with the cms extension as getters are triggering before the data is returned and have no access to the state, so just throws exceptions. 

" Error in render: "TypeError: state.cmsPage is undefined""   but also the data returns yet the catch stated above is always triggered even when that module is isntalled.

Issue #1274, opened 2018-06-09
Add more integration tests


pkarw:
We already have a Cypress framework integrated from the last hackathon (thanks to @marlass) and sole example test created: https://github.com/DivanteLtd/vue-storefront/blob/8fc6235fb7c24bb3d538e275fcdefdf349115b82/cypress/integration/basic-client-path.js

Would be great to add some more e2e tests, covering:

- [x] Login
- [x] Register
- [x] Configure product (select variant check if the right product has been added)
- [x] Filter products on category page check results
- [x] Change default shipping/billing address
- [x] Add bundle product to the cart
- [x] Search
- ...

PR #1340, opened 2018-06-22
Add login and register e2e tests

Aekal:
#1274

### DIFF
--- a/cypress/integration/login-path.js
+++ b/cypress/integration/login-path.js
@@ -1,0 +1,14 @@
+describe('login path', () => {
+  it('should login user', () => {
+    cy.visit('/');
+    cy.get('.modal-close').click();
+    indexedDB.deleteDatabase('shop');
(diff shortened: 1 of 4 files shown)


pkarw:
@Aekal i belive this test will fail the second time You run it (email in use) :)

pkarw:
it should rather test some validation (to show, not show error) and to not modify the magento2 data (by registering account); please remember that tests should be executed before any PR

Aekal:
I'm checking only validation, if errors won't show up test will succeed - it works each time I run it ;) But you're right, it shoudl't pass any additional data to magento, so I'll remove form submits

pkarw:
OK, lets just not create the real account

Aekal:
I disabled form submits, so data won't be send. Also added additional validations on focusing out input - now validation will work similar to e.g PersonalDetails in Checkout

PR #1422, opened 2018-07-11
Integration test for compare products

patrykpiston:
I had some time so I thought I'd add a test for comparing products.


cypress/integration/add-to-compare.js
```diff
@@ -0,0 +1,19 @@
+describe('add to compare', () => {
+  it('Two products should be added to comparison table', () => {
+    cy.visit('/c/jackets-23')
+    cy.setCookie('shop/claims/onboardingAccepted', 'test')
+    indexedDB.deleteDatabase('shop')
+    cy.clearLocalStorage()
+    cy.get(':nth-child(1) > .product > .no-underline > .product-image > img').click()
```
Aekal: Please don't use css selector like this, it's very easy to break this test just by changing one class... 
You can check [my issue](https://github.com/DivanteLtd/vue-storefront/issues/1377), I also did a lot of refactoring similar code in [last PR](https://github.com/DivanteLtd/vue-storefront/pull/1419/files).

patrykpiston: Changes made in https://github.com/DivanteLtd/vue-storefront/pull/1422/commits/cbdfafb3b62f326ede36e014be4dae568518cea9

PR #1436, opened 2018-07-18
Price tiers

mdanilowicz:


core/store/lib/search.js
```diff
@@ -45,15 +46,25 @@ function search (elasticQuery) {
...
-  return fetch(url, { method: 'POST',
-    mode: 'cors',
-    headers: {
-      'Accept': 'application/json',
-      'Content-Type': 'application/json'
-    },
-    body: JSON.stringify(elasticQuery.body)
-  }).then(resp => { return resp.json() })
+  return new Promise((resolve, reject) => {
+    global.$VS.db.usersCollection.getItem('current-refresh-token', (err, refreshToken) => {
```
pkarw: I think that we shouldn't use refresh token for passing the group id by two reasons:
- first is that it contains username + password (ofc. encoded, but we shouldn't transfer it unecessarely via net)
- second is that it should be used JUST FOR refreshing the tokens, and by using it also to send user group data we're violating SRP rule (single responsibility principle)

Because after user is logged in there is always a call to user/me executed; maybe we should just add a JWT token to the user/me method (server side generated) containing username + groupid + userid and encrypted server side; Then You could use this token (stored client side with the user profile as it will be just another user profile field)

I mean adding kind of: "token" field to what /user/me is returning

pkarw:
@thomasthune I think this is good point

PR #1446, opened 2018-07-20
[WIP] Enable e2e tests in CI

Aekal:


Aekal:
When I run `yarn start && yarn tests:e2e:ci` on my local all tests pass in less than 1:30 min

zrzut ekranu 2018-07-20 o 14 54 54

But `yarn start` doesn't work in CI, is there any way to make this work also there with pm2?
I tried to run tests after `yarn installer:ci` but I get many of browser and memory errors (https://travis-ci.org/DivanteLtd/vue-storefront/jobs/406221380).

pkarw:
@Aekal there was a problem with that and that's the reason it was commented out :( I think it's a problem with resource limitations (memory?) in the Travis CI.

I think we should still keep it commented out, but nevermind: great job with the test improvements!

Issue #1426, opened 2018-07-16
CMS Integration - add cms data by idntifier


anqaka:
add feature to cms-data extension: Get CmsBlock & CmsPage by Magento `identifier`
Magento API module:  https://github.com/SnowdogApps/magento2-cms-api/pull/2

PR #1452, opened 2018-07-23
Feature/1426

anqaka:
#1426 
this API in vue-storefront-api repo needed:
https://github.com/DivanteLtd/vue-storefront-api/pull/74
and SnowdogApps/magento2-cms-api version 1.1.0 need to be installed in Magento 2 instance


pkarw:
Am I right that currently you’re downloading all the cms content to the state and then get the specific page block from it? That’s good

I think it would be best if the Vue-storefont-api could use the elasticsearchjs to index this content while downloading it from Magento to elasticsearch index, page / block entities


By having this lazy loaded elastic cache the app can provide the search in cms feature as a next step

Could you take care of that?

anqaka:
@pkarw yup, for now the cms data are stored in state. I'll take care of love it with elasticsearch but can't handle it this week for sure. I'll do it asap.

PR #1453, opened 2018-07-24
Fix eslint and unit test errors

Aekal:


Aekal:
@pkarw ESLint and unit tests in CI are now fixed, I also moved all scripts to one stage (works faster - currently 4~ min).
Do we need `yarn build` in CI? Is it used for autodeployment somehow or it's unnecessery here?

PR #1463, opened 2018-07-27
Replacing select with base-select component in MyProfile component

Aylay:
The base-select component was created in release 1.1.0.
So I think that it's better to use the component instead of `<div><select>...`

What do you think ?

There are the changes in **MyProfile component** : 
- Replacing select with base-select component 
- Importing base-select component
- Adding countryOptions function in computed


pkarw:
Thanks, have You tested it? :)

Aylay:
@pkarw Yes of course.

There is the list of the countries displayed :)

PR #1467, opened 2018-07-30
Bugfix - Reorganize taxcalc.js; cleanup float conversion

thomasthune:
Makes code more readable as well as fixes bug where special_price tax is not added to the product.

Flip priceTax and ProductInclTax calculation as to only calculate price tax once. Use const rateFactor instread of calculating it each time.

Example
```
  product.priceInclTax = (priceExclTax + priceExclTax * (parseFloat(rate.rate) / 100))
  product.priceTax = (priceExclTax * (parseFloat(rate.rate) / 100))
```

becomes 

```
  product.priceTax = priceExclTax * rateFactor
  product.priceInclTax = priceExclTax + product.priceTax
```

### DIFF
--- a/core/store/lib/taxcalc.js
+++ b/core/store/lib/taxcalc.js
@@ -1,24 +1,25 @@
 export function updateProductPrices (product, rate, sourcePriceInclTax = false) {
+  const rateFactor = parseFloat(rate.rate) / 100
   product.price = parseFloat(product.price)
+  product.special_price = parseFloat(product.special_price)
 
   let priceExclTax = product.price
   if (sourcePriceInclTax) {
-    priceExclTax = product.price / (1 + (rate.rate / 100))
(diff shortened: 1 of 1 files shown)


pkarw:
Could You please fix the eslint issues: https://travis-ci.org/DivanteLtd/vue-storefront/builds/409840203?utm_source=github_status&utm_medium=notification ?

thomasthune:
Will do, didn't get the error locally though 🤔 
About removing the extra ```parseFloat()```; price, special_price and the rate, are at the only 3 numbers that are actually strings (as far as i can see). Converting them to ```number``` means that the rest of whats happening is math involving numbers, and hence any extra ```parseFloat()``` would be superfluous.
Please correct me if I'm wrong. 😄 

Will push update for linting during the day.

pkarw:
Thanks! That make sense, thanks for clarifying

Issue #1286, opened 2018-06-12
JS Bundles are too large for mobile


slightlyoff:
Howdy!

This project was brought to my attention by @tomayac, @ devnook, and @rowan-m.

Running the [live demo](https://demo.vuestorefront.io/) through [Web Page Test unearthed some surprises](https://www.webpagetest.org/result/180612_2W_282b204cdbfb9ff26d9ea34730770892/).

One of the most concerning issues is the size of the JS bundles. The [Content Breakdown view](https://www.webpagetest.org/result/180612_2W_282b204cdbfb9ff26d9ea34730770892/1/breakdown/) shows ~750KB of JS being served, which unzips to more than 2MB of code:

![screen shot 2018-06-12 at 10 36 20 am](https://user-images.githubusercontent.com/97331/41306910-7cd04bba-6e2c-11e8-8667-8bb9ec153ee1.png)

By comparison, the [Polymer Shop demo](https://www.webpagetest.org/result/180612_X7_f8adf5e327d22386aff811347a13a2f4/) -- an app of similar complexity which takes advantage of route-based code-splitting and granular loading --  [taken on the same HW and network configuration](https://www.webpagetest.org/result/180612_X7_f8adf5e327d22386aff811347a13a2f4/) confuses chrome devtools slightly, but [by inspection only needs ~200K of script to load a fully-functioning experience](https://www.webpagetest.org/result/180612_X7_f8adf5e327d22386aff811347a13a2f4/2/details/#waterfall_view_step1).

The Vue demo is more than 3x larger and is unlikely to perform well on a low-end mobile device as a result.

pkarw:
Thanks for Your observation; this is one of the reasons we started working on https://github.com/DivanteLtd/vue-storefront/pull/1210

It’s still Work In Progress but You can test it if You like / have some free time. @DavidRouyer is working on this one and it probably solve this issue

Any support is appreciated, we’re open source, and the biggest challenge is always to find as much time for the developement and optimization’s as we would like to have :-)

pkarw:
We made a first step with removing two huuuge packages (#1293, #1294) / it gave us -40% in size

tomayac:
Brilliant progress! Tree shaking is good, but a common sense human audit is priceless!

PR #1479, opened 2018-08-01
feat: bundle optimizations

DavidRouyer:
Part of #1286. I'm using webpack-bundle-analyzer.

Before:
before

After:
after

Removed the duplicated vue.runtime.esm.js


pkarw:
Thanks @DavidRouyer !

Issue #1494, opened 2018-08-03
Change all online/offline dependencys and network status checks to use vue-offline


filrak:
Change all online/offline dependencys and network status checks to use vue-offline.

PR #1502, opened 2018-08-03
[#1494] Change event bus to vue-offline mixin

mercs600:
@filrak Generally done, but I'm not sure here, should we remove this emit from client-entry ? https://github.com/DivanteLtd/vue-storefront/compare/develop...mercs600:feature/1494?expand=1#diff-83c5dc07ca8885f0200ed709394dbbc3L285


mercs600:
@filrak I removed isOnline from microcart.js

PR #1511, opened 2018-08-03
Feature/849 Add dynamic port allocation 

kruchy8:
Solved with simple recurring function & incrementation of port number.


pkarw:
Please add the --open option to yarn dev which will open the browser

pkarw:
https://stackoverflow.com/questions/7664605/how-would-you-launch-a-browser-from-the-a-node-js-command-line-script

PR #1543, opened 2018-08-07
Fix/typescript config

NataliaTepluhina:
Based on to [#1538](https://github.com/DivanteLtd/vue-storefront/pull/1538) with some little fixes:

- added aliases to `tsconfig.json` to handle file imports nicely and prevent IDE error highlighting on imports;
- fixed `ts` extension in webpack config to be `.ts`
- added all previously reverted changes


filrak:
Thank you very much, the only thing left is to change aliases for `theme` to `theme/themename` but I'll do a separate issues for this.

PR #1556, opened 2018-08-08
chore: cleanup & give TypeScript superpowers

DavidRouyer:


pkarw:
This is fine. Please just add a note to CHANGELOG or UPGRADE NOTES about this change; it should not cause any backward compatibility issues but just in case let's keep it clearly stated in the docs

PR #1558, opened 2018-08-08
feat: new folder structure

DavidRouyer:
- moved core files from `/core` to `/src/core`
- moved store files from `/core/store` to `/src/store`
- moved build files from `/core/build` to `/build`
- moved script files from `/core/scripts` to `/scripts`
- moved `/core/server-entry.js` to `/src/server-entry.js`
- moved `/core/client-entry.js` to `/src/client-entry.js`


pkarw:
@DavidRouyer some time ago we’ve agreed that /src folder is for users to put the themes and extensions on there and /core is for us for ... maintaining the core :)

In this manner - we should have /core and /store on the same level - but it’s rather one level higher (not in /src).

We currently have core and store as a separate npm modules so we can create kind of project boilerplate (empty project just with the modules imported - template for eCommerce builders) - that was an idea from @igloczek and @filrak that came across during last Hackathon 

I think that this convention - src is for user, other folders are for framework devs is good, easy to understand.

@Igloczek @filrak thoughts?

I agree with the rest of changes in this PR

pkarw:
We must vote and have 100% agreement on this to merge it in guys :)

filrak:
I'm also against. We would lost all external linka in tutorials, videos will not be valid any more, people will be confused etc. I also see no point in moving this folders conaidering our new modular approach. I don't want to kill enthusiasm but just now this change feels wróg especially that it's introducing a lot of changes when more of them are waiting to introduce

Igloczek:
IMO it's a good change, will be way cleaner and simpler to extend in the future. Also helps in moving to smaller modules, like, there is no point to keep `store` inside `core`, no point to keep `client-entry` inside `core` etc.

Docs/resources are mostly in the repo, so after updating a few paths, they will still useful. Videos will be outdated soon too, since we should kinda force people to start using a boilerplate, not the core repo, so this stuff will be hidden.

filrak:
I'm also against. We would lost all external linka in tutorials, videos will not be valid any more, people will be confused etc. I also see no point in moving this folders conaidering our new modular approach. I don't want to kill enthusiasm but just now this change feels wróg especially that it's introducing a lot of changes when more of them are waiting to introduce

PR #1568, opened 2018-08-09
feat: Typescript migration

DavidRouyer:
### Related issues

Milestone 14

- Migrate entrypoints to Typescript
- Change the Typescript configuration to allow the use of require(), dynamic imports, reflection...
- Add a few types
- Store attribute as an example for stores

### Contribution rules acceptance


core/app.ts
```diff
@@ -12,17 +12,20 @@ import EventBus from '@vue-storefront/core/plugins/event-bus'
...
 import registerExtensions from '@vue-storefront/core/lib/extensions'
 import i18n from '@vue-storefront/core/lib/i18n'
-import shippingMethods from '@vue-storefront/core/resource/shipping_methods.json'
 
 import store from '@vue-storefront/store'
 import coreModules from '@vue-storefront/store/modules'
 import { prepareStoreView } from '@vue-storefront/store/lib/multistore'
 
-import App from 'theme/App'
+import App from 'theme/App.vue'
```
filrak: shouldn't it be resolved by default without `vue` (look at extensions in webpack config)

core/store/modules/attribute/types/AttributeState.ts
```diff
@@ -0,0 +1,5 @@
+export default interface AttributeState {
```
filrak: Shouldn;t it be just 'Attribute' ?

pkarw: I think it should stay AttributeState as it's vuex state object

filrak:
Regarding the types. We discussed with @pkarw that maybe it'll be better to have as few types as it is possible to allow clear understanding of data structures (eg one/two Product types, not 15 for different purposes etc). It's an open question and I don't feel like TS expert so really want to hear your opinion ;)

pkarw:
Having store states typed has additional benefits of: intelisense support + easier to understand state structure for new comers

DavidRouyer:
Yes, our data are in the vuex stores so it's much better if they have types!

core/store/modules/attribute/types/AttributeState.ts
filrak: But it's also the same object as the one in View

filrak:
I mean we should have type checking available for everything in Vuex, but not sure if we should introduce more than 1-2 types per entity

DavidRouyer:
What do you mean by entity?

filrak:
Product, Category, Attribute etc

DavidRouyer:
The product model (from the Ajax call) and the product state aren't the same thing :)

pkarw:
For me is fine. I'll merge it in Guys. Thanks @DavidRouyer great job!

Issue #1525, opened 2018-08-03
Add PWACompact


filrak:
It'll be great to have this tool included into our app for the better support of various devices and browsers

https://developers.google.com/web/updates/2018/07/pwacompat

PR #1574, opened 2018-08-11
feature/pwa-compact

filrak:
### Related issues

#1525 

Added PWA Compact for better manifest support in older browsers
https://developers.google.com/web/updates/2018/07/pwacompat

- [x] I read about [Vue Storefront Modules](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md) and adjusted my PR according to it (details in contribition rules)


DavidRouyer:
I'm not a fan at all of third-party JS scripts, loaded from an external source, what problem does it solve? Can't we just improve the manifest?

pkarw:
@filrak I think that external script on 3rd party servers will decrease the overall Lighthouse score (as we had with fonts) and it’s in general: risky.

Can we move the script to /assets?

DavidRouyer:
There is a lot of tools like https://www.pwabuilder.com/ to generate a correct manifest, icons, etc., why doing this at runtime with an extra JS script?

filrak:
This one is from Google and aims to deliver backward compatible manifest experience. I think it's much better than figuring it out manuallyx especially when it comes to splash screen. Since it's a tool from Google itself I'd trust it. 
@pkarw this script is loaded async wo it should lower the lh score.

DavidRouyer:
I don't know what they mean by backward compatibility, since PWA capabilities are only available in evergreen browsers and are updated automatically?
If you want to integrate a tool like that, it should be integrated inside the Webpack process no?

filrak:
Let's Say it differently. If someone will add numerous fallbacks and icons to index file and then test out pwa capabilities in all desired browsers then I can delete it. For now it's the easiest way to be sure that it just works everywhere it should. And again it's a small async script, i doesn't affect our performance at all

Igloczek:
The idea of "polyfiling" web app manifest is kinda weird for me, but I can imagine that various Chrome forks on Android or iOS devices might benefit from it.

But I also dislike relying on some CDN, will be better to try to load their npm package and import it locally, although I don't get why they don't recommend this way 🤔

filrak:
I also don;t like the way they force devs to include it but since it's recommended one I'll leave it like this yet I'm open to further improvements.

Since it helps with mobile Safari support a lot I'm merging it in right now but feel free to propose PRs with improvements.

PR #1609, opened 2018-08-20
Fix/1601

vue-kacper:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/1601

Introduced computed properies instead of non-reactive data() properies. Fixed issue with custom product attributes have not been updated correctly when navigating from one product page directly to another product page.

@pkarw  Solution propsed by you doesn't work so I decided to use different approach


vue-kacper:
Sorry, wrong branch - let me fix it quick

PR #1610, opened 2018-08-20
fix/1601 incorrect productinfo

vue-kacper:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/1601

Introduced computed properies instead of non-reactive data() properies. Fixed issue with custom product attributes have not been updated correctly when navigating from one product page directly to another product page.

@pkarw Solution propsed by you doesn't work so I decided to use different approach

### DIFF
--- a/src/themes/default/components/core/ProductAttribute.vue
+++ b/src/themes/default/components/core/ProductAttribute.vue
@@ -9,36 +9,6 @@
 import ProductAttribute from '@vue-storefront/core/components/ProductAttribute'
 
 export default {
-  mixins: [ProductAttribute],
-  computed: {
-    label () {
-      return (this.attribute && this.attribute.default_frontend_label) ? this.attribute.default_frontend_label : ''
-    },
-    value () {
-      let parsedValues = this.product[this.attribute.attribute_code]
-      if (!parsedValues) {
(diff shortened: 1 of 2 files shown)


src/themes/default/components/core/ProductAttribute.vue
```diff
@@ -10,37 +10,34 @@ import ProductAttribute from '@vue-storefront/core/components/ProductAttribute'
...
-      let results = []
-      for (let parsedVal of parsedValues) {
-        if (this.attribute.options) {
-          let option = this.attribute.options.find(av => {
-            /* eslint eqeqeq: "off" */
-            return av.value == parsedVal
-          })
-          if (option) {
-            results.push(option.label)
+  computed: {
```
pkarw: Shouldn’t this logic be just in the core? I’m not sure but I think it’s duplicated in the *.vue component is kind of DRY principle violation.
Shouldn’t it be just left in the core?

vue-kacper: You are absolutelty right, It was originally duplicated - to make sure I'll compare both files and remove duplicated logic if there are no differences.

pkarw:
Thanks!

PR #1625, opened 2018-08-23
Price tiers.

afirlejczyk:
Add user groupToken to state, cache result with groupToken.


core/store/lib/search.js
```diff
@@ -43,27 +44,26 @@ function search (elasticQuery) {
     throw new Error('elasticQuery.index and elasticQuery.type are required arguments for executing ElasticSearch query')
   }
 
+  if (config.usePriceTiers && rootStore.state.user.groupToken) {
+    elasticQuery.body.groupToken = rootStore.state.user.groupToken
+  }
+
+  /* use only for cache */
```
pkarw: This shouldn’t be here - search.js should not now about any specific content of the query. Please move it one level higher - on the hashing level

PR #1658, opened 2018-08-29
Additional entry point for order review section. 

dimasch:
@pkarw Hi, Piotr. This is point for integrate external extensions like Paypal.
\+ @collymore

### DIFF
--- a/core/components/blocks/Checkout/OrderReview.js
+++ b/core/components/blocks/Checkout/OrderReview.js
@@ -11,10 +11,16 @@ export default {
   data () {
     return {
       isFilled: false,
+      placeOrderVisible: true,
       orderReview: {
         terms: false
       }
     }
+  },
+  created () {
+    this.$bus.$on('place-order-button-visible', (val) => {
+      this.placeOrderVisible = val
+    })
   },
   methods: {
     placeOrder () {
--- a/src/themes/default/components/core/blocks/Checkout/OrderReview.vue
+++ b/src/themes/default/components/core/blocks/Checkout/OrderReview.vue
@@ -63,8 +63,10 @@
       <div class="col-xs-12 col-sm-9 col-md-11">
         <div class="row">
           <div class="col-xs-12 col-md-8 px20">
+            <div v-show="!placeOrderVisible" id="place-order-container"/>
             <button-full
               @click.native="placeOrder"
+              v-show="placeOrderVisible"
               data-testid="orderReviewSubmit"
               :class="{ 'button-disabled' : $v.orderReview.$invalid }"
             >


core/components/blocks/Checkout/OrderReview.js
```diff
@@ -11,11 +11,17 @@ export default {
...
     return {
       isFilled: false,
+      placeOrderVisible: true,
       orderReview: {
         terms: false
       }
     }
   },
+  created () {
+    this.$bus.$on('place-order-button-visible', (val) => {
```
pkarw: Please rename it to: checkout-before-placeOrder

pkarw:
The intention is that Paypal will hide "Place Order" and will add "Checkout with Paypal" button right?

core/components/blocks/Checkout/OrderReview.js
dimasch: No, it event for separate goal.

dimasch: @pkarw paypal integration example https://github.com/develodesign/vsf-payment-paypal/pull/41/files

dimasch:
> The intention is that Paypal will hide "Place Order" and will add "Checkout with Paypal" button right?
@pkarw Yes.

dimasch:
Reference pull request https://github.com/develodesign/vsf-payment-paypal/pull/41/files

core/components/blocks/Checkout/OrderReview.js
dimasch: It solve integration with PayPal and for other payment methods it works out of the box.
If we figure out how to improve - it will be cool.

pkarw:
Place Order button should be in kind of slot which is then replaced by payment extension or populated with default button if any extension is not modifying it. I think that the simplest solution is to automatically hide this button without any additional event in case where the container content length>0 (some extension added it’s own button); the container should be cleared each time user starts the checkout 

Does it make sense? Would be much better than this very specific event 

I can merge this as it is but this solution is not quite elegant, I don’t like it much to be honest it’s too specific

src/themes/default/components/core/blocks/Checkout/OrderReview.vue
```diff
@@ -63,8 +63,10 @@
       <div class="col-xs-12 col-sm-9 col-md-11">
         <div class="row">
           <div class="col-xs-12 col-md-8 px20">
+            <div v-show="!placeOrderVisible" id="place-order-container"/>
```
pkarw: This div could be visible all the time but empty / cleared each time when user starts the checkout. Then if some method will add content to i (for example the PayPal button), default button should be hidden automatically

dimasch:
@pkarw Ok, i try to improve with use slot concept.

pkarw:
@dimasch any updates?

dimasch:
@pkarw For me need today evening or tomorrow day for try to improve.

dimasch:
Miss close

src/themes/default/components/core/blocks/Checkout/OrderReview.vue
```diff
@@ -63,13 +63,16 @@
...
         <div class="row">
           <div class="col-xs-12 col-md-8 px20">
-            <button-full
-              @click.native="placeOrder"
-              data-testid="orderReviewSubmit"
-              :class="{ 'button-disabled' : $v.orderReview.$invalid }"
-            >
-              {{ $t('Place the order') }}
-            </button-full>
+            <slot name="placeOrderButton">
```
dimasch: \+ @pkarw Updated pull with slot and additional class instead global event.
Extension pull https://github.com/develodesign/vsf-payment-paypal/pull/42/files
\+ @collymore

dimasch:
@pkarw Ready. Have a nice weekends!

PR #1674, opened 2018-08-31
Feature/dynamic shipping state

benbenek:

As discussed before via Slack the current implementation of the default shipping state uses a seperate JSON-File and overrides custom state definitions in the theme because it gets defined ind app.ts/app.js...

In this PR the shipping configration has been moved to the global configuration file (default.json) and the default shipping state is getting defined directly in the vuex part of the core files which won't break the overriding feature for themes...


pkarw:
Thanks, it's much better approach!

Issue #1520, opened 2018-08-03
incorrect shipping carrier code 


mercs600:
Right now when we place the order we have the same values in method_code and carrier_code, what is not correct for each shipping method 
```js
shipping_method_code: this.shipping.shippingMethod,
shipping_carrier_code: this.shipping.shippingMethod
```

PR #1689, opened 2018-09-03
[#1520] Fix shipping carrier code assignment

mercs600:
### Related issues

#1520 

Fixed carrier code assignment on default to avoid event emit about shipping change. It should not be required on mounted shipping component. 

### DIFF
--- a/core/components/blocks/Checkout/Shipping.js
+++ b/core/components/blocks/Checkout/Shipping.js
@@ -54,9 +54,8 @@ export default {
         shipping = this.shippingMethods[0]
       }
       this.shipping.shippingMethod = shipping.method_code
-      this.shipping.shippingCarier = shipping.carrier_code
+      this.shipping.shippingCarrier = shipping.carrier_code
     }
-    this.changeShippingMethod()
   },
   methods: {
     sendDataToCheckout () {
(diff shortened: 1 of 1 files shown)


pkarw:
Adam, ive added this checkout-after-shippingMethodChanged emoter because without it in the Checkout.js no this.shipping is set to the default shipping method

If user is not selecting any other shipping method but using default then this field is empty what causes putting null value into order object and sends it with null to server

Please test it out - if the object sent to server does have the shipping carrier and method set when user was not switching the shipping method

mercs600:
@pkarw Yes I've noticed and tested it today. Now you should have shipping method and carrier code without emit by event bus. Sorry for the confusion ;-)

pkarw:
Are You 100% sure? 

Please note:
- https://github.com/DivanteLtd/vue-storefront/blob/cd7ebcdd934b6dc4855287de2907f73548fd6f68/core/pages/Checkout.js#L140 - this.shippingMethod is set just by that event that You've removed
- then `this.shippingMethod` is being used when sending the order to the server: https://github.com/DivanteLtd/vue-storefront/blob/cd7ebcdd934b6dc4855287de2907f73548fd6f68/core/pages/Checkout.js#L316

Sorry for bothering You :) Probably I'm not getting it right; please just double check

Issue #1649, opened 2018-08-29
[default] filters are not updating


vue-kacper:
# Steps to reproduce:

- Visit https://demo.vuestorefront.io
- Refresh browser window
- Select category Women -> Tops -> Jackets
- Click on Breadcrumbs "Tops"

![filters](https://user-images.githubusercontent.com/42430550/44771229-b8cb1480-ab6a-11e8-90ba-6fda10988f8c.PNG)

# Expected result:

Filters on the left sidebar should update accordingly without need for page refresh

# Actual result:

Filters are not updating, page refresh is needed to see filters current state

pkarw:
Good point

PR #1750, opened 2018-09-20
Filters are not updating - issue #1649

vue-kacper:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/1649

Fixed reactivity so filters are updating correctly


DavidRouyer:
The store should never be mutated directly, we should always use `Vue.set()` @filrak?

Issue #1733, opened 2018-09-17
Catalog view infinite scroll not working on mobile 


jorkvist:
* [bug] 
  - Version: 
  - Browser: Firefox 62.0.1 | Mobile Chrome 69.0.3497.91

Products lazyload (infinite scroll) is broken on mobile, when scrolling to the bottom, no new products is loading. Tested on my live site but also on https://demo.vuestorefront.io/

(bonus, the counter is not working always says 50 items)

pkarw:
@mdanilowicz @vue-kacper or @qiqqq could you please take a look  at this one, please?

PR #1755, opened 2018-09-22
Issue #1733 Fixed catalog view infinite scroll on mobile browsers

vue-kacper:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/1733

Catalog view infinite scroll fix for mobile browsers


vue-kacper:
Tested on Android mobile Chrome and Firefox and works fine. We need to test it on iOS mobile Chrome.

BrowserStack doesn't allow me to test localhost URLs on iOS devices

Issue #1713, opened 2018-09-08
Add AMP renderer


pkarw:
* **What is the motivation / use case for changing the behavior or adding new 
AMP version of the site can bring lot of improvements to the eCommerce SEO discover ability and performance.

However, there is no external JS allowed in AMplified pages. So probably we need to add the alternative renderer to generate just the AMP product and category pages and then distribute the traffic to existing VS pages

Regarding the docs:
In order for Google (and other technologies supporting the AMP Project) to detect the AMP version of your article, you will need to modify the original version of the article page. The original article page must include the following tag, essentially a canonical tag for AMP pages:

link rel="amphtml" href="http://www.example.com/blog-post/amp/"

This is how Wordpress official AMP plugin works

* **Acceptance criteria**
- [ ] when there is a query parameter ?amp=1 we should use different index.html template and render the plain amp html version of the Product and Category pages

Issue #1688, opened 2018-09-03
Add an nice-looking error page


pkarw:
In case of internal server error (for example ElasticSearch is not responding) we're currently displaying the very sad-looking message:

https://www.dropbox.com/s/rue8h3n3sur85bc/Screenshot%202018-09-03%2015.18.22.png?dl=0

We should create a kind of special error page based on the [PageNotFound](https://github.com/DivanteLtd/vue-storefront/blob/master/core/pages/PageNotFound.js). 

This page should:
- [ ] not contain dynamic data get from Elastic (for example categories)
- [ ] contain just the header + footer + styles

After creating such a page (and adding it to the routing) You need to add the redirect in case of error in here:
https://github.com/DivanteLtd/vue-storefront/blob/cd7ebcdd934b6dc4855287de2907f73548fd6f68/core/scripts/server.js#L81

mercs600:
@pkarw @filrak To achieve error page without any async data we have to create new layout page without components which can use potential connection to some server service. 

So we have to implement layout system, because right now App.vue use the same components for each route path

```vue
<template>
  <div id="app">
    <overlay v-if="overlayActive"/>
    <loader/>
    <div id="viewport" class="w-100 relative">
      <microcart/>
      <search-panel/>
      <wishlist/>
      <sidebar-menu/>
      <main-header/>
      <router-view/>
      <main-footer/>
      <notification/>
      <sign-up/>
      <newsletter-popup/>
      <cookie-notification/>
      <offline-badge/>
      <modal-switcher/>
      <order-confirmation :orders-data="ordersData"/>
    </div>
    <vue-progress-bar />
  </div>
</template>
```

So I created POC of layout components with default and plain template. You can pass in router/path setting what layout you want to use for specific path. 

```vue
<template>
  <div id="app">
    <overlay v-if="overlayActive"/>
    <loader/>
    <div id="viewport" class="w-100 relative">
      <component :is="layout">
        <router-view/>
      </component>
    </div>
    <vue-progress-bar/>
  </div>
</template>
```

Check it please and let me know what do you think, maybe do you have better idea ? -> https://github.com/mercs600/vue-storefront/commit/d19e4c4c0ffdc5c271d78d78388d58786115901a

pkarw:
@filrak 
I don't like this solution as it's quite a massive change just for single error page. Maybe we can create just a plain HTML error page instead ...

pkarw:
[...]

PR #1787, opened 2018-09-27
Layout + template switching mechanism

pkarw:
### Related issues

#1713, #1688

I've changed the way SSR pages are generated:
- now distinct routes can set "context.serverOutputTemplate" in `asyncData` and by doing so You can skip using `dist/index.html` (which contains typical HTML5 elements - like `<head` ...). This is important when we're going to generate AMPHTML pages (that can not contain any `<script>` tags) - either xml files - You name it
- distinct routes can set `meta.layout` and by doing so switch the previously constant `App.vue` layout file - this is what @mercs600 proposed some time ago in slightly changed form

The `themes/default/index.template.html` and `src/index.template.html` template files have been modified by adding `renderStyles()`, `renderState()` helpers. AFAIK nobody is modifying these templates - but if sbd. did - he/she must change it accordingly otherwise styles won't be rendered properly


src/extensions/raw-output-example/pages/RawOutputExample.vue
```diff
@@ -0,0 +1,30 @@
...
+  <raw-content>
+    This page is using empty layout set in routes + no html template
+  </raw-content>
+</template>
+
+<script>
+export default {
+  name: 'RawOutputExample',
+  asyncData ({ store, route, context }) {
+    context.response.setHeader('Content-Type', 'text/xml')
```
pkarw: Here You can now set from the Vue component level server response headers + choose template

src/extensions/raw-output-example/router.js
```diff
@@ -0,0 +1,5 @@
+import RawOutputExample from './pages/RawOutputExample.vue'
+
+export default [
+  { path: '/raw-output-example.xml', component: RawOutputExample, serverOutputTemplate: '', meta: { layout: 'empty' } }
```
pkarw: You can now also set the layout from the routes level

filrak:
Just one question regarding this one
"distinct routes can set meta.layout and by doing so switch the previously constant App.vue layout file - this is what @mercs600 proposed some time ago in slightly changed form"

Shouldn't we do it in a vue way just by creating one component inside `App.vue` that is switchable. I think the proboem solved by this solution is default-theme specific. This problem will no longer occur on new default theme since i wanted to make layouts system there (like in nuxt.js, they had the same problem and made it in vue way).

filrak:
Also, can we add posibility to server side render onlyc ertain pages (i know it's not connected but still alout SSR, there was an issue for this0. This would prevent a lot of SSR errors that we have right now in pages that just don't need to use it.

pkarw:
Filip its done this way - with one component which is switchable inside - App.vue? I didn’t get this point?

pkarw:
We can theoretically filter the urls from SSR routing - however I don’t see many use cases for that. How wupuld you like to use it?

filrak:
basically it's like this:
App.vue has 'layout' data property and dynamic component that renders appropiate compoennt based on value of `layout` data prop. Also we just HAVE 'layouts' folder with layouts to render in this dynamic component slot. That's all, few lines of code'

Abstract of the solution:
```<template>
  <component :is="layout" />
</template>
import { LayoutA } from './layouts/A.vue'
import { LayoutA } from './layouts/B.vue'

data: {
  return {
    layout: layoutA // will be a  computed property taking layout to render from vuex so it can be set in vuex in the app dynamically
  }
}

**Comment by filrak** [comment]
@pkarw i added 2 things to the post btw, about modules and ssr plugin vue

**Comment by filrak** [comment]
We can theoretically filter the urls from SSR routing - however I don’t see many use cases for that. How wupuld you like to use it?

We discussed it earlier. For SEO ssr is useful only for homepage and catalog. By disabling ti on unnecesary pages people will be able to use client-side libs without any problems which will be a huge benefit in dev experience

File: core/build/webpack.base.config.js [hunk 3]
```diff
@@ -40,7 +44,19 @@ const postcssConfig =  {
 module.exports = {
   plugins: [
     new CaseSensitivePathsPlugin(),
-    new VueLoaderPlugin()
+    new VueLoaderPlugin(),
+    // generate output HTML
```
filrak: maybe this files should be provided in config or exported from some place that develoeprs can reach (not from core)?

PR #1798, opened 2018-10-01
WIP: New modules PoC with Cart Module

filrak:
Work in progress.

New (and final) version of VS modules. After long discussion with @pkarw we agreed that proposed architecture for modules can be improved which will end up with much better approach and make refactoring to modules easier.

**What changed**
- We are leaving the `features` idea. Instead of creating granular micro-mixins we will go again with bigger components. The purpose of `features` was to provide a granular and feature-oriented approach to composing components. We agreed that such functionality should be provided by Vuex stores instead of a mixins so instead of splitting whole app into granular features that are provided by mixins we will focus on encapsulating most of business logic into Vuex stores and providing easy to use API with it's actions, getters and mutations.
- Added `-module-name` postfix to all components from a given module. Just a minor tweak to navigate through components easier with Vue Devtools.
- ...more to come

**Backward compatibility plan**
- Even tho a lot of components will change due to refactoring and moving to modules which we will use in theme 2.0 we still need to remain backward compatible with theme 1.0 This is why we will treat `core/components` folder as a backward-compatible layer that proxys new API to the old one. This proxys will be maintained in the future so everything will be backward compatible BUT some new features may not be introduced to them.

To better undestand this proxing you can take a look at `core/components/Microcart/Product.js` or smaller `core/components/Header/MicrocartIcon.js` file


pkarw:
We need to have @igloczek and @DavidRouyer opinion on that

tests/unit/core/api/cart/cartShipping.spec.js
```diff
@@ -1,37 +0,0 @@
...
-        }
-      }
-    })
-  })
-
-  it('cart shipping equal to returned from api', () => {
-    const wrapper = shallowMount(TestInstance, {store, localVue})
-    expect(wrapper.vm.cartShipping).to.equal(store.state.cart.shipping)
-  })
-})
```
pkarw: These specs are missed; should it be like that?

core/modules/cart/cart/TestInstance.vue
```diff
@@ -13,20 +13,6 @@
 <script>
 import microcart from '@vue-storefront/core/components/blocks/Microcart/Microcart'
```
pkarw: Should be uppercase

core/components/blocks/Header/MicrocartIcon.js
```diff
@@ -1,20 +1,19 @@
...
   computed: {
-    totalItems () {
-      // return this.$store.getters.totals.quantity
-      return 3
-    },
-    ...mapGetters({
-      totals: 'cart/totals',
-      totalQuantity: 'cart/totalQuantity'
-    })
+    totalQuantity () {
```
pkarw: We should mark these things like @deprecated

core/components/blocks/Microcart/Microcart.js
```diff
@@ -1,36 +1,14 @@
...
-    cartShipping,
-    cartPayment
-  ],
-  props: {
-    isCheckoutMode: {
-      type: Boolean,
-      required: false,
-      default: () => false
+  methods: {
+    closeMicrocart () {
```
pkarw: Should be marked as deprecated

pkarw:
I’ve put some open questions regarding modules future and if they replace extensions one day (they should) and if they interoperate with npm (they should)

filrak:
@pkarw please don't comment directly here since I'm applying changes every 10 mins, just discuss it on slack to not confuse others ;)

filrak:
and @DavidRouyer @Igloczek please take a look at this, think but comment after I'll mark it as done :)

tests/unit/core/api/cart/cartShipping.spec.js
filrak: nope, still need to add this

filrak:
I'd consider most of stuff done. What's left
- [] rewrite tests for components
- [] typings

pkarw:
I think we should agree on the structure and align other modules at first then add missing parts (like tests) because what was the most misleading part of current modules was that unclear / duplicated / messy structure (old vs new)

pkarw:
Vuex Store should be ported at the very very end of the process

pkarw:
Filip let me also know regaridng the questions on npm and extensions vs modules question

DavidRouyer:
What about class-based components for components? https://alligator.io/vuejs/typescript-class-components/ Much nicer syntax and we'll be ready for Vue.js 3.x

core/components/blocks/Microcart/Product.js
```diff
@@ -1,53 +1,44 @@
...
-      }
-    },
     updateQuantity () {
+      // additional logic will be moved to theme
       this.qty = parseInt(this.qty)
       if (this.qty <= 0) {
         this.qty = this.product.qty
       }
-      this.$store.dispatch('cart/updateQuantity', { product: this.product, qty: this.qty })
+      Product.methods.updateQuantity.call(this, this.qty)
```
DavidRouyer: Don't like that at all

filrak:
@pkarw to your questions 3x Yes but registration will happen without the index file (just one entry for every module) so we have more control over how it's registered

filrak:
@DavidRouyer this will add unnecessary complexity with decorators etc. Vue 3 will still use object syntqx by default so I will stick with it

core/components/blocks/Microcart/Product.js
filrak: Why? It's kind of hack but its just for backward compatibility..It’s a solution to fact that mixed metods are overrided by methofs with the same names so if I want to extend it without repeating the code it's probably the only way

filrak:
FYI I'm moving to another modules. I will refactor the tests at the end since it's super time-consuming

pkarw:
For me - @filrak this is accepted. I like this approach, let's go with it.

Component options / props is much better for module path than name with postfix (it's messy for me). 

Can You add example of:
- package.json
- registration mechanism as You've described

filrak:
@pkarw I'll add the registration after finishing the refactoring

filrak:
Most of the current modules (except wishlist) ar refactored. They'll need some more work and poslihment later but for now they are acceptable. I also rewrote newsletter totally since it was mixed between extension, core and theme + completely not working (it still needs some work tho, it's more like a MVP)

DavidRouyer:
Nice work @filrak! 🎉

Issue #830, opened 2018-03-10
Improve modals/pop-ups animations


karolbzik:
We need to improve animation for modals.
TODO - 1) check good practices for material motion 2) describe improvements for modals.

jahvi:
I'd like to give this one a go if that's ok

pkarw:
Sure, please propose sth :)

jahvi:
I'm giving this a go at #1859, any feedback is appreciated 🙂

PR #1859, opened 2018-10-13
Update base transitions easing/speed

jahvi:
### Related issues

#830

The following is an attempt at improving the base transitions to provide a more natural and smooth motion. It's largely based on best practices from [Material Design Motion](https://material.io/design/motion/understanding-motion.html) and include these transition speed adjustments:

#### Decrease transition duration on "exit"

> Transitions that close, dismiss, or collapse an element use shorter durations, as they require less user attention than the user’s next task.

I set a shorter transition duration and easing for the "exit" animation, instead of using the same as "enter".

#### Change transition easing

> Incoming elements are animated using deceleration easing, which starts a transition at peak velocity (the fastest point of an element’s movement) and ends at rest.

> Elements exiting a screen use acceleration easing, where they start at rest and end at peak velocity.

I changed both timing functions so the "enter" animation decelerates sooner to put more emphasis on the element and increase acceleration on the "exit" animation as attention is not as necessary.

#### Reduce transition duration

> Appropriate transition speed helps users to follow UI changes, without complicating tasks. The shortest duration possible should be used that isn’t abrupt or jarring.

I found the previous 300ms could be decreased a bit while still making them easy to follow. I lowered "enter" durations to 250ms and "exit" durations to 200ms.

The quotes above were taken from the [Material Design Speed](https://material.io/design/motion/speed.html) recommendations. These changes affect mostly modals and notification popups but they can be applied to all elements in time.

The differences are very subtle but should give a more natural feel to modals/popups.

**Before**
motion

**After**
motion

### DIFF
--- a/src/themes/default/css/animations/_transitions.scss
+++ b/src/themes/default/css/animations/_transitions.scss
@@ -1,9 +1,15 @@
-$motion-main: cubic-bezier(0.54, 0.02, 0.35, 0.88);
-$duration-main: .3s;
+$motion-main: cubic-bezier(0.4, 0.0, 0.2, 1);
+$duration-main: .25s;
 
-.fade-enter-active,
-.fade-leave-active {
+$motion-exit: cubic-bezier(0.4, 0.0, 1, 1);
+$duration-exit: .2s;
+
+.fade-enter-active {
   transition: opacity $duration-main;
+}
+
+.fade-leave-active {
+  transition: opacity $duration-exit;
 }
 
 .fade-enter,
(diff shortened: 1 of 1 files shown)


filrak:
Good point and so far one of the best explained motivatiosn for PR I've seen :) 
We are also planning to follow this rules in new theme, would you be interested in helping us with animations there? (or e2e tests, excellent talk btw ^^)

jahvi:
Glad I could help! Let's keep in touch and let me know what you're thinking for the new theme I'll be happy to help.

PR #1898, opened 2018-10-25
Notification module

akbarik:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/1878

Currently notifications work by means of Event Bus. Proposed is a module for notifications with its own store where it keeps track of all notifications. Whenever any component wants to show a notifcation, it needs to import this module as a mixin and call its showNotification method. Also, action is now sent as a callback function from the component itself and not as string as before, which means any component decided on its own what to do after a certain action was triggered from a notification. 
Also, existing notifications from core components were moved to their theme components.


core/modules/notification/store/index.ts
```diff
@@ -0,0 +1,40 @@
+import { Module } from 'vuex'
+import NotificationItem from '../types/NotificationItem'
+
+export const module: Module<any, any> = {
```
filrak: you can create NotificationState object in /types to  match `Module<>` first 'argument'

core/modules/notification/components/Notification.ts
```diff
@@ -0,0 +1,17 @@
+import NotificationItem from '../types/NotificationItem'
+
+export const Notification = {
+  computed: {
+    notifications () {
+      return this.$store.getters['notification/notifications']
+    }
+  },
+  methods: {
```
filrak: maybe add/remove as a names of the functions will be more descriptive (and match Vuex actions)

src/themes/default/components/core/blocks/Auth/Login.vue
```diff
@@ -79,10 +79,10 @@ import ButtonFull from 'theme/components/theme/ButtonFull.vue'
 import BaseCheckbox from '../Form/BaseCheckbox.vue'
 import BaseInput from '../Form/BaseInput.vue'
 import { required, email } from 'vuelidate/lib/validators'
-import i18n from '@vue-storefront/i18n'
+import { Notification } from '@vue-storefront/core/modules/notification/components/Notification'
```
filrak: Instead of mixing in whole Notification component with 2/3 unused parts we can just dispatch appropiate action.
` this.$store.dispatch('notification/addNotification', notification)`

core/modules/notification/components/Notification.ts
filrak: Or spawnNotification (would also fit as a name of Vuex action but addNotification is also descriptive enough, just not so precise)

filrak:
Thanks buddy!

PR #1911, opened 2018-10-26
Proper name and private set for docs package

Igloczek:
Apart of these changes, I have a question about adding `docs` directory to  `package.json` yarn `workspaces` array. 
Docs are not part of the VSF app itself, so it's not important to install their dependencies for everyone. Also, this package is not going to be published to npm, so for me, it should be separated.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,4 +1,7 @@
 {
+  "name": "@vue-storefront/docs",
+  "private": true,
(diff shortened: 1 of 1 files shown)


Igloczek:
@pkarw eh, I asked the question there :D
@NataliaTepluhina I still want to know your point of view about workspaces.

Issue #971, opened 2018-03-31
Adjustments on "Search sidebar widget" component


karolbzik:
Current state:
<img width="876" alt="zrzut ekranu 2018-03-31 o 14 27 53" src="https://user-images.githubusercontent.com/18116406/38163168-c85b079e-34ef-11e8-9643-2981bfddf601.png">
<img width="935" alt="zrzut ekranu 2018-03-31 o 14 27 34" src="https://user-images.githubusercontent.com/18116406/38163170-c87d2f04-34ef-11e8-9240-94118cbfc0ec.png">
<img width="366" alt="zrzut ekranu 2018-03-31 o 14 27 18" src="https://user-images.githubusercontent.com/18116406/38163171-c89e6282-34ef-11e8-9f9a-c21224fc0396.png">

Desired state:
<img width="640" alt="zrzut ekranu 2018-03-31 o 14 29 29" src="https://user-images.githubusercontent.com/18116406/38163187-f4cba2d4-34ef-11e8-871e-23239fe1b2c7.png">
<img width="634" alt="zrzut ekranu 2018-03-31 o 14 29 21" src="https://user-images.githubusercontent.com/18116406/38163188-f4ed7850-34ef-11e8-82b2-469abea50ec8.png">
<img width="257" alt="zrzut ekranu 2018-03-31 o 14 29 09" src="https://user-images.githubusercontent.com/18116406/38163189-f50f8224-34ef-11e8-91f8-38dd80b9907a.png">
<img width="629" alt="zrzut ekranu 2018-03-31 o 14 28 52" src="https://user-images.githubusercontent.com/18116406/38163190-f531c0b4-34ef-11e8-8783-f06b2299df3a.png">

Adjustments:

1. We need to change a font, size and position of a search field, as a font we need to use Playfair Display 18px, the height of the search field should be 60px, the search field should be 60px height with 40px margin from left and right, search icon is the same as the one used in the top
![12](https://user-images.githubusercontent.com/18116406/38163228-dfa661cc-34f0-11e8-9e94-a91b7849a74e.png)

2. We should apply styles for an inactive stage of the field 
<img width="634" alt="zrzut ekranu 2018-03-31 o 14 29 21" src="https://user-images.githubusercontent.com/18116406/38163246-24b71fe0-34f1-11e8-9b2e-5b2635858e9c.png"> and different styles for active field
[...]

PR #1912, opened 2018-10-26
Fix - Adjusted search sidebar component #971

Progowi:
### Related issues

#971

I adjusted search sidebar component to look like template in figma:
- change width of sidebar from 800px to 928px
- add new scss file with all media queries and grid gutters widths to use them globally
- add new scss file with variables for fonts to use them globally
- grid gutters widths on mobile from 20px to 10px
- change font size in search panel input to 18px on desktop and 16px on mobile


Progowi:
@filrak 
before:
prev2
prev
after:
after
after2

And I added variables to scss for breakpoints, grid gutters and font-family

Issue #1667, opened 2018-08-30
Recently View products Module


pkarw:
We should add the "Recently view products" to the My Account section.

Acceptance criteria:
- [ ] to keep the Recently view products in localForage / local storage in the browser
- [ ] to have the My Account view for it (using product Tiles - the same layout as we do have in the category)

PR #1917, opened 2018-10-26
#1667: add recently viewed products module

talalus:
#1667


core/pages/Product.js
```diff
@@ -99,6 +100,7 @@ export default {
           this.loading = false
           this.defaultOfflineImage = this.product.image
           this.onStateCheck()
+          this.$store.dispatch('recently-viewed/addItem', this.product)
```
filrak: this will not work if we click on any of suggested products since `product` component will remain mounted (in other words hook will not be called). I believe we can add this product automatically for example 
- create some router afterEach hook for this or 
- extend current product store (store.extends property) to dispatch this action each time product is changed
- listen for product object mutation (preferred but you need to take care of variants - we need to push the object from mutation only if the whole product has changed, not only it's variant)

filrak:
can you also solve the conflicts

PR #1921, opened 2018-10-26
'core' alias removed

Igloczek:
We should prevent devs from using this imports, especially like `core/store` etc., b/c that should be used as a package, not part of monolithic aplication.
It's part of cahnges necessary to build boilerplate.

Tests are currently failing, b/c wrong imports are still in code, it's fixed on #1920


filrak:
@igloczek can you merge latest develop and check if it'll work properly with aliases removal?

Igloczek:
I'm going to remove Stripe payment module, since it's not a core feature.
Ofc these aliases changes need to be reflected in this module too.

Igloczek:
@collymore I just removed a few things from the VSF core that were related to your `vsf-payment-stripe` module.
It's not a core feature, so it should work standalone, including having own docs about configuring it in VSF. Apart of docs, you need to update imports, to reflect changes in webpack aliases + update dependencies in `package.json`

Igloczek:
@filrak Merged, quickly tested, works.

filrak:
Ok, I'm merging ;)

PR #1922, opened 2018-10-26
WIP: Dependecies cleanup

Igloczek:
I moved dependencies from root to packages and added missing one.
Also `i18n` and `store` are removed from `core` package.


Igloczek:
I need to work a bit more on it...
image

filrak:
Is it yes? :D

Igloczek:
I'll try to do it in smaller chunks, kinda hard to track down what went wrong editing everything at once.

PR #1923, opened 2018-10-26
Lerna non-independent versioning

Igloczek:
I think it will be easier to have one version for all packages, since at all we keep these versions almost the same.


filrak:
Agree. It'll be easier to know which modules are from which versions :)

Issue #1903, opened 2018-10-26
Create Compare module


filrak:
## What is the motivation for adding / enhancing this feature?

Refactoring to modules ;)

## What are the acceptance criteria 

- Move all compare-related functionality into a VS module.
- Change the API of vuex actions/components if needed to follow module good practices. Use core/components folder as a facade for backward compatibility (you can find examples of such porting in Microcart).

## Additional information

https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md
You can contact me directly on slack before starting this issue.

PR #1943, opened 2018-10-30
Feature/1903

szafran89:
### Related issues

#1903 

I've moved some compare related stuff to VS module. I know there are missing docs but want to make sure if it's ok what I did. 


szafran89:
@filrak check my changes and let me know if it's ok what I did please.

core/modules/compare/components/Compare.ts
```diff
@@ -0,0 +1,47 @@
+import i18n from '@vue-storefront/i18n'
+import Product from '@vue-storefront/store/types/product/Product'
+
+export const Compare = {
+  name: 'Compare',
+  props: {
```
filrak: I don't think the title should be determined by a prop. Can you take it from products info? Moving to modules also involves correcting poor APIs like this

core/modules/compare/components/CompareButton.ts
```diff
@@ -0,0 +1,8 @@
+export const CompareButton = {
+  name: 'CompareButton',
+  computed: {
+    compareIsActive () : boolean {
```
filrak: `isEmpty` with opposite value may be more descriptive (also to change in Vuex) what do you think?

core/modules/compare/components/Product.ts
```diff
@@ -0,0 +1,22 @@
+import Product from '@vue-storefront/store/types/product/Product'
+
+export const CompareProduct = {
+  name: 'CompareProduct',
+  computed: {
+    isOnCompare () : boolean {
```
filrak: aren't all products passed to this component on compare? isn't this property useless?

core/modules/compare/components/Product.ts
```diff
@@ -0,0 +1,22 @@
...
+      return !!this.$store.state.compare.items.find(p => p.sku === this.product.sku)
+    }
+  },
+  methods: {
+    addToCompare (product: Product) {
```
filrak: same here, what is the purpose of adding product from comapre to comapre? It should be a separate component like `AddToCart` or `AddToWishlist`

filrak: wishlish and cart are also w good benchmarks for this module since functionality is almost the same

core/modules/compare/store/index.ts
```diff
@@ -2,10 +2,9 @@ import { Module } from 'vuex'
 import actions from './actions'
 import getters from './getters'
 import mutations from './mutations'
-import RootState from '../../types/RootState'
-import CompareState from './types/CompareState'
+import CompareState from '../types/CompareState'
```
filrak:
```suggestion
import CompareState from '../types/CompareState'
import RootState from '@vue-storefront/store/types/RootState'
```
you can use `RootState` instead of any

core/modules/compare/store/index.ts
```diff
@@ -2,10 +2,9 @@ import { Module } from 'vuex'
...
 
-const compare: Module<CompareState, RootState> = {
+export const module: Module<CompareState, any> = {
```
filrak: you can use `RootState` instead of any
```suggestion
export const module: Module<CompareState, RootState> = {
```

core/modules/compare/components/Compare.ts
```diff
@@ -0,0 +1,47 @@
...
+    title: {
+      type: String,
+      required: true
+    }
+  },
+  metaInfo () {
+    return {
```
filrak: should be moved to theme. If this is meant to be a component instead of page (which is correct) then it shouldn't interact with meta.

core/modules/compare/components/Product.ts
filrak: + can we move it to vuex getter? I see a lot of use cases . I'll create task for isOnCart and isOnWishlist too

So we will end up with this.$store.getters.isOnCompare(product)

core/pages/Product.js
```diff
@@ -115,10 +113,10 @@ export default {
       return this.$store.state['wishlist'] ? this.$store.dispatch('wishlist/removeItem', product) : false
     },
     addToList (list) {
-      return this.$store.state[list] ? this.$store.dispatch(`${list}/addItem`, this.product) : false
+      CompareProduct.methods.addToCompare.call(this, this.product)
```
filrak: dont forget about comments what has changed in this method. they'll be used for migrations guide

core/modules/compare/components/Compare.ts
szafran89: This is prop which comes from routing. I moved it to theme with metaData object, b/c it's related.

core/modules/compare/components/CompareButton.ts
szafran89: changed to isEmpty, but IMO we should change `CompareButton` component to  `isCompareEmpty` and add there `isCompareEmpty` computed property. We use this computed property in two places but can be helpfull in multiple places. What fo you thing?

core/modules/compare/components/Product.ts
szafran89: Created getter `isOnCompare`, I've used arrow functions there. Is it ok for you?
```
const getters: GetterTree<CompareState, RootState> = {
  isEmpty: (state) => state.items.length === 0,
  isOnCompare: (state) => (product) => state.items.find(p => p.sku === product.sku)
}
```

core/modules/compare/components/Product.ts
szafran89: Created separate component `addToCompare `

core/modules/compare/store/index.ts
szafran89: done

core/modules/compare/store/index.ts
szafran89: done

core/modules/compare/components/Compare.ts
szafran89: moved to theme.

core/pages/Product.js
szafran89: 👍

core/modules/compare/components/Compare.ts
filrak: damn, you're right. I somehow missed that it's a prop

core/modules/compare/components/CompareButton.ts
filrak: You mean completely remove`CompareButton` or add `isCompareEmpty` component? in both cases i think it should be a vuex action, not a component since it's just a function with return value,

PR #1981, opened 2018-11-10
Feature/dynamic config load

CodeSkills:
### Related issues

None

This enables to load dynamic config on every page load. 

Anyone can simply define loadConfig function within `src/server/index.js` that returns a Promise.

```javascript
module.exports.loadConfig = (req) => {
  const axios = require('axios')
   return new Promise((resolve, reject) => axios.get('myapi.com/config', {
     params: {
       domain: req.headers.host
     }
   }).then(res => {
    resolve(res.data)
   }).catch(error => reject(error)))
 }
```

None

Using own API - tested locally (debug and visual)

### DIFF
--- a/core/modules/user/store/actions.ts
+++ b/core/modules/user/store/actions.ts
@@ -315,9 +315,9 @@ const actions: ActionTree<UserState, RootState> = {
     const usersCollection = Vue.prototype.$db.usersCollection
     usersCollection.setItem('current-token', '')
   
-    if (rootStore.state.route.path === '/my-account') {
+    // if (rootStore.state.route.path === '/my-account') {
       // router.push('/')
-    }
+    // }
   },
   /**
    * Load user's orders history
--- a/core/scripts/server.js
+++ b/core/scripts/server.js
@@ -243,6 +243,10 @@ app.get('*', (req, res, next) => {
       serverExtensions.configProvider(req).then(loadedConfig => {
         config = Object.assign(require('config'), loadedConfig) // merge loaded conf with build time conf
         dynamicCacheHandler()
+      }).catch(error => {
+        console.log(error)
+        config = require('config') // reload config
+        dynamicCacheHandler()
       })
     } else {
       config = require('config') // reload config
--- a/core/scripts/server.js
+++ b/core/scripts/server.js
@@ -150,10 +150,6 @@ app.get('*', (req, res, next) => {
           '  </html>')
       return next()
     }
-    if (config.server.dynamicConfigReload) {
-      delete require.cache[require.resolve('config')]
-      config = require('config') // reload config
-    }
     const context = {
       url: req.url,
       output: {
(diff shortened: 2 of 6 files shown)


filrak:
With this approach is keeping config in apps state still necessary? @pkarw wdyt?

core/modules/user/store/actions.ts
```diff
@@ -315,9 +315,9 @@ const actions: ActionTree<UserState, RootState> = {
     const usersCollection = Vue.prototype.$db.usersCollection
     usersCollection.setItem('current-token', '')
   
-    if (rootStore.state.route.path === '/my-account') {
+    // if (rootStore.state.route.path === '/my-account') {
       // router.push('/')
```
pkarw: Why this got commented? Please revert the original code in here :)

core/scripts/server.js
```diff
@@ -208,35 +204,52 @@ app.get('*', (req, res, next) => {
...
-    }).catch(errorHandler)
+      }).catch(errorHandler)
+    } else {
+      dynamicRequestHandler(renderer)
+    }
+  }
+
+  if (config.server.dynamicConfigReload) {
+    delete require.cache[require.resolve('config')]
+    if (typeof serverExtensions.loadConfig === 'function') {
```
pkarw: Can we remove “loadConfig” to “configProvider”? I would make it a convention of providers for other resources as well

core/modules/user/store/actions.ts
CodeSkills: Within the Develop Branch the state did not had route defined. So it wont let me build the app. And since it did nothing (the router.push was already commented) I just commented out the whole function until somebody implements it. 

Should I revert it back or leave it there (but it will throw an error "route not defined on rootStore.state")?

core/scripts/server.js
CodeSkills: Sure, this is a great idea. Will do.

core/modules/user/store/actions.ts
pkarw: Ok! It’s fine, there is second redirect in pages/MyAccount.js, asyncData - and this one must stay there :)

CodeSkills:
@pkarw Renamed the function and added basic Docs to the end of Config Docs.

pkarw:
Awesome, Thanks! Will merge it once get in front of my computer :)

CodeSkills:
Last thing, when the Promise is rejected (for some reason which could happen). There should be 2 options 

1) Continue with default (build time/runtime config)
2) Throw Error 500 page 

These are rare occurrences but will eventually happen. 

@pkarw Should I add a config for dynamicConfigReload that can decide between these two?

pkarw:
Yes Please, thats a good point. We should have the option in config for that 👍

CodeSkills:
Added the option and tested all options with latest changes to check that all works as expected.

Let me know if you come up with better naming for the option :-)

It now allows to configure an option if it should continue with built time config or to be redirected to /error page.

pkarw:
Perfect. Thanks!

PR #1982, opened 2018-11-10
Feature/currency sign after

CodeSkills:
### Related issues

None

Enable to change the position of currency sign to be prepended or appended depending on `currencySignBehind` i18n config value. (Default false)

screenshot 2018-11-10 at 15 07 09

Visual & functional test


config/default.json
```diff
@@ -267,6 +267,7 @@
       "defaultLocale": "en-US",
       "currencyCode": "USD",
       "currencySign": "$",
```
pkarw: Can we rather use the naming like:
currencySign: “append”/“prepend”? Plus add a line into config docs and let know @NataliaTepluhina to update the official docs :)

CodeSkills: Not sure exactly what to change it to, @pkarw you mean `"currencySignAppend": false` ?

pkarw: I was rather up to changing this from bool to string (well - enum kind of with “append” and “preppend” options)

CodeSkills: To something like `currentSignPlacement: 'append' || 'prepend'` ? Or how would you @pkarw name the option "currentSignPlacement" ? :)

pkarw: Correct, i was thinking about something like this. I’m not a naming-nazi ;) so please use the naming of your choice it was more like suggestion / idea :-)

CodeSkills:
@pkarw Changed the option to ENUM and added info to Docs.

CodeSkills:
Should be solved now.

PR #2001, opened 2018-11-16
Feature/cms content

yuriboyko:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/1987
https://github.com/DivanteLtd/vue-storefront/issues/1894

This PR adds functionality for using CMS content from Elasticsearch with supporting SSR
It uses dynamic routes for CMS pages

For CMS block usage example please check demo page Vue file 
src/themes/default/pages/CmsBlockDemoPageSsr.vue

### Data structure

For CMS pages uses elasticsearch index entity type 'cms_page':

    `id`
    `title`
    `meta_keywords`
    `meta_description`
    `identifier`
    `content_heading`
    `content`
    `creation_time`
    `update_time`
    `active`

For CMS blocks uses elasticsearch index entity type 'cms_block':

    `id`
    `title`
    `identifier`
    `content`
    `creation_time`
    `update_time`

### Announcement
Later will be provided functionality for page hierarchy 

### Magento support
Data migration tools for Magento 1 and Magento 2 will be updated accordingly very soon with correspond PRs


core/components/blocks/Cms/Block.js
```diff
@@ -0,0 +1,41 @@
+export default {
```
filrak: core components folder is now depreciated (pull latest dev), please use it in appropiate module

core/modules/cms/store/block/actions.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (state.cmsBlocks.length > 0) {
+          // SSR - there were some issues with using localForage, so it's the reason to use local state instead, when possible
+          let cmsBlock = state.cmsBlocks.find((itm) => { return itm[key] === value })
+          if (cmsBlock) {
+            resolve(cmsBlock)
+          } else {
+            reject(new Error('CMS block query returned empty result ' + key + ' = ' + value))
+          }
+        } else {
+          Vue.prototype.$db.cmsBlocksCollection.getItem(entityKeyName(key, value), (error, cmsBlock) => {
```
filrak: please use InitCacheStorage from `module/helpers` for cache storage instead of addint this reference to each vue instance, you can find examples in other modules like compare/wishlist

core/modules/cms/store/block/mutations.ts
```diff
@@ -0,0 +1,33 @@
...
+   */
+  [types.CMS_BLOCK_UPDATE_CMS_BLOCKS] (state, cmsBlocks) {
+    // save cmsBlocksList data to the state and DB
+    state.cmsBlocks = cmsBlocks
+
+    console.log('cmsBlocks mutation', cmsBlocks.length)
+
+    for (let cmsBlock of cmsBlocks) {
+
+      const cmsBlocksCollection = Vue.prototype.$db.cmsBlocksCollection
```
filrak: like in `store/block/actions.ts` use initCacheStorage please. Vue instance shouldn't be a place to access data. Only vuex store should be responsible for accessing data and it can take this data from cache/server

core/pages/CmsPage.js
```diff
@@ -0,0 +1,55 @@
...
+  },
+  asyncData ({ store, route, context }) { // this is for SSR purposes to prefetch data
+    return new Promise((resolve, reject) => {
+      if (context) context.output.cacheTags.add(`cmsPage`)
+      store.dispatch('cms_page/single', {
+        value: route.params.slug
+      }).then(res => {
+        if (res) {
+          store.state.cms_page.current = res
+          EventBus.$emitFilter('cms_page-after-load', { store: store, route: route }).then(results => {
```
filrak: we are slowly depreciating event bus, please don't base your solutions on it

core/pages/CmsPage.js
```diff
@@ -0,0 +1,55 @@
...
+      })
+    })
+  },
+  methods: {
+    validateRoute () {
+      this.$store.dispatch('cms_page/single', { key: 'slug', value: this.$route.params.slug }).then(cmsPage => {
+        if (!cmsPage) {
+          this.$router.push('/')
+        } else {
+          this.$bus.$emit('cms_page-changed', this.$store.state.cms_page.current_path)
```
filrak: EventBus is depreciated

core/store/index.ts
```diff
@@ -202,6 +202,16 @@ export function initStore () {
       storeName: 'ordersHistory',
       driver: localForage[config.localForage.defaultDrivers['ordersHistory']]
     })),
+    cmsPagesCollection: new UniversalStorage(localForage.createInstance({
```
filrak: this should be initialised in before/after module hooks and not bound to Vue instance

core/store/lib/search/adapter/graphql/queries/cmsBlock.gql
```diff
@@ -0,0 +1,13 @@
+query cmsBlocks ($filter: CmsInput) {
```
filrak: should be in module/queries

core/store/lib/search/adapter/graphql/queries/cmsHierarchy.gql
```diff
@@ -0,0 +1,14 @@
+query cmsHierarchies ($filter: CmsInput) {
```
filrak: should be in module/queries

core/store/lib/search/adapter/graphql/queries/cmsPage.gql
```diff
@@ -0,0 +1,14 @@
+query cmsPages ($filter: CmsInput) {
```
filrak: should be in module/queries

core/store/mutation-types.ts
```diff
@@ -1,3 +1,3 @@
 export const SN_SYNC = 'sync'
 export const SYNC_ADD_TASK = SN_SYNC + '/ADD_TASK'
-export const SYNC_PROCESS_QUEUE = SN_SYNC + '/PROCESS_QUEUE'
\ No newline at end of file
+export const SYNC_PROCESS_QUEUE = SN_SYNC + '/PROCESS_QUEUE'
```
filrak: should be in module

core/store/lib/search/adapter/graphql/queries/cmsBlock.gql
yuriboyko: This gql is related to the ceratin searchAdapter. Not sure if it should be in the module/queries

core/store/lib/search/adapter/graphql/queries/cmsHierarchy.gql
yuriboyko: This gql is related to the certain searchAdapter. Not sure if it should be in the module/queries

core/store/lib/search/adapter/graphql/queries/cmsPage.gql
yuriboyko: This gql is related to the certain searchAdapter. Not sure if it should be in the module/queries

core/pages/CmsPage.js
yuriboyko: Removed

core/pages/CmsPage.js
yuriboyko: Removed

config/default.json
```diff
@@ -293,6 +293,12 @@
       "endpoint": "http://localhost:8080/api/ext/cms-data/cms{{type}}/{{cmsId}}",
       "endpointIdentifier": "http://localhost:8080/api/ext/cms-data/cms{{type}}Identifier/{{cmsIdentifier}}/storeId/{{storeId}}"
     },
+    "cms_block": {
+      "max_count": 500
+    },
+    "cms_page": {
+      "max_count": 500
+    },
```
pkarw: We shouldn’t load all cms blocks and pages as it will cause sending them with the __INITIAL_STATE__ output and can be harmful boot h for performance + SEO

core/store/lib/search/adapter/api/searchAdapter.ts
```diff
@@ -170,5 +170,32 @@ export class SearchAdapter {
...
+    })
+    this.registerEntityType('cms_hierarchy', {
+      queryProcessor: (query) => {
+        // function that can modify the query each time before it's being executed
+        return query
+      },
+      resultPorcessor: (resp, start, size) =>  {
+        return this.handleResult(resp, 'cms_hierarchy', start, size)
+      }
+    })
```
pkarw: These entities should be registered from the module level - keeping all the CMS oriented logic inside the cms module - not in the core I believe (currently CMS is also a module without references from the core)

core/store/lib/search/adapter/graphql/queries/cmsBlock.gql
pkarw: I agree with @filrak; if there is currently no way to fully add query/resultAdapter from the module level - then we need to extend the graphQL API to have this capability. Anyway - everything should be registered from the module level not modifying the core; CMS module should be a example how decoupled the app is with new module system

core/store/lib/search/adapter/graphql/queries/cmsHierarchy.gql
pkarw: Same as above - it should be within the module only. We need to have a capability of extending graphQL entities for custom ones from the module level

core/store/lib/search/adapter/graphql/queries/cmsPage.gql
pkarw: Same as above

src/themes/default/components/core/blocks/Cms/Block.vue
```diff
@@ -0,0 +1,51 @@
...
+      required: false
+    },
+    sync: {
+      type: Boolean,
+      default: false,
+      required: false
+    }
+  },
+  asyncData ({ store, route, context }) {
+    // @TODO to cover SSR need find a way to pass props identifier/id to the asyncData()
```
pkarw: asyncData() won’t be called for child components; so if You place 3 different CmsBlocks on one single page - asyncData won’t be called for these CmsBlocks. Therefore we need this dataLoader pattern implemented - and then we can call store.dispatch(‘dm/registerPromise’, this.loadData()) in created() event handler as we discussed with @yuriboyko on Slack

core/store/lib/search/adapter/graphql/queries/cmsBlock.gql
yuriboyko: ok i see the idea. then probably it makes sense 
i will try move register entity Type to the module level

yuriboyko:
@pkarw  I hope to finish all changes today

pkarw:
@yuriboyko  - perfect! Thanks

core/store/lib/search/adapter/api/searchAdapter.ts
yuriboyko: same as for gql comments - Lets move this update to another PR

src/themes/default/components/core/blocks/Cms/Block.vue
yuriboyko: As i see asyncData() is called for child components in our case but the main problem is what we are not able to use component props to fetch correspond block data by identifier. Suggested solution by @pkarw is stll in the progress.  For now I suggest this using SSR for CMS blocks it as a stuff for next pull request

core/store/lib/search/adapter/graphql/queries/cmsPage.gql
filrak: okay ;)

filrak:
@yuriboyko can you please let us know when its ready ?

config/default.json
yuriboyko: this config option will need to easily manage  load of all blocks list if it needs as we need to have size provided for ES to get the data overwise it will be  only 10

yuriboyko:
@filrak 
At this moment all possible changes are done and commited.
As i mentioned before regarding moving queries to module level it will be separate PR
Regarding using SSR for blocks and do not using preload i already talked to Piotr 
and i think for now would be better to not use SSR for blocks, as to use it we need to preload all blocks to solve issues with not accessible props, but unfortunately it is also not possible to read cachestorage on the asyncData level. This means even if we would use SSR with load all blocks and saving them to cachestorage next time with page load there will be not way to verify if blocks are loaded to cachestorage (as cachestoarge  not acessible) so blocks loads will happen again which is not good. So for now i suggest to not change anything and then if we find some other solution for blocks SSR it will be another PR.

So for now i think current state is ready for approve PR to have this CMS functionality. if you see some other issues let  me know

src/themes/default/components/core/blocks/Cms/Block.vue
yuriboyko: As i mentioned. For now CMS blocks recoded to not use SSR as solution initially suggested  does not work and the way with loading all blocks also is not good.  Need find a new solution and it will be a new PR.

core/store/lib/search/adapter/api/searchAdapter.ts
pkarw: Could you create an issue for that?

yuriboyko: Issue is created

filrak:
@pkarw if everything is fine from your side can you approve ?

Issue #1864, opened 2018-10-15
Simplify ProductBundleOptions component


patzick:
## What is the motivation for adding / enhancing this feature?
ProductBundleOptions has too complex logic. This broke e2e tests lately.
Example:
https://github.com/DivanteLtd/vue-storefront/blob/868cd8f2926de0b34ab90083c5afce7133016c7e/src/themes/default/components/core/ProductBundleOptions.vue#L31

## What are the acceptance criteria 
- [ ] each option should be a separate component
- [ ] avoid concatenations such as an example above

# Can you complete this feature request by yourself?
YES

## Additional information
Related PR: https://github.com/DivanteLtd/vue-storefront/pull/1861

filrak:
Feel free to do this, thanks!

patzick:
PR merged, closing.

PR #2006, opened 2018-11-20
1864 simply productbundleoptions

patzick:
### Related issues

#1864

I've separated ProductBundleOption from ProductBundleOptions. It has simplified structure a little bit.
Tried not to touch too much at once to avoid business logic errors.
There are no visual changes.


pkarw:
Cool! Have You tested it with Bundle products on demo data set?
Please fix the "This branch is out-of-date with the base branch" issue to let me merge this in

core/modules/catalog/components/ProductBundleOptions.ts
```diff
@@ -3,19 +3,6 @@ import * as types from '../store/product/mutation-types'
 import rootStore from '@vue-storefront/store'
 import i18n from '@vue-storefront/i18n'
 
-function _defaultOptionValue (co, field = 'id') {
-  if (co.product_links && co.product_links.length) {
-    const defaultOption = co.product_links.find(pl => { return pl.is_default })
-    if (defaultOption) {
```
pkarw: Don't we need this logic - I mean the case of field === '*'? I don't see it anywhere in the newly proposed solution

core/modules/catalog/components/ProductBundleOptions.ts
```diff
@@ -57,83 +38,69 @@ export const ProductBundleOptions = {
...
-            const validationResult = validator(this['inputValues'][fieldName])
-            this.validation.results[fieldName] = validationResult
+            const quantityValidationResult = validator(qty)
+            if(quantityValidationResult.error) validationResult = quantityValidationResult
+            const optionValidationResult = validator(optionId)
+            if(optionValidationResult.error) validationResult = optionValidationResult
+            this.$set(this.validationResults, fieldName, validationResult)
             if (validationResult.error) {
-              this.product.errors['bundle_options_' + fieldName] = i18n.t('Please configure product bundle options and fix the validation errors')
+              this.product.errors['bundle_options_' + fieldName] = i18n.t('Please configure product custom options and fix the validation errors')
```
pkarw: shouldn't it be "bundle" instead of "custom" options?

patzick: Translation file doesn't have "bundle" keypath, so i picked up closest entry match. With "bundle" string console shows i18 warnings:
`vue-i18n.esm.js:14 [vue-i18n] Cannot translate the value of keypath`

core/modules/catalog/components/ProductBundleOption.ts
```diff
@@ -0,0 +1,74 @@
...
+    quantity(value) {
+      this.bundleOptionChanged()
+    }
+  },
+  methods: {
+    setDefaultValues() {
+      if(this.option.product_links) {
+        const defaultOption = this.option.product_links.find(pl => { return pl.is_default })
+        this.productOptionId = defaultOption ? defaultOption.id : this.option.product_links[0].id
+        this.quantity = defaultOption ? defaultOption.qty : 1
```
patzick: sorry, i have been setting default quantity always as 1, now it's ok and is set from default option if found. 1 is otherwise

core/modules/catalog/components/ProductBundleOptions.ts
pkarw: Please just add this string to translation file (en-US)

patzick: Sure, done :)

Issue #1906, opened 2018-10-26
Rewrite extensions to modules.


filrak:
## What is the motivation for adding / enhancing this feature?

Current extensions system will be depreciated and we will use VS modules for this purpose. This task involves rewriting
## What are the acceptance criteria 

Woudl be amazing if extension creators can handle it by themselves
- [x] .droppoint-shipping 
- [x] GA 
- [ ] magento2-cms ext by @anqaka 
- [ ] payment-backend-methods ext by @MikeSheward 
- [ ] payment cash on delivery
- [ ] raw-output-example by @pkarw 
- [ ] sample-custom-entity-gql by @phoenix-bjoern 

## Additional information

https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md

kruchy8:
payment-cash-on-delivery seems to be unused on latest develop

pkarw:
:D @filrak so You should close this one :)

PR #2019, opened 2018-11-23
[WIP] PaymentBackendMethods - Rewritten as module

CodeSkills:
### Related issues

#1906 

Rewritten to module

### Problems

@filrak within [Core Cart Module](https://github.com/DivanteLtd/vue-storefront/blob/develop/core/modules/cart/store/actions.ts#L354) there is committed `setBackendPaymentMethods` but this is defined inside `src/modules` (was in `src/extensions`) which should not be depended on (core module cant use src modules). 

So I have changed this to emit and event instead of committing, so modules can listen to it and react as needed. Is this a good approach for this case?


pkarw:
@collymore this is sth You should take a look at this regarding https://github.com/develodesign/vsf-payment-stripe

I think that would be great to migrate `vsf-payment-stripe` + `vsf-payment-paypal` to new modules at some point - I don't say: now but at some point

src/modules/payment-backend-methods/hooks/afterRegistration.ts
```diff
@@ -0,0 +1,30 @@
+import VueAnalytics from 'vue-analytics'
```
filrak: im not sure if we should use vueanalytycs here since they ar edepending on some other module

CodeSkills:
@filrak should be ok now :)

filrak:
Thanks!

PR #2026, opened 2018-11-26
Framework agnostic vs modules

filrak:
-


pkarw:
I think modules can not be framework agnostic in the end ... some of them can include React components - when VS will be supporting React, some won’t contain any framework-related feature; some can support both or even more frameworks - it depends on our approach to the agnosticity which is not yet set; therefore it’s not a good idea to take this feature (or adding custom routes) away as long as we won’t be back with some general solution

pkarw:
So I think we should make custom routes optional either add abstraction layer for defining them either pass the router object and let the module check which type is it (Vue router either React router etc)

WDYT @filrak?

PR #2037, opened 2018-11-28
[WIP] fix infinity loop on product page, poc of product variants availability

patzick:

I've found infinite loop on product page. It was caused by continuously configuring product invoked by product attributes visibility condition.
This PR is now more like PoC for that option. @pkarw has mentioned to me, that `configurable_children ` cannot be trusted in case of product accesibility.

So my proposal is:
When current product has changed, then asynchronous vuex action checks in background every variant (which method should it invoke?). When it checks variants, then puts available as new variable in store. Then `getAvailableProductOptions` getters gets directly that options to show on product page.

Things to do to finish this work:
- [ ] move ProductStatus enum; where to? (i got statuses from [this docs](https://devdocs.magento.com/guides/m1x/api/rest/Resources/Products/products.html), anyone has better source?)
- [ ] clarify how exactly check if product variant is available
- [ ] clarify why we do not show outOfStock product when config `listOutOfStockProducts` is set to true


pkarw:
@patzick - I’m available for a discussion at 9:00 at Divante Caffee ;)

patzick:
Sure, I'll be there :)

Issue #1749, opened 2018-09-20
RFC: i18n


DavidRouyer:
- Feature Name: i18n overhaul
- Start Date: 2018-09-20
- RFC PR: 

# Summary
[summary]: #summary

Overhaul of the i18n module by:
 - Supporting multistore scenarios & multilanguage stores
 - Move all the translations into one JSON file inside the theme, and introduce core/modules/extensions translation keys scoping
 - Add a mechanism for modules to expose translations keys & add warnings for missing translations keys for enabled languages in development mode
- BONUS: better translation key naming

# Motivation
[motivation]: #motivation

As of today, we only support multistore scenarios.
It's impossible to know which translation key is exposed by a module or an extension and it's hard to find them because they are all mixed up.
We have to merge files and convert them from CSV files.
As the translations keys are not in the theme file, I don't know what are the default translations and I don't know which languages are supported. If I want to translate my store in Latin, I have to implement ALL core translations keys.
Translations keys can clash: if I add an extension with an existing core translation key I can override it without knowing it

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

Example of the new translation files:
```json
{
  "core": {
    "ok": "OK",
    "remove-compare": "Remove from compare",
    "resetting-password": "Resetting the password ...",
    "error-reset-password-email": "Error while sending reset password e-mail"
  },
  "offline-order": {
    "confirm-order-offline": "Your order has been confirmed but you're offline"
  },
  "google-analytics": {
    "cookie-google-analytics": "This website uses Google Analytics"
  },
  "home": {
    "home-page": "This is our home page"
  },
  "about": {
    "introduction": "Introducing our team"
  }
}
```

How to use them?

`t('core.ok')`

Each time you add a page, a module or an extension, you add the scoped section inside the JSON file and you add the translations exposed by the module/extension.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

# Drawbacks
[drawbacks]: #drawbacks

The users who wants to upgrade will have to migrate their CSV files to the new format.

# Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

# Prior art
[prior-art]: #prior-art

# Unresolved questions
[unresolved-questions]: #unresolved-questions

filrak:
Totally agree that it's a good way, now it's such a mess

Regarding extension scoping - extensions will also be a modules so I'm wondering if the separation is needed. H

[...]

PR #2042, opened 2018-11-29
[DO NOT MERGE] i18n improvements

DavidRouyer:
### Related issues

#1749 

This is the current goal I want to achieve

- Put all translations in the theme folder
- Translations files are now JSON files (translation.preprocessor is not needed anymore)
- Use `i18n.setLocaleMessage('en-US', require('theme/resource/i18n/en-US.json'))` in the theme to bundle the default language and lazy load the others
- Remove unneeded config entries
- **TODO**: make the JSON file look more like the RFC one (I will simply create a `core` section to avoid breaking user's current translations, future work will be done in the v2 of the theme)

- [ ] No upgrade steps required (100% backward compatibility)
- [ ] I've updated the [Upgrade notes](https://github.com/DivanteLtd/vue-storefront/blob/develop/doc/Upgrade%20notes.md) and [Changelog](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md) on how to port existing VS sites with this new feature **TODO**


pkarw:
Ok, I see this for v1.7.0 @filrak

filrak:
@DavidRouyer do we really need to lazy load all languages? I mean we can even load them on demand (or on hover of language switcher, this is a nice way to do preloads 1-2 sec before click)

pkarw:
We should preload only these langs that are set in the`config.storeViews` - it makes no sense to preload languages that user can not use anyway

core/i18n/index.ts
```diff
@@ -19,27 +19,20 @@ function setI18nLanguage (lang: string): string {
 }
 
 export function loadLanguageAsync (lang: string): Promise<string> {
-  if (!config.i18n.bundleAllStoreviewLanguages) {
```
pkarw: Wht this option has been removed? It was a solution for "lang blinking" and to not preload all the translations all the time

DavidRouyer:
@filrak Then, how to load them on demand? 😄 That's the point of the `loadLanguageAsync()` which is called when you switch between stores.

I removed the line https://github.com/DivanteLtd/vue-storefront/blob/92310aa37d004ef8dac22f75af73354ce3739ec8/core/i18n/index.ts#L43 so they are not preloaded anymore.

@pkarw There is no "lang blinking" anymore since I `require()` the English language AND the defaultLocale language in the theme.

pkarw:
@DavidRouyer it still can be a problem as user can use LanguageSwitcher to switch to any of the `storeViews` - this switch reloads the page: https://demo.storefrontcloud.io -> https://demo.storefrontcloud.io/de 

And then - when `de` is not defaultLocale - there will be blinking anyway. So we need to preload all the langs that are in the `config.storeViews.i18n` as it was done before. Please not remove this feature - it's necessary to keep it.

You Comment to @filrak is not apropriete because we're RELOADING the page when swithcing the lang (it's required to reload it by many reasons - clearing the cache/memory objects related to current locale is one of them)

PR #2046, opened 2018-12-01
fix infinity loop on product page

patzick:
### Related PR

Follow up of [#2037](https://github.com/DivanteLtd/vue-storefront/pull/2037)

In product page there was an infinity loop on checking available product options.
After debuging it turns out, that configureProductAsync is changing product reference. That's why checking available options is fired in infinity loop.
Safe and simple solution is to deep copy product object.


pkarw:
If not object.assign then maybe some copy function from lodash package (please use lodash-es and import just the required function because of the tree shaking in such a case)

pkarw:
Moreover - this function is used when switching filters on category page to update the tile colors I’m afraid that after doing this copy on object we’ll broke this feature as the product tile assigned products won’t be modified 

Please check ProductTile.js how it calls configure action on product.

I think that maybe we should:
A) do this copy when a proper argument is set to true like “immutableProduct” = true / false by default,
B) perform a copy on Product.js level before calling configureProductAsync - probably it’s simpler / more elegant however doing A is a signal for developers that they may have the same issue in future 

Please check it and choose the best option @patzick

patzick:
@pkarw `Object.assign` creates only shallow copy of object, could not work in this case changes are deeper in object. Also lodash deepClone compared to JSON serialization is way much slower. Lodash deep copy is good fit for copying JS objects (with some functions etc.), not JSONs.

Sure, we can extract it to some util, do you want me to move it to core helpers?

patzick:
i'll look out to this case that you mentioned

pkarw:
In that case (I really don’t like serialization as a solution for copying object) maybe we just need to refactor configureAsynx and extract small method for checking if the configurable child does exist - use it in configureAsynx and separately in our case in Product.js isOptionAvailable. Would Ben probably most elegant option

pkarw:
One more thing. There is a possible case when none of the variants is available. In this situation we should set ‘currentProduct.errors.variants’ with the proper error message. Check this line for a reference 
https://github.com/DivanteLtd/vue-storefront/blob/589d07c1997b32263ac49edceb0c8033cd0a6abf/core/modules/catalog/helpers/index.ts#L20

This is the filter for unavailable product variants (based on stock quantities). 

Side note: This visibility filter we’re discussing right now is executed AFTER this filter - therefore we can use ‘configurable_children.stock’ item as it’s prefetched 
https://github.com/DivanteLtd/vue-storefront/blob/589d07c1997b32263ac49edceb0c8033cd0a6abf/core/modules/catalog/helpers/index.ts#L55

Anyway, by setting product.errors.{anyKeyYouWant} You’re blocking the ‘Add cart button’ and it’s vuex action - 
https://github.com/DivanteLtd/vue-storefront/blob/589d07c1997b32263ac49edceb0c8033cd0a6abf/core/modules/cart/store/actions.ts#L240

Plus:  
https://github.com/DivanteLtd/vue-storefront/blob/589d07c1997b32263ac49edceb0c8033cd0a6abf/src/themes/default/components/core/AddToCart.vue#L22

(Side note: this action should be named “cantBeAdded” as it’s used for “disabled=canBeAdded()“ reversed logic :-))

@NataliaTepluhina can You add some note on “product.errors” (take a look at the snippets above) to the docs? Not sure where - maybe FAQ, or maybe we should have separate section called “Processes” either “Use cases” and there “Add to cart” where we should describe how the process works in details:
- Vuex action is executed, 
- errors are being checked on the product,
- product is added to localStorage,
- then the Cart synchronization is being executed
..

(Of course this process description is simplified - to describe it in details You need to went thru the whole process step by step (possible with debugger)). So please decide how You like to add it to the docs, but I belive it’s a important note as using product.errors user can add different validations as well

Adding @filrak, @vue-kacper, @mdanilowicz  - I’ll be adding You guys to such a informative comments to better share the knowledge how “backend” works ok?

patzick:
@pkarw thank you for these informations; i was considering both options and i think A was a better one, because this problem could return in future and this flag gives developers more awareness that their product property might change.

I agree that serialization isn't the cleanest solution, thought about that too and i've presented some idea in #2048, please consider it as well. Hope that we can talk about that too.

patzick:
About that `ProductTile`. It’s really not clean solution. Every component of productTile is listening of `product-after-priceupdate` event and then it mutates it’s prop. That should never happen.
I’ve tracked all ProductTile invocations to find all sources of products. And they are all from store, like:

* product.list.items
* recently-viewed.items
* product.current_product
* product.related
* homepage.new_collection
* compare.items
* homepage.bestsellers

And also there are two cases - in search panel and slider - that source is result of `products/list` dispatch.

I think that ProductTile shouldn't listen to this event. And especially should never mutate it’s props.

patzick:
@pkarw because extracting suggested code won't prevent us from mutating product object. Just first forEach:
https://github.com/DivanteLtd/vue-storefront/blob/589d07c1997b32263ac49edceb0c8033cd0a6abf/core/modules/catalog/helpers/index.ts#L434-L443
changes it's `configurable_children` params. 

So lets get back to this tomorrow :)

pkarw:
@patzick as the release is close and this refactor was apparently more heavy that I initially thought - I've just finished it by myself - please check it out: https://github.com/DivanteLtd/vue-storefront/commit/141d71938e93d7fdcd75f0ff861ff0b1e2acd57e and test if everything works just fine. By this chance, I've improved the general error messaging per variants availability

Issue #1947, opened 2018-10-30
Base authentication on navigation guards instead of current behavior


filrak:
## What is the motivation for adding / enhancing this feature?

Right now we have logic responsible for checking if user is authenticated before entering routes splitted across components. It implies undesired behavior like this ( https://github.com/DivanteLtd/vue-storefront/issues/1945 ), introduces a lot of potential bugs and adds complexity to the project.

Moreover we will always redirect to home for non-SSR pages with this behavior which isn't what we want.

We should base this authentication checking on navigation guards and routes meta.
Ideally we should specify in `routes` object of `vue-router` in `meta` property if route requires authentication. 

Then in `beforeEach` navigation guard we should check if this property is set to `true`. If so we should check if there is a token in local storage to authenticate user session -> wait for authorization -> navigate to proper route.  If token is not present redirect to homepage with notification 'You need to be logged in to see this page'.

This task is rather complex but extremely important in terms of removing unnecessary complexity in our app.

## What are the acceptance criteria 

- Add route guard based authentication like described
- Remove all current authentication checks from pages and vuex stores

# Can you complete this feature request by yourself?

Unfortunately I'm busy doing modules refactor
## Additional information

In case of any questions please reach me on slack.
This article may be helpful (`Authenticated routes` section describes exactly what I'm talking about)  https://blog.sqreen.io/authentication-best-practices-vue/

Issue #1945, opened 2018-10-30
Refresh on user profile redirects to home page


filrak:
## Current behavior

1. Visit user profile (any section)
2. Refresh
3. You are redirected to home
## Expected behavior

We should stay in the same page we refreshed
## Repository

https://demo.vuestorefront.io/
## Enviroment details

- Browser:  GH68
- OS:  MacOS HS
- Node:  8.x

pkarw:
It's not a bug. SSR is not working well with Account nad we wanted to avoid situation when the user profile is loaded without user being authroized.

filrak:
okay, can we have a task to fix this? I can imagine some situations where merchants can have direct links to account in emails etc. We should check if user is authorised in navigation guards and decide on redirects from there

PR #2047, opened 2018-12-01
Refactor auth routes guard

jahvi:
### Related issues

#1947

Initial implementation of auth route guard refactoring, it should move the auth check logic out of components into navigation guards at the router level using a new `requiresAuth` meta options.

This is just an initial PoC to get some feedback, at the moment it just "works" on the "My Account" page only.

I wasn't sure where to put the `beforeEach` router check so for now it's on the `src/themes/default/index.js` file so it works globally, let me know if there's a better place for it.

Also the guard works when you navigate to the `my-account` route client side, if you reload the page or try to go to the URL directly then the `beforeEach` function in the router is not called, still trying to figure out how to fix this.

N/A yet


jahvi:
> Also the guard works when you navigate to the my-account route client side, if you reload the page or try to go to the URL directly then the beforeEach function in the router is not called, still trying to figure out how to fix this.

Looks like I'm registering the `beforeEach` check too late, should I move it to `core/app.ts`?

src/themes/default/index.js
```diff
@@ -8,4 +8,26 @@ export default function (app, router, store, config, ssrContext) {
   // { name: 'de-checkout', path: '/checkout', component: CheckoutCustomized },
   setupMultistoreRoutes(config, router, routes)
   router.addRoutes(routes)
+  router.beforeEach((to, from, next) => {
+    if (to.matched.some(route => route.meta.requiresAuth)) {
+      const usersCollection = app.$db.usersCollection
```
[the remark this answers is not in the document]
jahvi: Can you show me an example of this? I couldn't find any reference somewhere else in the app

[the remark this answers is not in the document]
jahvi: NVM just saw how to do it from the wishlist module 🙂

jahvi:
I moved the guard logic to the `core/app.ts` file for now until the 1.7 refactor just to see if it works. It seems to do the right thing however when you are logged out and try to access the `/my-account` page directly then the page "flickers" for a second and the redirects back to the homepage:

motion

Still looking into it but it's as if the guard is still running too late.

filrak:
@jahvi i guess its a standard behavior since the redirect to homepage happens. We can ommit it just by not invoking `next()` at all but i'm not sure if it won't make direct urls not working

core/app.ts
```diff
@@ -105,6 +126,7 @@ function createApp (ssrContext, config): { app: Vue, router: VueRouter, store: S
   store.state.storeView = storeView
   // store.state.shipping.methods = shippingMethods
 
+  Vue.use(VueRouter)
```
filrak: not sure if ssr will work as intended after moving this to createApp, have you tested?

jahvi: This seems to work as far as I can tell, the reason I moved this is because I think I need to call `use` after `beforeEach`. Not sure if this is the best place to put it though.

jahvi:
The problem with the flickering looks a bit more involved, it looks like when you reload `my-account` or try to visit the URL directly the router tries to load the page initially and then it performs the redirect if not logged in.

I added a console log to the `beforeEach` function to log the `to` variables I got the following in the console:

screen shot 2018-12-09 at 7 44 40 pm

The first log shows a route with no `name` and `meta` keys which is why the guard doesn't fire, then it logs it a second time and this is when the guard runs and redirects to the homepage. If you navigate to the account page with a link or any other way it works fine.

I haven't been able to debug why the `to` route is null on page load initially.

jahvi:
I moved the route guard implementation to the user module, hopefully it will make things easier to debug.

core/app.ts
jahvi: I moved the logic to the user module so I reverted the `app.ts` file back to its original state.

filrak:
Thanks @jahvi . I'll test it today and merge if everything is fine!

jahvi:
Making some progress here, I realised that the flickering is because I forgot to remove this redirect from the `core/pages/MyAccount.js` component:

```
if (context) context.server.response.redirect('/')
```

This redirect should be handled by the guard now, and I changed the guard to use the vuex store so it can be accessed client and server side easily.

However there's another problem, I haven't been able to find a way to check if the user is logged in server side so when you reload the `my-account` page or navigate directly this line is always false `rootStore.getters['user/isLoggedIn']` so you will always get redirected to the homepage...

Is there a way to check if the user is logged in server side? Should I change the approach here instead?

pkarw:
@jahvi the authorization service is available  only client side. Serverside - the user is always unauthorized and this redirect works just on server side so shouldn’t we just keep it?

jahvi:
@pkarw Well the `beforeEach` guard also fires server side so it still redirects to the homepage, my concern is that since we can't check for auth then issues like #1945 will still happen.

pkarw:
Correct. The only thing we can do is not redirecting in the ssr but it can cause displaying the blank account page with no data (as it’s loaded later on).

I belive we can just add a config option if the user like to have this ssr redirect (default on) or not (then to be guarded client side but can do Refresh)

Does it make sense?

jahvi:
Makes sense, as an alternative would it be worth considering storing the user token in a cookie? That way it can be used client and server side and avoid the issues mentioned above.

pkarw:
No, we shouldn't use cookies -> it will complicate the SSR cache, the scalability of the app (the usage of the caching proxies will be limited as they're usually not cache any pages containing cookies)

jahvi:
I see it makes sense, let me refactor it a bit then.

filrak:
@jahvi what is the status?

jahvi:
Looking into it again today, sorry it took so long.

jahvi:
Just trying to finish this one up but ran into another bug, when you login and reload a page (any page, not just auth protected ones) I get the following error in the console and the page doesn't render:

screen shot 2018-12-19 at 1 33 24 pm

Still trying to figure it out but if anyone can point me in the right direction it would be great.

filrak:
@patzick was working on hydration problems recently, maybe he could help

patzick:
@jahvi  the hydration bug should no longer be a problem on develop branch. But let me know if something like this would happen again

jahvi:
@patzick Thanks I just rebased develop and everything seems to work now.

I was trying to make the SSR change but I feel like the current behaviour (always redirect to homepage on the server) is the most efficient solution, at least until we have a way of checking authentication on the server.

Having auth pages display blank content, even behind a config option, doesn't feel like a good user experience, what would be the point of being able to get to this pages if users won't be able to see the auth content anyways?

Having the redirect, while not pretty, would mean:

- Make VS behave the same as it does now, so we're not introducing a new bug or anything.
- In most cases we don't even want to search engines to index auth pages so the benefits of SSR on those are minimised. 

Maybe I'm missing something but it's just my opinion for now, I can make sure to also document all this in the docs so users are aware.

filrak:
@jahvi I think its ready to merge with current state, @pkarw also accepted (ill do it when come back from vacation on 6th). Thank you for this exhaustive PR man and sorry you had to wait so long!

Issue #2048, opened 2018-12-02
[RFC] Refactor vuex store idea


patzick:
Guys, I’ve been thinking of some possible changes and I’m coming up with an idea. I’d like to discuss with you pros and cons of this solution, thats why I'm writing this Request For Comments. Hope this will be an educational discussion for all of us. It is not meant to be long, it’s more like PoC for reorganising things.

For deeper background documentation of current product module can be found [here](https://divanteltd.github.io/vue-storefront/guide/vuex/product-store.html).

I’ve noticed, that we have numerous challenges with synchronising data in application. 
Problems i’d like to address to:
* not using vuex getters - everything from vuex should be accessed by getters. There shouldn’t be a single `$store.state` or `mapState` in components.
* mutating state from components - it’s like with state - every mutation should be invoked from vuex actions. This way we can keep clean logic and there’s a possibility to change behaviour or add functionalities without breaking things.
* multiple objects instances - it’s what my idea is all about - we have a list of products, we have current product,  products in cart, related products, parent product, products in wishlist….and now we got product price update or its quantity - we have to update all instances of product which is a really complex operation. Tiny mistake can cost a lot of time to be founded.

Heres my PoC for organising vuex. I’ve merged in it options from different modules, but only for purpose of showing connections.
Store:

```
{
  "products": {
    "WS08": {
      "sku": "WS08",
      "name": "Minerva LumaTech&trade; V-Tee"
      // other options
    },
    "WS12": {
      "sku": "WS12",
      "name": "Radiant Tee"
      // other options
    },
    "WS08-XS-Black": {
        "sku": "WS08-XS-Black",
        "name": "Minerva LumaTech&trade; V-Tee"
        // other options
    }
    // maaaaaaaany more products
  },
  "currentProduct": "WS08-XS-Black",
  "wishlist": ["MP01-32-Black", "MSH05-32-Black"],
  "cartItems": [
    {
      "sku": "WH09-XS-Green",
      "qty": 3
    },
    {
      "sku": "WH09-S-Red",
      "qty": 1
    }
  ]
}
```

As you can see my intention is to have only one instance of each object and  address them by ids. Then getters could look like this:

```
{
  getProductsMap: state => state.products,
  getCurrentProduct: state => state.products[state.currentProduct],
  getWishlist: state => state.wishlist,
  getWishlistProducts: state => state.wishlist.map(productId => state.products[productId]),
  getCartProducts: state => cartItems.map(cartItem => {
[...]

PR #2049, opened 2018-12-02
Refactor category vuex module. Use getters and actions

patzick:
### Related issues

#2048 

For the first step of #2048 checklist i took `category` vuex module.
I've created getters, actions and mutations and changed all state references to this getters, and every state direct object change to vuex action.

It doesn't change any logic (except adding available filter - i'm adding whole filter with its options, not per option), just separates layers responsible for changing store.


core/compatibility/components/blocks/Category/Sidebar.js
```diff
@@ -24,7 +26,7 @@ export default {
       this.$bus.$emit('filter-reset')
       this.$store.dispatch('category/resetFilters')
       this.$store.dispatch('category/searchProductQuery', buildFilterProductsQuery(this.category, this.activeFilters))
-      this.$store.dispatch('category/products', this.$store.state.category.current_product_query)
+      this.$store.dispatch('category/products', this.getCategorySearchOptions)
```
pkarw: Should be getCategoryCurrentProductQuery

core/modules/catalog/store/category/actions.ts
```diff
@@ -268,6 +268,10 @@ const actions: ActionTree<CategoryState, RootState> = {
                 }
               }
             }
+            context.dispatch('addAvailableFilter', {
```
pkarw: Shouldn’t be a mutation?

core/pages/Category.js
```diff
@@ -92,9 +94,11 @@ export default {
           reject(err)
         }).then((attrs) => {
           store.dispatch('category/single', { key: 'slug', value: route.params.slug }).then((parentCategory) => {
```
pkarw: We should move all these vuex calls to vuex action and separate it from component properties itself like it’s currrntly (almost) achieved in Product.js

I mean: we should have as thin ui components as we can as it gives us a possibility to better separate Ui and business logic in the next rrleases

core/modules/catalog/store/category/actions.ts
```diff
@@ -147,7 +147,7 @@ const actions: ActionTree<CategoryState, RootState> = {
    * Filter category products
    */
   products (context, { populateAggregations = false, filters = [], searchProductQuery, current = 0, perPage = 50, sort = '', includeFields = null, excludeFields = null, configuration = null, append = false, skipCache = false }) {
-    rootStore.state.category.current_product_query = {
+    context.dispatch('setSearchOptions', {
```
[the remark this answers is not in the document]
patzick: maybe i'll explain my thoughts - if we ever want to add some other behavior when setting searchQuery, then we would do that in single action. Mutation is also okey, especially if it's the same vuex module. I won't argue on that, just wanted to let you know why i've used action.

core/compatibility/components/blocks/Category/Sidebar.js
```diff
@@ -9,11 +10,12 @@ export default {
     }
   },
   computed: {
+    ...mapGetters('category', ['getCurrentCategory', 'getActiveCategoryFilters', 'getCategorySearchOptions']),
     category () {
-      return this.$store.state.category.current
+      return this.getCurrentCategory
```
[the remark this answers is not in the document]
patzick: if we shouldn't change naming it should be then just `getCurrent`, but i added name of module in getters to better finding all usages and not repeat getters from other modules. When you implement this getter it's like:
```
...mapGetters('category, ['getCurrentCategory'])
```
so you already have context for module. Just my thought is - its not in english - it answers to question `what's the current category?` not `what's the category current?`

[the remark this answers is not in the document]
pkarw: OK I got it. Let's start with naming convention first to avoid such a discussion in the future ok?

core/compatibility/components/blocks/Category/Sidebar.js
```diff
@@ -9,11 +10,12 @@ export default {
...
     },
     activeFilters () {
-      return this.$store.state.category.filters.chosen
+      return this.getActiveCategoryFilters
```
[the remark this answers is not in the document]
patzick: my explanation is like with `current category`

core/compatibility/components/blocks/Category/Sidebar.js
patzick: ok thanks - i was wondering how to approach this. I havn't name it `query` just because it's not string but object; but i thought that i'll commit that and ask today for advice :)

but again - not sure about that naming convention. It's start of big store  refactor and i think we should discuss it first:)

core/compatibility/components/blocks/Category/Sidebar.js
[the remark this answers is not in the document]
patzick: Sure, i agree with you :)

core/modules/catalog/store/category/actions.ts
filrak: in our case I'm strongly against plain mutations exposed for users. General rule is to use action instead of mutation if you think that any additional logic may be added in the future It's better to have future-proof action rather than changing the API again ;)

core/modules/catalog/store/category/actions.ts
```diff
@@ -305,11 +309,20 @@ const actions: ActionTree<CategoryState, RootState> = {
     }
     return productPromise
   },
+  addAvailableFilter ({commit}, {key, options} = {}) {
```
filrak: shouldn't be addFiler? can we add unavailable filters?

I'm not sure if it's responsible for adding filter to the list of available filters or using the filter to filter date. If 2nd option It should eb sth like `useFilter`

core/modules/catalog/store/category/getters.ts
```diff
@@ -3,8 +3,22 @@ import RootState from '@vue-storefront/store/types/RootState'
...
+  getAllCategoryFilters: state => state.filters,
+  getActiveCategoryFilters: state => state.filters.chosen,
+  getAvailableCategoryFilters: state => state.filters.available,
+  getCategorySearchOptions: state => state.current_product_query,
+  getCategories: state => state.list,
+  getCategoryBreadcrumbs: state => state.breadcrumbs,
+  /**
+   * @deprecated use getCurrentCategory instead
+   */
+  current: (state, getters) => getters.getCurrentCategory,
```
filrak: Please add Logger warning (core/lib/logger) for usage of depreciated apis

filrak:
This branch should not be merged before 

- [ ] naming convention specified and added to our docs
- [ ] 1.6 release

pkarw:
@filrak afet apllying my suggestions: yes. We must minimize the number of changes

core/modules/catalog/store/category/actions.ts
patzick: @filrak in state we have something like this:
```
filters: {
      available: {},
      chosen: {}
    },
```
so it's in fact adding available filters, because you can also add chosen filters. I didn't want to change any logic, only to mutate state through actions and mutations :)

filrak:
@pkarw @patzick is it ready to be merged? (after resolving conflicts)

filrak:
@pkarw since we have feature freeze now this will be merged in 1.8

pkarw:
@filrak sure.

patzick:
@filrak , @pkarw i've updated dosc and relase notes.
It's not a feature itself, more like refactor and also backward compatible, so safe to add this in current release.

pkarw:
Thanks! I’m fine with this but it’s up to @filrak if to merge this in 1.7 either 1.8

PR #2067, opened 2018-12-07
[WIP] Lazy vs modules

filrak:

This is a PoC of lazy loaded VS module. Some of non-critical modules will be written following the same pattern. It allows us to lazily load whole VS module and component for a given module (in this case wishlist)

Avoiding global registration of VS modules and putting this registration into components themeselves will also allow us to properly split VS modules into route chunks. 

How it works:
1. Module registration is removed
2. Module registers itself if any on it's components (except the ones responsible for opening it like WishlistButton) is visible on current view (it registers itself on created hook) so if we have Product page with AddToWishlist component (lazily loaded) it will be loaded with whole module and it's created hook will register it
3. If you want to use lazy vs modules store (like wishlist) without components you just need to register it before.

Example:
While we entering Home Page there are no components using Wishlist module therefore it's not downloaded:
image
When we click on WIshlistIcon the Wishlist component (along with module) will be lazily downloaded 
image

I have plans to make most of VS modules work like this. This does not imply that all of them will be lazy loadable but moving registration to components ensures that if they are not lazy loaded directly - they'll be properly distributed into route chunks.

### DIFF
--- a/core/app.ts
+++ b/core/app.ts
@@ -60,9 +60,9 @@ function createRouter (): VueRouter {
   })
 }
 
-function registerModules (modules: VueStorefrontModule[], store: Store<RootState>, router: VueRouter, context): void {
+function registerModules (modules: VueStorefrontModule[], context): void {
   let registeredModules = []
-  modules.forEach(m => registeredModules.push(m.register(store, router)))
+  modules.forEach(m => registeredModules.push(m.register()))
   Logger.info('VS Modules registration finished.', 'module', {
       succesfulyRegistered: registeredModules.length + ' / ' + modules.length,
       registrationOrder: registeredModules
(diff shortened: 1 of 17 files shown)


pkarw:
Wow, nice idea!

patzick:
@filrak that's a really cool idea! :)

Have one question - is `.register()` a sync or async method? It downloads file from server, so maybe it could be better if it would return a promise? This way we won't block UI if there is problem (for example slow download)

filrak:
@patzick `register()r` itself is not invoking any download process. The downloaded part (components + VS module) usually contains register function just to register downloaded vuex store and hooks.

In other words this function is used only to register module from already downloaded chunk. Sorry if my explanation was misleading ;)

This method main purpose is to allow route-level module splitting (or lazy loading components+modules for specific use cases like wishlist which is useful only when opened no matter of route) instead of global registration in one file (and therefore including all modules in main bundle)

filrak:
To put more into context. This is modules javascript that we can split into route chunks 
image

PR #2068, opened 2018-12-08
Feature codesplitting per route

patzick:
#### Note
It's new copy of [2062](https://github.com/DivanteLtd/vue-storefront/pull/2062). Accidentally merge there another changes so here's clean version of PR.

Code should be invoked only which is needed per route. The rest of if should be prefetched.

By default i prefetched:
- home page
- product page
- category page
- checkout page - as it was prefetched before, but another way

On demo page, served from server after build whis should be even better, but just on local is an improvement:

homepage before and after
home_before
home_after

category before and after
category_before
category_after


pkarw:
Cool! @filrak WDYT?

patzick:
After resolving hydration problems, results are even better on PWA and SEO :)

For homepage:
homepage_v2

For category page:
category_v2

pkarw:
Awesome! By this chance can you also check the service worker registration code? Currently it’s registering the service worker ONLY on home page - it’s because the given service worker url is relative :/ it’s probably in service-worker-registration.js or in client-entry.js

Anyway: it’s cool you’ve solved the hydration problem

filrak:
This pull request deserves a bottle of vodka :D

patzick:
@pkarw sure, i'll try to fix this (already did some digging), i think it's better to work on this in another PR

@filrak thanks buddy, but you've made a lot better upgrades! ;D

PR #2092, opened 2018-12-14
Feature - async/await support

patzick:

I've added babel-polyfill to our build process, mostly to support async/await.

After full integration with babel7 this won't be necessary but we can white readable async code right away! :)

I'll show Home.vue asyncData, because it's not so clear in diff mode
so before:
home_before

aaand after:
home_after

### DIFF
--- a/core/build/dev-server.js
+++ b/core/build/dev-server.js
@@ -16,7 +16,7 @@ module.exports = function setupDevServer (app, cb) {
   let template
 
   // modify client config to work with hot middleware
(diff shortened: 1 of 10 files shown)


filrak:
Do you have before/after bundle sizes with this polyfills?

patzick:
@filrak Sure, i've done bundle analyzer. It's adding ~20kB for now, but first - it's only temporary workaround till we get full babel7; second - we need to work on vendor itself because it's huge anyway; and third - i think it's totally worth it for code readability like up above.

screen shot 2018-12-16 at 18 04 52

PR #2109, opened 2018-12-22
WIP: Performance optimization

filrak:
- lazy wishlist module
- meta tag for easier VS identification (not performance related)
- lazy loaded vue-carousel
- global vendor libs splitted from route-specific  - currently most of libs in `app.js` can be lazily loaded or even removed
- google analytics module with it's dependencies is now optional and not included into main bundle (it wasn't connected to GA anyway)

- bundleAnalyser is ofc temporary here and will be removed before merge ;)


core/build/webpack.base.config.js
```diff
@@ -5,7 +5,7 @@ const CaseSensitivePathsPlugin = require('case-sensitive-paths-webpack-plugin')
 const VueLoaderPlugin = require('vue-loader/lib/plugin')
 const autoprefixer = require('autoprefixer')
 const HTMLPlugin = require('html-webpack-plugin')
-// const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
filrak: temporary until WIP ;)

src/themes/default/pages/Home.vue
```diff
@@ -72,57 +72,44 @@ export default {
     // Load personal and shipping details for Checkout page from IndexedDB
     this.$store.dispatch('checkout/load')
   },
-  beforeMount () {
+  async beforeMount () {
```
filrak: result fo resolved merge conflicts @patzick @pkarw please check if everything is correct

patzick: for me it looks okay :)

core/lib/module.ts
```diff
@@ -15,7 +16,14 @@ export interface VueStorefrontModuleConfig {
   afterRegistration?: (Vue?: VueConstructor, config?: Object, store?: Store<RootState>, isServer?: boolean) => void,
 }
 
+const moduleExtendings = []
```
patzick: should be let, not const, because is modified later

src/themes/default/components/core/ProductGalleryZoom.vue
```diff
@@ -50,8 +50,8 @@ export default {
   },
   components: {
     'no-ssr': NoSSR,
-    Carousel,
-    Slide
+    'Carousel': () => import('vue-carousel').then(Slider => Slider.Carousel),
```
patzick: we should discuss this
because it is showed on initial view. There is no v-if to delay it's load so i think it should stay statically loaded

filrak: but the load is delayed to the point when initial js is parsed and executed so the preformance gain is still valid

Issue #2116, opened 2018-12-27
Cart Configurable item pulled from magento loaded as Simple


valeriish:
## Steps to reproduce the issue

1. Login-in as customer
2. Add configurable product to cart
3. Examine minicart, new item have parent name, image, url, type = configurable
4. Login-out and login-in

## Expected behavior
Cart item have type 'configurable'.
Name, image, etc are getting from assigned parent product

## Current behavior
Cart item have type 'simple'.
Cart item displays name, image, etc  of 'not visible individually' current simple product

# Can you handle fixing this bug by yourself?
Yes
Issue: cart item is loading by serverItem.sku after ServerPull for configurable item

pkarw:
Thanks, how would You like to fix this one? Please feel free to report a Pull Request

pkarw:
Correct. Method “single” is not using the “childSku” at all as far as I remember (not being in front of my computer to check it right now ...) I think it makes sense to fix it at the origin - in the single action. However your fix is also just fine.

By this chance you might want to add code comments on how to use the action parameters correctly wdyt?

valeriish:
This is workaround, i tried to update 'product/single' method, but it's affect on other logic.

pkarw:
Thanks I’ll take a look at it later on I believe that we just should add a check if there is options.childSku set and pass this argument to configure some way but I must take a look at it

Your desired behavior is that call to single action will load parent product pointed out by ”sku” and them configure it to ”childSku” variant. Correct?

valeriish:
Yes, you're correct

pkarw:
OK I found the problem and added an option `assignDefaultVariant` to have this side effect of the `single` - I mean the configured product as a result.

Now its: 
`single (context, { options, setCurrentProduct = true, selectDefaultVariant = true, assignDefaultVariant = false, key = 'sku', skipCache = false })`

and in Your case You should call sth like:

```js
dispatch('product/single', {
          options: actionOptions,
          assignDefaultVariant: true
        })
```

valeriish:
servercartAfterPulled method was updated please review

PR #2119, opened 2018-12-28
Fix Cart Configurable Item pulled from magento

valeriish:
### Related issues

#2116 

Fix Cart Configurable Item pulled from magento


pkarw:
That one was pretty clever! :) Thanks!

Issue #2120, opened 2018-12-28
Optimize the picture switching when entering product page 


pkarw:
When you enter the product page from category listing filtered by some color:
https://test.storefrontcloud.io/c/men

The visible effect is switching the product to the right gallery element (regarding selected color).

The right picture should be displayed immediately after entering product page:
- the slider should be set to proper element initially before page is loaded
/ 
- or maybe we should use <img tag displaying the right picture and only after image gallery is loaded - to be switched to the dynamic component.

PR #2123, opened 2019-01-01
bugfix/2120-picture-switching-when-entering-product

vue-kacper:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/2120

Optimized picture switching when entering product page


pkarw:
Thanks! I have merged this in however in #2120 I was thinking rather about kind of carousel lazyloding - I mean loading single static picture then loading the carousel itself -
Because this way it should positively affect the performance of first page load for the product

vue-kacper:
@pkarw sure :) since we are planing to change vue-carousel component in https://github.com/DivanteLtd/vue-storefront/issues/2111 it would be better to implement lazy loading based on new carousel plugin

PR #2133, opened 2019-01-06
fix multiple placing order invocation after changing payment methods

patzick:
### Related issues

This resolves multiple order placing.
After each payment method change there was new place order binding, so on checkout iw was invoked multiple times.


pkarw:
as far as I understood the problem beret was that this module supports multiple payment methods (so the event has been bound multiple times). Could you just to make sure check if the https://github.com/develodesign/vsf-payment-paypal/blob/master/hooks/beforeRegistration.ts isn’t also affected by this issue? I think it’s not a case but better be safe than sorry I’m case of payments :-)

@patzick

patzick:
@pkarw good suggestion. I've made PR there. Now we'll be sure that nothing like this would happen there.

PR #2149, opened 2019-01-09
Added bash command for collecting i18n phrases.

qbo-tech:

Added bash command for collecting i18n phrases into Recipies docs.


pkarw:
Thanks!

PR #2155, opened 2019-01-10
lazy load orders confirmation

patzick:
### Related issues

Lazy load order confirmation


filrak:
Isn't whole checkout lazily loaded anyway?

patzick:
it's modal with order confirmation which is loaded on event `offline-order-confirmation` so it's unnecessary for initial render. It saves ~40ms of scripts evaluation.

PR #2164, opened 2019-01-11
WIP: Dynamic module loading

filrak:

Now VSModules can be seamlesly lazy loaded. The dynamic loading is based on actions. 
How it works:
- somewhere in the theme we are dispatching action e.g `wishlist/load`
- VS is checking if module `wishlist` is registered
- if it is it's just dispatching the action (normal behavior) 
- if it's not it lazy loads module, registers it and dispatches the action

With this approach VS modules can be lazily loaded seamlessly and users don''t need to take care of the registration themselves. They just need to register module as a lazy one in `modules.index.ts` and vs will take care of lazy loading it in a right moment. This will end up with HUGE reduction of bundle size.

To add
- dynamic loading based on getters.


filrak:
@pkarw this feature needs to be finished and polished tho so this PR will not be merged into 1.7

pkarw:
@filrak OK

Issue #2244, opened 2019-01-18
Docs: Fix images on guide/installation/magento.md


wilfriedwolf:
## Current behavior
Image links are broken

## Expected behavior
They should appear

## Repository
guide/installation/magento.md

# Can you handle fixing this bug by yourself?
YES

PR #2248, opened 2019-01-18
corrects issue #2244

wilfriedwolf:
Just found the images and corrected magento.md


pkarw:
Thanks @wilfriedwolf! This is awesome. @filrak after You’ve removed the `vue-storefront` dir from the docs it’s worth checking if some other places in the docs don’t require similar changes?

filrak:
@pkarw right. Added to my todo list :)

wilfriedwolf:
I checked the whole docs for ![ tags and corrected everything, so I guess, all images should be ok?

filrak:
@wilfriedwolf thank then! Merging it in  ;)

PR #2264, opened 2019-01-19
Hotfix: fixed category filters reset functionality

vue-kacper:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/2262

Fixed category filters reset functionality (on mobile)

### DIFF
--- a/core/compatibility/components/blocks/Category/Sidebar.js
+++ b/core/compatibility/components/blocks/Category/Sidebar.js
@@ -26,7 +26,7 @@ export default {
       this.$bus.$emit('filter-reset')
       this.$store.dispatch('category/resetFilters')
       this.$store.dispatch('category/searchProductQuery', buildFilterProductsQuery(this.category, this.activeFilters))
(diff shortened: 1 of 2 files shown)


pkarw:
@vue-kacper look at this one 
https://github.com/DivanteLtd/vue-storefront/blob/6c100f978aa79975e4db22be3cefa7f8d38b4c97/core/pages/Category.js#L205

Works with the same usage of the same getter. Seemingly the problem is because in this case getCurrentCategoryQuery.searchQuery is an empty object.

I mean - if this is true this fix can be misleading and it would be better to have explicitly:

searchProductQuery: {}

Do you know what I mean? I mean that searxhLroductQuery is probably a property of the getter value - so by solving this issue this way you’re doing potentially a circular refeeence which could be misleading for other deva reading the code @patzick was refactoring the getters in here so may have something to add - as the problem was introduced with this refactor (I mean it works in 1.6)

Thanks for such a quick fix!

vue-kacper:
@pkarw Aww.. you are right, it would be much better to fix it at getter level

@patzick I'm leaving it to you then :)

PR #2275, opened 2019-01-22
Correctly merge the store modules array with extended module config

DaanKouters:
### Related issues
https://github.com/DivanteLtd/vue-storefront/issues/2274

The store modules array gets merged as intended by this fix. Previous the merge would have unexpected results

@filip-modlinski i believe the _extend function can be refactored to a bit cleaner code. 
I delete the store property in both configs before the merge to get a clean object for the other properties like the beforeRegistration + afterRegistration methods
After that i merge the modules arrays and put them back in de store property of the module. May you know a cleaner way to do this.

I used a custom es index, my e2e test failed because of this

It works now as in the documentation is described.

### DIFF
--- a/core/lib/module.ts
+++ b/core/lib/module.ts
@@ -2,6 +2,8 @@ import { Module, Store } from 'vuex'
 import { RouteConfig, NavigationGuard } from 'vue-router'
 import Vue, { VueConstructor } from 'vue'
 import merge from 'lodash-es/merge'
+import some from 'lodash-es/some'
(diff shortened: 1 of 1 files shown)


filrak:
@DaanKouters thanks for the fix! The solution is as we discussed and yeah you're right that some parts of the modules lib should be refactored to cleaner code since they have grown a bit too much. I'll merge it as soon as it passes CI assuming you tested it by yourself

PR #2292, opened 2019-01-27
Changelog.md update

Igloczek:
I moved release notes from https://github.com/DivanteLtd/vue-storefront/releases to CHANGELOG.md file and add the list of changes that are currently pending on master branch waiting for release.

Although, I think that @patzick changes related to the home page are just visual changes, not hotfixes, so shouldn't be released as 1.7.x, so probably we need to recreate `master` branch cherrypicking only actual hotfixes and docs updates.


pkarw:
Thanks! With one except: I think we won’t do cherrypick on master but I can’t agree that we should avoid such a changes. There is a whole long story on why it has been applied to master :)

PR #2309, opened 2019-01-30
Output cache versioning

Igloczek:

Currently, the cache is persisted in Redis until the expiration, which can lead to issues with serving old content after deployment, if this cache will not be manually cleaned.

These changes are related to the build process and `redis-tag-cache` package setup.

In the first step during the server build, I generate the UUID and save it to `core/build/cache-version.json` file. 

Then when the server is firing up, it reads saved previously UUID and uses it to set `keyPrefix` which help us distinguish the correct version of the cache.

Why not generate UUID directly in `server.js`? We can spawn multiple instances of this server, so each of them will have on the cache, which is not efficient.

There are no changes required, just under the hood magic, but if someone was currently clearing all the cache during deployment, it's not necessary anymore.

### Questions
If manual cache clearing is not so useful anymore, should be kept `yarn cache` and `core/script/cache.js`?
If we should keep it, it also needs to read the content of `cache-version.json`, so maybe will be good to abstract the logic of creating cache instance 🤔

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ config/local.json
 var
 build/config.json
 core/build/config.json
+core/build/cache-version.json
 theme.js
 desktop.ini
 src/themes/catalog/resource/i18n.json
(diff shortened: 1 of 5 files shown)


pkarw:
That's cool. So we're clearing the cache with each build

Igloczek:
@pkarw what about the questions part? I'd like to make it consistent and easy to use.

pkarw:
I think that `yarn cache` is still useful and we should add the `cache-version.json` support to it

pkarw:
The `cache` command could be still useful as the cacheKey is regenerated only while buildinf the app

PR #2330, opened 2019-02-01
New Module Hooks API

filrak:
### Related issues

-

This change is meant to prevent future problems with before/after registration hooks in modules. Grouping it into one object allows much easier modifications and future extending than relying on params order. it's the very same approach that nuxt is using with their modules.

TODO:
- [x] Changelog

(describe in a few words what is this Pull Request changing and why it's useful)

-

-

- [ ] I read the [TypeScript Action Plan](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/TypeScript%20Action%20Plan.md) and adjusted my PR according to it
- [ ] I read about [Vue Storefront Modules](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md) and I am aware that every new feature should be a module


pkarw:
It's way better to pass the VSF object because it will allow us to extend `before` and `after` hooks without changing the definitions. 

However, there are currently some modules on Github created with the previous approach so we should be compatible with it (I mean with passing Vue, config .. as separate arguments). Would it be possible?

filrak:
@pkarw i already spoke withe xtension devs and will propose a proper PRs to their repos ;)

core/scripts/server.js
```diff
@@ -144,20 +144,87 @@ app.get('*', (req, res, next) => {
       next()
     }
   }
-
+  const HTMLContent = `<html lang="en">
```
patzick: i think this should be imported from separate file

core/modules/cart/hooks/afterRegistration.ts
```diff
@@ -1,6 +1,6 @@
 import * as types from './../store/mutation-types'
 
-export function afterRegistration(Vue, config, store, isServer) {
+export function afterRegistration({ { Vue, config, store, isServer } }) {
```
patzick: here we have double parenthesis

core/scripts/server.js
filrak: yeah, i had no idea it was included here btw, thought i emrged it before

PR #2367, opened 2019-02-04
Bugfix/2363

vue-kacper:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/2363

User account menu autohide on select (desktop)


src/themes/default/components/core/blocks/Header/AccountIcon.vue
```diff
@@ -2,10 +2,12 @@
   <div
     class="inline-flex relative dropdown"
     data-testid="accountButton"
-    @click.self="goToAccount"
+    @click.self="goToAccount();showMenu = true;"
     @keyup.enter="goToAccount"
     tabindex="0"
     role="button"
+    @mouseover="showMenu = true"
```
patzick: isn't this solve issue?
I'm little bit worried about above syntax
```js
@click.self="goToAccount();showMenu = true;"
```
it opens modal window or changes route to account, but mouse pointer is still over, does he?

vue-kacper: This is a fallback for touch devices where we won't get mouseover event.

Adding click event we are making sure that on touch devices "User menu" will appear correctly

patzick: Okay, thanks:)

vue-kacper:
@patzick updated :)

patzick:
Thanks! :)

Issue #2010, opened 2018-11-21
SEO friendly urls


ngeleedst:
## What is the motivation for adding / enhancing this feature?
The urls that are being generated in the Vue Storefront are not SEO best practices, we would like to use product slugs without sku. Next to this issue it would be nice to have some kind of url rewrite possibility because now you have to allocate every module to a route like:

Category: /category-slug
Pages: /page/page-slug
Product: /product/123/product-url/123

I think we need to develop a general entry point for all types and from that point match if the slug is for a category, page or product.

## What are the acceptance criteria 
- Product urls from /123/product-slug/123 to /product-slug
- Advanced routing module for url rewrites

# Can you complete this feature request by yourself?
No.

## Additional information
None

afirlejczyk:
- Advanced routing module for url rewrites
We have to remember about urls with extra slashes, e.g `women/tops.html` - this url also should work for category.

- there should be new option in config for this new feature

jahvi:
Is it worth leveraging/porting the `urlResolver` GraphQL query introduced in 2.3? from the docs:

> The urlResolver query returns the canonical URL for a specified product, category or CMS page

The query returns the type of entity (`PRODUCT`, `CATEGORY`, or `CMS_PAGE`) so the right component can be used depending of this.

**Example Query**
```graphql
{
  urlResolver(url: "venia-tops.html") {
    id
    canonical_url
    type
  }
}
```

**Result**
```graphql
{
  "data": {
    "urlResolver": {
      "id": 7,
      "canonical_url": "catalog/category/view/id/7",
      "type": "CATEGORY"
    }
  }
}
```

pkarw:
@madonzy we can probably store the `sku/url_key` mapping in kind of `localStorage` - populated in the [`product/list` method](https://github.com/DivanteLtd/vue-storefront/blob/fbe4b148e5f7044541018eaf49d122925d9a894b/core/modules/catalog/store/product/actions.ts#L327)

Please take a look at #2021 - where I've described exactly how offline cache works

jahvi:
@pkarw I see this sounds more complicated than I thought then, perhaps it might be too hard to implement with the current indexedDB approach.

jahvi:
I see, so you mean for example instead of having product keys in indexedDB as `sku$$MP01` change it to something like `url$$josie-yoga-jacket` and select based on that?

ngeleedst:
Any updates on this request?

[...]

PR #2401, opened 2019-02-08
[WIP] PoC - multipart, dynamic URL rewrite

pkarw:
### Related issues

#2010 

This is proof of the concept of dynamic URL mapper. The challenge is that we need to support the fully customizable URLs - that don't have any specific elements pointing out if that's the Product, Category either CMS page.

 Unfortunately, we CAN NOT use the vue-router guards as there is no way to `rewrite` URL (display different route/different component under the same URL). The only option was to create a page which is including the proper subcomponents.

We should decide if that's the way we would like to go. If so, we should then decide if this module should be placed in `core` or rather `src` - to let the users modify it when needed. After all -  I think it should be `src`. The Dispatcher can vary from project to project - mostly because some custom page types added.

If we decided that this is the way to go ->  I'll add the rest of the business logic.

Other options for URL dispatching mechanism:
- kind of refactor Product, Category, CmsPage pages - but in the end it comes up to kind of "multipurpose" component which is breaking the SR Principle
- Express.js level routing - but it doesn't solve the CSR routing isses (the URLs will differ CSR/SSR which is not good).
- ?? 

To test this out - please enter the http://localhost:3000/fake/product/url.html in the browser - and You'll get the `Gwyn Endurance` product page.

I've got an idea that the `url/registerMapping` will be called inside the `category/list` and `product/list` matching the products and categories with their custom URLs (`product.url_path` + `product.url_key` and `category.url_path` + `category.url_key`). If some random URL will be called (that's been not yet mapped) then the `url/mappingFallback` action will be called.

This action will be a subject to override by user - but be default it will be calling first `product/single` and then `category/single` to check if specified URL is a product or category URL.
Once mapped URL will be stored in localStorage + `url.state` (because we'll be getting the prefetched mappings from SSR renderer within this initial state)

screenshot 2019-02-08 23 58 23

(run `yarn test:e2e` and paste the results. If you are not using our standard backend setup or demo.vuestorefront.io you can ommit this step) 

- [ ] I read the [TypeScript Action Plan](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/TypeScript%20Action%20Plan.md) and adjusted my PR according to it
- [ ] I read about [Vue Storefront Modules](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md) and I am aware that every new feature should be a module


pkarw:
The second phase can - adding the `/api/url/map` - that will be doing the Category/Product/CMS distinction Server Side in the `vue-storefront-api`. This will save some % of users the second call (if URL is not product then second call for category will be made in `mappingFallback` - in the MVP)

pkarw:
Reference materials:
- https://github.com/vuejs/vue-router/issues/988
- https://github.com/vuejs/vue-router/issues/703
- https://github.com/vuejs/vue-router/issues/1858
- https://github.com/vuejs/vue-router/issues/1283

pkarw:
Notes:
- `rootStore.state.config.products.useShortCatalogUrls` [should be set to true](https://github.com/pkarw/vue-storefront/blob/4fb53e5009e8b83ef2901502fe150e403b63d920/core/helpers/index.ts#L57)
- `product.slug` and `category.slug` are used for generating the links so we could just change the way these links are created to get the proper, absolute link (`product.slug` should be like `product.url_path` + `product.url_key` for example - and the same for categories)

filrak:
Imho we should just use vue router aliases  https://router.vuejs.org/guide/essentials/redirect-and-alias.html

filrak:
Where the mappings are stored initially? On the server side? If we have list of them upfront why we can't just dynamically add routes/aliases instead of doing this complicated chain of store calls.

Also imho if we can't do it in pure vue-router way all of this should be done in a single component. Right nwo the logic is  so complicated that I even can't properly understand how it works.

pkarw:
Filip I was considering these options:
- we cant list all the URL’s to create aliases as for shops with like 100k ski (typical fashion) it will be roughly in megabytes
- we must use ‘*’ to have multi part urls: /:uniquekey won’t handle “/category1/category2”
- we cant use “next()” - it’s redirect and there is no way to keep the original url - this was my initial idea and I spent 2.5h testing it out yesterday without any success

pkarw:
So my approach is mixed of what you’ve proposed (url/mappingFallback and url/mapUrl is this fallback to determine which route is it) adapted to make no redirects. Users can override these actions with their modules to have different mapping logic

I plan to make it simpler and more elegant and even more flexible with async components: https://vuejs.org/v2/guide/components-dynamic-async.html

So it probably won’t by a single file component (I mean url dispatcher) but just JS function + Product, Category, ... will be loaded using import().

With this urldispatcher approach - which Ian pretty standard for any CMs on the market (including Magento, Drupal and others) - having a Front Controller / Dispatcher design pattern implemented - we will have the flexibility of adding the redirects support when products are updated etc

We could then also include a support for the Magento url resolver: https://github.com/DivanteLtd/vue-storefront/issues/2010#issuecomment-441302434

Basically - my approach isn’t any fresh idea, it’s just kind of implementation for Magentos resolver in Vue.js :)

pkarw:
Vue.js should have not only “next()” but also “router.forward()” forwarding the request flow to different route without switching the url. This is reason we can’t use this elegant solution that I’ve spent 2.5h on yesterday (I’ve implemented it as a route guard and it was almost perfect :))

filrak:
@pkarw i was thinking more about having this mapping unique propertyin product object (so it can be searched through elasticsearch)not as one downlaodable api endpoint. 

regarding pattern matching `*` will also work

This is not a redirect if we keep it in the same route. Your problem is with browser history so it's worth investigating if the redirection record can be ommited since there is no actual redirect happening at all - only a javascript code adding route to browser history.

I'm not tied to my solution if you tested it out and it didn't work.

I just think implementing logic from magento backend and forcing frontend devs to use different patterns they used to  is opposite of what we are trying to achieve with Vue Storefront and it's backend decoupling approach. It's extremely hard to understand it even from me since it's a completely different way of thinking that is common in frontend world (and it's a symptom of bigger problem bc we have many things like this in VS that are not intuitive to learn for frontend devs).

So assuming I understood you well mix of this two solutions would be just one component (instead of whole module) similar to what <router-view> is. Basically our router on top of the router.

````html
<template>
  <component :is="pageType"
</template>

<script>
export default {
  data () {
    return {
      pageType: null
    }
  },
  props: {
    uniqueId: { required: true } // unique identifier used to specify which page we should render
  },
  methods: {
    setPage () {
      const pageData = getPageData(identifier) // gets route params for a given page require to render it (like product SKU etc) and type of page to render
      setPageData(pageData) // sets Vuex for given product/category/whatever was matched by getPageData()
      this.pageType = pageData.type
   }
}
````

**Comment by pkarw** [comment]
Ok! Let me work on this for a little bit longer so I would be able to show more clearly how I see it working. Yesterday I was more focused on checking the options than on final solution :) thanks for the feedback it’s very helpful 👍

**Comment by pkarw** [comment]
TODO: 
- caching support for the mappings
- tests
- error handling
- refactoring (I'll try to merge UrlClientDispatcher and UrlServerDispatcher into one component)

File: core/modules/url/store/actions.ts [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+  registerMapping ({ commit }, { url, routeData }) {
+    return new Promise ((resolve, reject) => {
+      commit(types.REGISTER_MAPPING, { url, routeData })
+      cacheStorage.setItem(url, routeData).then(result => {
+        resolve(routeData)
+      }).catch(() => reject())
+    })
+  },
+  mapUrl ({ state, dispatch }, { url, query }) {
+    const parsedQuery = typeof query === 'string' ? parseURLQuery(query) : query
```
pkarw: I'll change this one to https://www.npmjs.com/package/query-string

filrak:
@pkarw can you describe more or less where this URL comes from and how on abstractive level this mechanism works? I have troubles understanding it

core/modules/url/hooks/afterRegistration.ts
```diff
@@ -0,0 +1,5 @@
+import { Logger } from '@vue-storefront/core/lib/logger'
```
filrak: file is empty

core/modules/url/hooks/beforeRegistration.ts
```diff
@@ -0,0 +1,5 @@
+import { AsyncDataLoader } from '@vue-storefront/core/lib/async-data-loader'
```
filrak: file is empty

core/modules/url/index.ts
```diff
@@ -0,0 +1,79 @@
+import { module } from './store'
```
filrak: helpers should be placed in proper folder, not in a root.

core/modules/url/pages/UrlClientDispatcher.vue
```diff
@@ -0,0 +1,76 @@
...
+    if (routeData) {
+      if (store.state.config.seo.useUrlDispatcher && routeData) {
+        const userRoute = userRoutes.find(r => r.name === routeData['name'])
+        if (userRoute) {
+          userRoute.component().then(rootComponent => {
+            _matchingComponentInstance = rootComponent.default
+
+            if (from.name) { // don't execute the action on the Client's side of the SSR actions
+              if (_matchingComponentInstance.asyncData) {
+                AsyncDataLoader.push({
```
filrak: AsyncDataLoader will be removed in next version, instead update to vue 2.6 and use native nehavior

core/modules/url/pages/UrlClientDispatcher.vue
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      next()
+    }
+  })
+}
+
+export default {
+  computed: {
+    page: () => {
+      return _matchingComponentInstance
```
filrak: this shouldn't be a computer property + you shouldn't use arrow functions in computed properties

core/modules/url/types/UrlState.ts
```diff
@@ -0,0 +1,5 @@
+// This object should represent structure of your modules Vuex state
+// It's a good practice is to name this interface accordingly to the KET (for example mailchimpState)
+export interface UrlState {
+  dispatcherMap: {}
```
filrak: missing typings

pkarw:
Let’s discuss this on Monday - core planing at 10:00.
Please bring some constructive counter proposals. This structure is not proposed by coincidence. It’s extremely hard to develop plugin solution for touting the irls both csr and ssr

filrak:
Ok, but please just describe how it works. We can't understand it with @patzick

pkarw:
UrlDispatcher is checking the custom url with mappingFallback - method that can be customized/overridden. It’s returning the route data where this irks is pointing to. So this mechanism was created as a separate layer over existing Product and Categor /CMS pages. More over it’s supporting all content types not only lroducts and categories as url dispatchers operates on the touring tables defined in the theme. I had huge problems with dynamically loading and hydrating the components and with the changes of url within the same routes - for example clickingvrelated products on product page don’t execute full router flow and not resolving lazy loaded components (kind of cache).

pkarw:
Overall I’ve spent 24 hours in total just on experiments.:;

filrak:
where this custom url comes from?

pkarw:
Alternative solution is to totally refactor / rewrite url dispatching  - for example not rely on sky but only on url_path in product.js - marching “*” with both product and category pages and then executing kind of “next” if this url doesn’t match category but product etc. but this will be really hard to understand ...

Now it’s pretty simple: this is another layer - separate module - which is optional. Can be plug in if you like to have these routable urls.

core/build/webpack.base.config.ts
```diff
@@ -78,6 +78,7 @@ export default {
       inject: isProd == false
     })
   ],
+  devtool: 'source-map',
```
filrak: it's longest sourcemap alghorithm that we prev changed to increase build time speed

core/modules/url/store/actions.ts
```diff
@@ -0,0 +1,77 @@
...
+          dispatch('mappingFallback', { url, params: parsedQuery }).then(resolve).catch(reject)
+        }
+      }).catch(reject)
+    })
+  },
+  /**
+   * Router mapping fallback - get the proper URL from API
+   * This method could be overriden in custom module to provide custom URL mapping logic
+   */
+  mappingFallback ({ commit, dispatch }, { url, params }) {
```
pkarw: Here is the fallback if route hasn’t been cached we’re mapping url to route using server call - by default request for category/product are separated but it can be optimized with just one call to /api/mapUrl - or something like this. Here is the source for where the urls come from @filrak 

Please analyze this solution carefully and propose other approach as really: there are not many different options for having this feature. At last from my r&d. I’m totally frustrated with this feature to be honest. Would love if somebody else will do this better way be cause I’m also not very happy with the overall complication and of the duplicates logic of client-entry.ts vs UrlClientDispatcher.ts.

I’ve had some partial successes with other approaches but always there was some deal breaker. The biggest problem was with not executing the full Vue-router flow in case when route has not changed. Imagine the situation when user request is handled by the by “urldisparcher” and user then clicks another time product link. Vue is just executing the watch on $route inside current component. In case if you are on product component clicking on category link you have no way to lazy load category component and replace the current one ....

core/build/webpack.base.config.ts
pkarw: Change it please to other one. Currently it’s disabled at all which makes the debugging almost impossible. It was a big oversaw issue with the latest release :(

pkarw: It’s disabled in prod with my change. You haven’t checked the *.prod.ts config where it’s disabled.

pkarw:
I’m open to other solutions / PoC @filrak and @patzick. Acceptance criteria were like this:
- there is no distinction in the url to which content it points out: /something/category could point to other content than /somethingelse/product- so the urls are totally custom
- we don’t want to do redirects: page is displayed under custom url with no additional url change after content type is discovered
- ssr must work
- offline mode must work
- routes must be cached for not executing the additional network request if we could have obtained the url mapping from product/list, category/list before hand and cache it
- the routing mechanism should be opened for other pages / content types added by the user (CMS pages under the same url structure / whatever)

core/build/webpack.base.config.ts
filrak: Ok, sorry i overseen this

filrak:
Thanks we will try to figure it out. Basically we will just try to take your solution as a base and see how we can simplify it in terms of code. If it can't be then we just need a different solution. I believe that if achieving something requires too much conplication then the approach is wrong and it should be rethinked from a different perspective

pkarw:
To not make the core more complicated we can probably publish this module as a separate / non-core one. Maybe someone will bring better solution later :)

pkarw:
@filrak sure, that makes sense. I haven’t found any other dynamic routing solution for Vue. It could be cool hook for contribution if done properly. 

The simplest way it should work would be if Vue-router supported plugins / or dynamic routes so we could inject them dynamically.

In that case we could have just insert static mapped url { path:  “/something/product”, params: { sku: mappedSku..... 

This approach would be the cleanest one. It must be dynamic and we can’t geberate static routing for all urls as there could be lol 100k+ entries on mid size fashion store.

Maybe there is way to insert routings dynamically so we don’t have to deal with all the other hacks like UrlClientDispatcher things.

Haven’t tested dynamic “addRoutes” executed in the guard.

Maybe we should inject addditiknal static route (from url/mapUrl) in the guard and call just next????

core/modules/url/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      })
+      return routeData
+    } else {
+      return null
+    }
+  })
+}
+
+export const UrlDispatcherGuard = (to, from, next) => { 
+  if (store.state.config.seo.useUrlDispatcher) {
```
pkarw: Maybe we should just leave this guard for “*” and dynamically call “router..addRoutes([routeData]); next();?”  to create dynamic route, with “to.fullPath” set as “path” and leave the all dirty job to Vue Router?

Can You test this way @filrak and @patzick as a PoC?

filrak:
Dynamic routes generation seems fine. I'll ask Eduardo who is a current maintainer of Vue router about this approach

pkarw:
If this dynamic routes will work then we’re saved 😎it will be ultimate solution: simple, elegant and compatible with all the acceptance criteria.

Thanks for this feedback anyway  👍

pkarw:
https://github.com/vuejs/vue-router/issues/888

core/modules/url/index.ts
filrak: I can take care of this

filrak:
Awesome. So i think we have some common ground now! ❤️

filrak:
Btw we should also think how to get rid of this config 'ifs' all around the project. Its producing module-specific side effects and makes our codebase bigger even when someone is not using particular features. I guess we need some kind of composition here instead of conditions

pkarw:
Closing in favour of #2446

PR #2432, opened 2019-02-13
Feature/correct color configurable

DaanKouters:
### Related issues
https://github.com/DivanteLtd/vue-storefront/issues/2419

Sometimes the wrong image was shown on a configured productpage, after choosing a color from the category product list. This is caused by the fact that vue carousel wasn't initiated at the moment the configuration was being set.
This also was the case on the product gallery carousel in the overlay (if you enlarge a product photo).


core/modules/catalog/components/ProductGallery.ts
```diff
@@ -26,30 +24,12 @@ export const ProductGallery = {
...
-        let pagination = this.$refs.carousel.$children.find(c => c.$el.className === 'VueCarousel-pagination')
-        if (navigation !== undefined) {
-          navigation.$on('navigationclick', this.increaseCarouselTransitionSpeed)
-        }
-        if (pagination !== undefined) {
-          pagination.$on('paginationclick', this.increaseCarouselTransitionSpeed)
-        }
-      }
-    }, 0)
+    this.$forceUpdate()
```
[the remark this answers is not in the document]
DaanKouters: To be honoust, i left it in there from the previous version. This was wrapped inside the setTimeOut, maybe in an attempt to get the carousel instance?
When i remove it, there is no difference

src/themes/default/components/core/ProductGalleryOverlay.vue
```diff
@@ -0,0 +1,76 @@
...
+    }
+  },
+  data () {
+    return {
+      showProductGalleryZoomCarousel: false
+    }
+  },
+  components: {
+    'no-ssr': NoSSR,
+    'ProductGalleryZoomCarousel': () => import(/* webpackChunkName: "vsf-product-gallery-zoom-carousel" */ './ProductGalleryZoomCarousel.vue')
```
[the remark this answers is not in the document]
DaanKouters: @patzick what do you mean exactly? You want me to assign this to a constant and add this to components?

core/modules/catalog/components/ProductGallery.ts
```diff
@@ -26,66 +25,17 @@ export const ProductGallery = {
...
-        this.$forceUpdate()
-      }
-    },
     toggleZoom () {
       this.isZoomOpen = !this.isZoomOpen
     },
-    handleEscKey (event) {
-      if (this.isZoomOpen && event.keyCode === 27) {
+    onEscapePress () {
+      if (this.isZoomOpen) {
```
patzick: as i can see you've removed `isZoomOpen` from data of this component. Could you check if this logic should be here?

src/themes/default/components/core/ProductGallery.vue
```diff
@@ -63,34 +38,44 @@
...
   },
   mixins: [ProductGallery, VueOfflineMixin],
   watch: {
     '$route': 'validateRoute'
   },
   data () {
     return {
-      loaded: true,
-      carouselTransitionSpeed: 0
+      isZoomOpen: false,
```
patzick: oh here we have isZoomOpen, so i think handle escape event should be here also

src/themes/default/components/core/ProductGallery.vue
```diff
@@ -63,34 +38,44 @@
...
+    },
     validateRoute () {
       this.$forceUpdate()
     }
   }
 }
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
```
patzick: why you had to remove scoped style?

core/modules/catalog/components/ProductGallery.ts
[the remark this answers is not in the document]
patzick: i think opposite - if there is no difference we should remove that. Because if we need that there should be solid explanation why :)

src/themes/default/components/core/ProductGalleryOverlay.vue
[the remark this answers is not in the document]
patzick: exactly :)

src/themes/default/components/core/ProductGallery.vue
DaanKouters: yeah indeed, i moved it to this component.
But this makes me wonder about the right to exist of ProductGallery.ts. Because so much of the logic is now in the theme file, why still use the minimal ProductGallery.ts in core?

core/modules/catalog/components/ProductGallery.ts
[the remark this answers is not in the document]
DaanKouters: @patzick yeah i agree, i meant that if doesn't make a difference we should remove it. No need to keep it there for no purpose.

src/themes/default/components/core/ProductGallery.vue
DaanKouters: yeah, that was not very good. Because i separated the component in childcomponents quiet a bit, the styling was not correct anymore. I fixed this in a new commit on this branche.

DaanKouters:
> We're getting really close :)

Thanks, would be really cool to see this make it into the core!
I'm looking forward to your view on the matter of keeping ProductGallery.ts in the core because it became such a minimal component now. Or keep it in to function as a skeleton? Almost all the logic is now in theme file.

And also, i'm curious about the notation of dynamic loaded components. You asked me to assign these to a constant, is this for conventional purposes or is there another reason?

Thanks again for your support!

src/themes/default/components/core/ProductGallery.vue
patzick: generally it's a theme case, so yeah, maybe in future we'll get rid of it; but not now yet, it still could contain some utilities needed in every theme

patzick:
We need to think about product gallery. This is theme-specific component so maybe in future we'll do something with that, but for now it's okay.

About const with dynamic imports. It's generally a good practice to show all the imports on top of JS. If it would be an import, which is conditionally loaded on some specific method, then okay - it can stay there. But if it's either way loaded on start - then nice to see it with imports. There is no need to scan whole file to see a basic dependencies.

ibugaj:
Now it works good for me

Issue #2388, opened 2019-02-07
Add Support for Product Videos


rain2o:
## What is the motivation for adding / enhancing this feature?
Magento 2 supports the ability to add videos in the product gallery. While editing the product you can add a video using a URL from YouTube or Vimeo. Vimeo is supported by default, but you need API key for YouTube to work. Magento uses the API to pull in the title, description, and a placeholder image for the video. 
On the front-end of Magento's Luma theme it simply displays the video placeholder image as a product image in the gallery with a "play" icon over it. Once clicked the embedded video is displayed and played. 

There are additional configurations available such as auto-starting the video, auto re-starting the video, and showing related videos. 

## What are the acceptance criteria 
- Videos are available in the product gallery to be played using the video_url provided
- Placeholder images are used in place of videos until user interacts to initiate playback. 

# Can you complete this feature request by yourself?
Yes / Possibly. 

## Additional information
If anyone else takes this on before I get the chance, here is a sample JSON that is returned from the media API which contains an image, a YouTube video, and a Vimeo video. 

    [
        {
            "id": 3423,
            "media_type": "image",
            "label": null,
            "position": 1,
            "disabled": false,
            "types": [
                "image",
                "small_image",
                "thumbnail",
                "swatch_image"
            ],
            "file": "/s/h/shop_home_we_1.jpg"
        },
        {
            "id": 3431,
            "media_type": "external-video",
            "label": null,
            "position": 2,
            "disabled": false,
            "types": [],
            "file": "/h/q/hqdefault_1.jpg",
            "extension_attributes": {
                "video_content": {
                    "media_type": "external-video",
                    "video_provider": "",
                    "video_url": "https://youtu.be/M7R0bQHilgw",
                    "video_title": "Storefront Cloud - Technical Overview",
[...]

PR #2433, opened 2019-02-13
Added support for product videos

rain2o:
### Related issues

#2388 and [mage2vuestorefront#71](https://github.com/DivanteLtd/mage2vuestorefront/issues/71)

This PR adds the ability to load and play videos in the product gallery from YouTube and Vimeo. 

*Note* - As a part of this I created a new component in the default theme called `LoaderScoped`. It can be used to provide the same style loading indicator inside of a scoped area without triggering the full page loader. I thought this could be useful in other areas potentially and needed to indicate the video was being loaded without taking the whole page out. 

**BEFORE: Video Placeholder Image**
video-placeholder-before

**AFTER: Video Placeholder Image**
video-placeholder-after

**Video Loading**
video-loading

**Video Playing**
video-playing


rain2o:
@patzick Sorry for the delay on this! Had some network issues on the ride home from the hackathon. But I finally resolved the URL issue we had so I believe this is ready now.

core/modules/catalog/helpers/index.ts
```diff
@@ -529,18 +530,66 @@ export function configureProductAsync (context, { product, configuration, select
...
+* @param videoData Object
+*/
+
+function computeVideoData (videoData) {
+  let videoId = '';
+  let type = '';
+  let videoUrl = new Url(videoData.url);
+
+  if (videoUrl.host.match(/youtube\.com/) && videoUrl.search) {
+    videoId = videoUrl.search.split('v=')[1];
```
patzick: parsing in this method looks a bit complicated. Maybe we'd use for example parseUrl from https://www.npmjs.com/package/query-string for that?
And then we could have this method during a data import to elastic?

src/themes/default/components/core/ProductGallery.vue
```diff
@@ -47,6 +49,32 @@
                     data-testid="productGalleryImage"
                     itemprop="image"
                   >
+                  <div
```
patzick: all this video part should be a separate component which is displayed on v-if when image is video, to avoid to many logic in single component and make player reusable :)

patzick:
oh, bad thing i see now that we'll have conflict with #2432 because it changes gallery component. That's why video player have to be a separate component, then this wouldn't collide that much. Please keep in mind that PR too to avoid additional work on this feature :)

core/modules/catalog/components/ProductGallery.ts
```diff
@@ -88,6 +88,9 @@ export const ProductGallery = {
       if (this.isZoomOpen && event.keyCode === 27) {
         this.toggleZoom()
       }
+    },
+    initVideo (video) {
```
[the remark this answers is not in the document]
rain2o: Ah darn. I thought I removed this before pushing. I'll get rid of it.

rain2o:
Thanks for the suggestions @patzick ! I'll get this cleaned up and keep the other PR in mind while working on it.

core/modules/catalog/helpers/index.ts
rain2o: Regarding your suggestion to move this to the data import, I thought about that as well and it probably is the better solution. My only thought was if other platforms are integrated with the ability to import videos like this the url parsing logic would need to be recreated in each platform integration (e.g. mage2vuestorefront). I can move it to there unless you have a better suggestion as to where it could sit.

rain2o:
@patzick I have separated the video logs into a separate component, but after looking at that other PR it looks like he created a new component where the carousel is created. In order to make mine work with his I'll need to merge his updates. I guess the best approach is wait for that PR to be merged into develop then I can update my branch with develop.

core/modules/catalog/helpers/index.ts
rain2o: The challenge with using `query-string` or `url-parser`, or any other module like this is that YouTube URLs have many different formats. When running some tests against different formats sometimes the `url-parser` I was using would get what I needed, and sometimes `query-string` would get it. But none of them work all the time. 
So, I cleaned it up and decided to use some Regex for both YouTube and Vimeo to match the domains and try to locate the video ID.

patzick:
hey @rain2o :) 
#2432 has been merged to develop, could you adjust your changes to current state?

rain2o:
Thanks for the update @patzick . I have merged with develop and pushed after cleaning up. Let me know any further thoughts on the PR

pkarw:
@patzick how does the status look like in here?

rain2o:
@DaanKouters just pointed out that I forgot to add the video component to the zoom carousel. I'll get that added in before we move forward with this PR.

core/modules/catalog/components/ProductVideo.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+    id: {
+      type: String,
+      required: true
+    },
+    type: {
+      type: String,
+      required: true
+    },
+    index: {
```
filrak: What is the purpose of this  property? 
If it is emitted later in the code would be good to have a default value instead of undefined

src/themes/default/components/core/ProductGalleryZoomCarousel.vue
```diff
@@ -63,12 +72,15 @@ export default {
   },
   data () {
     return {
-      carouselTransitionSpeed: 300
+      carouselTransitionSpeed: 300,
+      currentPage: 0,
+      hideImage: null
```
filrak: `hideImage` is not very descriptive for what this value actually holds since it implies a boolean. Maybe we can come up with something better and more descriptive / make a computed property for checks like in line 31?

core/modules/catalog/components/ProductVideo.ts
rain2o: It started out as optional because I wasn't sure if the ProductVideo component would always have an index. But you're right, I'm using it later without checking if its set. I gave it a default value of `0` now.

src/themes/default/components/core/ProductGalleryZoomCarousel.vue
rain2o: Good point @filrak . I believe at the beginning this was a boolean but after some issues with that approach it became an integer and I just never renamed it. I updated now to `hideImageAtIndex`, does that work for you?

filrak:
@rain2o please resolve changelog conflict and we can merge in!

rain2o:
@filrak Conflicts have been resolved. Thanks!

PR #2448, opened 2019-02-18
Bugfix/#2437 user can add to cart 0 qty product

RakowskiPrzemyslaw:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/2437

When user enters 0 or value below 0 as quantity validation will set it to 1 as default value.


filrak:
Ok we discussed changes f2f please adjust

RakowskiPrzemyslaw:
use vuelidate, don't change value of the input, disable button, error label. New issue about validation in store

src/themes/default/components/core/AddToCart.vue
```diff
@@ -19,7 +19,7 @@ export default {
       this.$forceUpdate()
     },
     canBeAdded (product) {
-      return formatProductMessages(product.errors) !== ''
+      return formatProductMessages(product.errors) !== '' || product.qty <= 0
```
filrak:
```suggestion
      return formatProductMessages(product.errors) !== ''
```
This check is redundant since we have vuelidate

src/themes/default/components/core/blocks/Form/BaseInput.vue
```diff
@@ -71,6 +73,16 @@ export default {
       type: [String, Number],
       default: ''
     },
+    min: {
```
filrak: i guess it's also redundant because of the same reason. We don't need any internal validation for inpuits since they are validated outside with vuelidate

src/themes/default/pages/Product.vue
```diff
@@ -141,19 +141,26 @@
...
-                <input
+                <base-input
                   type="number"
-                  min="0"
-                  class="m0 no-outline qty-input py10 brdr-cl-primary bg-cl-transparent h4"
+                  :min="1"
+                  class="m0 no-outline qty-input"
                   id="quantity"
                   focus
+                  @input="validateQty"
```
filrak:
```suggestion
                  @input="this.$v.$touch()"
```
Let's follow the convention we have all over the project

src/themes/default/pages/Product.vue
```diff
@@ -292,6 +302,16 @@ export default {
         message: this.$t('No such configuration for the product. Please do choose another combination of attributes.'),
         action1: { label: this.$t('OK') }
       })
+    },
+    validateQty () {
```
filrak: This one is redundant

src/themes/default/components/core/blocks/Form/BaseInput.vue
RakowskiPrzemyslaw: if i remove this user will be able to add below 0 with buttons, this prevents that

src/themes/default/components/core/AddToCart.vue
RakowskiPrzemyslaw: this is to block the button

src/themes/default/components/core/AddToCart.vue
```diff
@@ -19,7 +19,7 @@ export default {
       this.$forceUpdate()
     },
     canBeAdded (product) {
-      return formatProductMessages(product.errors) !== ''
+      return formatProductMessages(product.errors) !== '' || this.disabled
```
patzick: I believe this should be an opposite - can be added if this is *not* disabled
```suggestion
      return !disabled && !formatProductMessages(product.errors)
```

core/modules/cart/components/AddToCart.ts
```diff
@@ -6,6 +6,10 @@ export const AddToCart = {
     product: {
       required: true,
       type: Object
+    },
+    disabled: {
+      required: false,
+      type: Boolean
```
patzick: should have default `false` value

filrak:
@patzick i guess we shouldn't change input value at any time. If someone will write wrong number and click 'add to cart' he/shi will add 1 item without even noticing it. I think there should be only a validation error

PR #2456, opened 2019-02-19
Bugfix/2449 menu not working on static pages

filrak:
### Related issues

closes https://github.com/DivanteLtd/vue-storefront/issues/2449
closes https://github.com/DivanteLtd/vue-storefront/issues/2441

We were fetching menu data only on Home so it wasn't working on any other page if accessed directly from the URL. I updated Vue to v2.6  and used native `serverPrefetch` to fetch sidebar menu data into SSR context right in the root component.


src/themes/default/App.vue
```diff
@@ -26,6 +27,14 @@ export default {
       return `${(this.$route.meta.layout || 'default')}-layout`
     }
   },
+  methods: {
+    fetchMenuData () {
+      return this.$store.dispatch('category/list', { includeFields: config.entities.optimize ? config.entities.category.includeFields : null })
```
pkarw: If we’re moving this to layout we should check all asyncDatas as I’m almost sure we’re also fetching categories in the Product.js and Category.js. Next thing is the sequence to make sure that App.vue serverPrefetch is being caked before Product and Category - otherwise it could geberate some side effects / bugs

src/themes/default/App.vue
```diff
@@ -7,6 +7,7 @@
 </template>
 
 <script>
```
pkarw: We should use the config from rootStore.state.config

pkarw:
I like this solution. We’re aiming at lean/dummy components so adding data prefetch shouldn’t be a general rule of thumb but I believe that layout is good place

src/themes/default/App.vue
pkarw: Moreover it should be fetched in Default and maybe Simple layout burn nit in Minkmal layout which is used on error page - which is displayed most often in case that there is no elasircsearch connectivity. We can’t  have this in App.vue but rather in specific layout which is dynamically loaded by App.vue

filrak: Regarding other pages - you're right I'll check it.
Regarding the sequence - what kind of side effects or bugs you are talking about? is this list from menu used anywhere else? Every page is working correctly

pkarw: I think that layout based prefetch will be always called first so it’s probably always safe. I think it will be safe.

filrak:
@pkarw @patzick it's ready for CR

src/themes/default/layouts/Default.vue
```diff
@@ -60,16 +60,21 @@ export default {
...
-
     })
   },
   methods: {
     onOrderConfirmation (payload) {
       this.loadOrderConfirmation = true
       this.ordersData = payload
       EventBus.$emit('modal-show', 'modal-order-confirmation')
+    },
+    fetchMenuData () {
```
pkarw: Shouldn’t we move: cart/load, wishlist/load and compare/load here as well to make the Woshlist.ts, Microcart.ts more “dummy”? It’s the same case like with categories

filrak: Good idea. Can you make an issue for this? It shouldn't be part of this PR.

pkarw:
Done! Thanks @filrak

PR #2465, opened 2019-02-20
Feature/improve load images on category page

patzick:
### Related issues

(is there any?)

I've removed lazy loading from product tile, because it caused problems with blinking images.
Also alternative view was changed, because if picture was not loaded then it looked really bad.

#### Before:
before
(and with alt)
image

#### After:
after_1
(and with alt)
image


pkarw:
Shouldn't we have the placeholders displayed instead of blank rectangles while images are being loaded? @filrak @patzick

filrak:
we should

filrak:
@patzick do you know why placeholder is not displayed?

pkarw:
Guys, by this chance please test and work on the "blinking" effect which is pretty common to product and cateogry pages (images) in VS - i really hate it :) It's probably because of wrongly applied fadeIn animations

patzick:
Guys, generally i agree with you, but now with low internet connection it looks like this:
jumping

everything is jumping, images are loaded and then v-lazy is starting by replacing image with placeholder image and then flashing back to image itself.

So rectangle block jumping content (which IMHO is really important). I've tested it with trying to set background-image with placeholder under image until it's loaded but sometimes this placeholder is loaded also after images so i think there is no gain in that.

filrak:
Momie i think it's cool like this. You got my props

patzick:
But wait before merging it, i'm checking option with placeholder

patzick:
@pkarw @filrak 
now we have this effect:

aft2

pkarw:
Let's have it set by config. Because if You've got precached images / service worker in place - generally havnig no placeholder will make much better efect (You'll get the images immediately after page refresh). Otherwise the placeholder is probably better option

OK?

pkarw:
Second best option is to have placeholder as a background not an image. So we're not replacing the "src" but rather - displaying the images with "fastest" strategy; placeholder  OR image - depends on the image is prefetched or not

pkarw:
OK, we've discussed it and config makes no sense

PR #2481, opened 2019-02-21
Feature/product gallery image blinking

patzick:
### Related issues

-

Corrects product gallery thumbnails, reduces first gallery slide when specific product color is chosen and turns off default lazy animation which causes blinking on gallery.

Gifs below are converted from mov files and for some reason are slowed down. Animations are faster, but generally shows the difference.

#### Before
before_1

#### After
after_1


pkarw:
That’s fine but shouldn’t we have a placeholder while image is loading?

ibugaj:
Verified with #2482

PR #2485, opened 2019-02-22
Bugfix/#2437 user can add 0 products

RakowskiPrzemyslaw:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/2437

now when user wants to add 0 or less products to cart, the button will be disabled and user will see validation error below input. Also the price will show the price of an single item not below 0 prices. 


core/modules/cart/components/AddToCart.ts
```diff
@@ -6,6 +6,10 @@ export const AddToCart = {
     product: {
       required: true,
       type: Object
+    },
+    disabled: {
+      required: false,
+      type: Boolean
```
patzick: if it's not required then we don't have to do `required: false`, but good practice is to provide a default value for that, so:
```js
disabled: {
  type: Boolean,
  default: false
}
```

src/themes/default/components/core/AddToCart.vue
```diff
@@ -18,8 +18,8 @@ export default {
     onAfterRemovedVariant () {
       this.$forceUpdate()
     },
-    canBeAdded (product) {
-      return formatProductMessages(product.errors) !== ''
+    cantBeAdded (product) {
+      return formatProductMessages(product.errors) !== '' || this.disabled
```
pkarw: This function should be named `isProductDisabled` rather than `canBeAdded` which is miss-leading. I know that it wasn't a part of Your PR but please do fix it by this chance :)

src/themes/default/components/core/AddToCart.vue
```diff
@@ -18,8 +18,8 @@ export default {
     onAfterRemovedVariant () {
       this.$forceUpdate()
     },
-    canBeAdded (product) {
-      return formatProductMessages(product.errors) !== ''
+    cantBeAdded (product) {
```
patzick: this shouldn't be a method, but computed property `isDisabled` which returns Boolean value.
Also i think this.disabled should be first to check, then we don't have to invoke `formatProductMessages`

src/themes/default/components/core/AddToCart.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <button-full @click.native="addToCart(product)" :disabled="canBeAdded(product)" data-testid="addToCart">
+  <button-full @click.native="addToCart(product)" :disabled="cantBeAdded(product)" data-testid="addToCart">
```
patzick:
```suggestion
  <button-full @click.native="addToCart(product)" :disabled="isDisabled" data-testid="addToCart">
```
this will be a result of computed property

src/themes/default/components/core/blocks/Form/BaseInputNumber.vue
```diff
@@ -11,6 +11,22 @@
       @input="$emit('input', $event.target.value)"
       @blur="$emit('blur', $event.target.value)"
     >
+    <template v-if="validation">
```
patzick: i see that it's the same code as in BaseInput - so lets make a component from that :)
phrase `<template v-if="validation">` appears in four input component so it's big code redundancy.

And here (and other components) we can just have a slot for this new validation component.

src/themes/default/pages/Product.vue
```diff
@@ -143,11 +143,19 @@
                 :name="$t('Quantity')"
                 v-model="product.qty"
                 :min="1"
+                @input="$v.$touch()"
```
patzick:
```suggestion
                @blur="$v.$touch()"
```
it's a good practice to show validation messages after user stops editing it for the first time. And when you once touched field then validation will be instant.

patzick: of course please ensure, that BaseInputNumber emits blur:)

src/themes/default/pages/Product.vue
```diff
@@ -143,11 +143,19 @@
...
+                :validations="[
+                  {
+                    condition: $v.product.qty.$error && !$v.product.qty.minValue,
+                    text: 'Quantity must be above 0'
```
patzick: every displayed text have to be with i18n, because we'd lost translation feature

ibugaj:
User can fill qty field by number with dash and add it to cart and proceed to checkout
image
(but in Cart is one product with price for one)

Issue #1104, opened 2018-04-30
Support URL filters on the Category page


pkarw:
We should have support for passing the filters in the URL to create the SEO friendly and promo links

I mean for example:
https://demo.vuestorefront.io/c/women-20/color/241/size/124

Right now the filters works fully just on the client side and are not supoprted by the SSR
After changing the filter the attributes should've been added to the window.location

pkarw:
OK; with the new UrlDispatcher we should use queryString instead `?filter=value` ...

patzick:
It's a good practice to add search filters as queryParams, not pathParams.
To complete this issue we should use queryString lib to just change filtersObject and parse it.

patzick:
yup, exactly @pkarw :)

pkarw:
This is related to #2180 so I've assigned it to  You @patzick

pkarw:
Closed by #2724

PR #2502, opened 2019-02-25
Bugfix/2435 inverted filters

### LINKED ISSUE #2435: Inverted filtering results on category pages
## Current behavior
Filter marked as ON has no effect
Filter marked as OFF is applied to displayed results

## Expected behavior
Filters should be marked as ON and work at the same time

1. Go to https://test.storefrontcloud.io/
2. Open any category page
3. Pick a filter (i.e. color) and leave it marked
4. Go to any other page (i.e. Home Page)
5. Go back to the same category page
6. Click the same filter you just used

Filter OFF:
filteroff
Filter ON:
filteron

no

## Environment details

- Browser: Chrome, Firefox, Opera, Edge
- OS: Win10

## Additional information
Same behavior for sure applies to price filter, and proopably to other filters as well

patzick:
### Related issues

closes #2435 

Filters are in store and are ignored on category page until filter changes. Its correct because filters should change queryParams in current site url.
This fix ensures that filters are cleared every time when category page is displayed, so no more filters sideffects. Also improved clearFilters button, because selecting filters has caused jumping.

Correct behaviour on filters is implemented in #2180 and then filters will be stored in url path as they should.

#### Before
filters_before

#### After
filters_after


pkarw:
I see no reason why not to apply these filters as they’ve been stored between requests?
It will be much better UX. If you are on the product page went there from red color blouses by clicking back you should get red blouses list.

And we do have these filters set in state so instead of clearAllFilters we should call on filters changed in this case

Agree that Bartek’s PR solves this once and for all. However we do have a better workaround at hand here so please reconsider

pkarw:
So please create a issue (I can’t at the moment) with this second option I presented and include it to 1.9.x @filrak @patzick

patzick:
@pkarw i've tried earlier your proposal. But filters are build based on own active flag, not on vuex; so they are active after click them. If we load page (example by going back in browser) and invoke search query filters still will be unmarked (just as this issue is all about). This should be done right - when we refactor so based on route, then user can go back, refresh page or even send a link to his friend with filtered products.
I started to filing an issue, but there is one already - #1104, let's move there further discussion.

pkarw:
OK, agree. Go on and merge this one in

PR #2507, opened 2019-02-26
Bugfix/#2409 check for multiple order placing

### LINKED ISSUE #2409: Add check for `checkout-do-placeOrder` event - to make sure it's called only once no matter how many events are emited

It's related to #2407. Currently, there is a high chance that by minor issue module developer can fire the `placeOrder` few times which can cause critical issues.

We should hash the orders (objHash) and make sure one order is transmitted just once to the server in the `placeOrder` method to make it completely safe

patzick:
### Related issues

closes #2409 

There are three things here:
- [x] fixes inputs validation types from #2485
- [x] refactor checkout/placeOrder action to async/await and adds store type to its getters
- [x] fixes problem if any payment module has error and accidentaly emit `checkout-do-placeOrder`. I tested it with multiple emiting and only single order is placed. In case of server error order can be placed once more. Tested also placing order on same item, with same params multiple times in a row - there is no problem with that.


core/modules/order/store/mutations.ts
```diff
@@ -30,6 +30,12 @@ const mutations: MutationTree<OrderState> = {
   },
   [types.ORDER_LAST_ORDER_WITH_CONFIRMATION] (state, payload) {
     state.last_order_confirmation = payload
+  },
+  [types.ORDER_ADD_SESSION_ORDER_HASH] (state, hash: string) {
+    state.session_order_hashes.push(hash)
+  },
+  [types.ORDER_REMOVE_SESSION_ORDER_HASH] (state, hash: string) {
+    state.session_order_hashes = state.session_order_hashes.filter(sessionHash => sessionHash !== hash)
```
pkarw: @patzick shouldn't we store it localStorage (plugin for syncing state with localStorage - sth like a shopping cart)? I I mean - I'm not sure. Most cases are covered just by this approach and its' totally fine with me

patzick: i think we shouldn't - it's a check for multiple order invocation and works realtime.
Aaand about this localStorage....when you place an order without registration and go to homepage, even refresh page, go place another order and...user data filled in checkout. I think user data shouldn't be stored there.

pkarw:
Ok! Go on with merge

PR #2510, opened 2019-02-27
Bugfix/#2450 show prices after adding product to cart in offline mode

### LINKED ISSUE #2450: No price displayed in products added in offline mode
When product is added in the offline mode there is no price displayed in the Microcart/Product.vue
36d2e549-c96c-41af-90be-23d2f9c66c2a

patzick:
### Related issues

closes #2450 

Shows regular price when there is no totals from server.

I also discovered, that cart summary is in `cart.state.platformTotalSegments` which is updated after response from server, so after adding new product summary still was not updated. So i made a reactive online helper and used it in getter which returns cart summary.

New Vue.observable is pretty cool! :boom:  :)

### DIFF
--- a/core/helpers/index.ts
+++ b/core/helpers/index.ts
@@ -2,6 +2,7 @@ import rootStore from '@vue-storefront/core/store'
 import SearchQuery from '@vue-storefront/core/lib/search/searchQuery'
 import { remove as removeAccents } from 'remove-accents'
 import { Logger } from '@vue-storefront/core/lib/logger'
+import Vue from 'vue'
 
 /**
  * Create slugify -> "create-slugify" permalink  of text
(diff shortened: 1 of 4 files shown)


core/helpers/index.ts
```diff
@@ -159,3 +160,10 @@ export function once (key, fn) {
 }
 
 export const isServer: boolean = typeof window === 'undefined'
+
+// Online/Offline helper
+export const onlineHelper = Vue.observable({
```
filrak: should be just isOnline like isServer imho por teh sake of consistance

patzick: yeah, but then it would be `isOnline.isOnline` which would be weird usage

PR #2570, opened 2019-03-07
Remove text/html from gzip_types

ludioao:
Nginx per default compress text/html and I nginx was returning a warning.

Per default, Nginx compress text/html gzip_types

https://stackoverflow.com/questions/6475472/duplicate-mime-type-text-html

- [X] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)
- [X] I read the [TypeScript Action Plan](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/TypeScript%20Action%20Plan.md) and adjusted my PR according to it
- [X] I read about [Vue Storefront Modules](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md) and I am aware that every new feature should be a module

### DIFF
--- a/docs/guide/installation/prod.vuestorefront.io
+++ b/docs/guide/installation/prod.vuestorefront.io
@@ -40,6 +40,7 @@ server {
 		text/javascript
 		application/javascript
 		application/json
+    text/xml
 		text/json;
 
 	location / {
--- a/docs/guide/installation/prod.vuestorefront.io
+++ b/docs/guide/installation/prod.vuestorefront.io
@@ -38,7 +38,6 @@ server {
 	gzip_types
 		text/css
 		text/javascript
-    text/xml
 		application/javascript
 		application/json
     text/xml
--- a/docs/guide/installation/prod.vuestorefront.io
+++ b/docs/guide/installation/prod.vuestorefront.io
@@ -38,6 +38,7 @@ server {
 	gzip_types
 		text/css
 		text/javascript
+    text/xml
 		application/javascript
 		application/json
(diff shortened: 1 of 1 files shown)


docs/guide/installation/prod.vuestorefront.io
```diff
@@ -38,11 +38,9 @@ server {
 	gzip_types
 		text/css
 		text/javascript
-		text/xml
```
janmyszkier: Why was this type also removed? It's not compressed by default, says nginx doc: https://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_types

ludioao: Ah, my mistake. Only text/html is compressed by default.

filrak:
Please update the changelog 9as it stands with checkboxes you accidentally deleted) and we can merge ;)

janmyszkier:
@filrak updated changelog with all 3 PRs:
- https://github.com/DivanteLtd/vue-storefront/pull/2568
- https://github.com/DivanteLtd/vue-storefront/pull/2571
- https://github.com/DivanteLtd/vue-storefront/pull/2570
because they all needed [unreleased] section which would require to resolve conflicts later for

docs/guide/installation/prod.vuestorefront.io
patzick: so should or shouldn't this be resolved?

PR #2590, opened 2019-03-13
set default storeId to 1 instead of 0 so default magento storeId is q…

janmyszkier:
…ueried instead of 0 which never has chance of existence as primary key

### Related issues
none

cmsModule by default uses endpoint:
https://github.com/DivanteLtd/vue-storefront/blob/master/config/default.json#L405
but the storeId is set to `0`. This PR is made for default magento storeId `1` to be queried instead of `0`.
0 storeId doesn't exist in default magento instances (nor any instance which usually supports primary keys). 

no changes

none, but storeId is nowhere compared to fallback value `0` 


src/modules/magento-2-cms/components/CmsData.vue
```diff
@@ -54,7 +54,7 @@ export default {
       return currentStoreView()
     },
     storeView () {
-      return (this.isMultistoreEnable && this.currentStore) ? this.currentStore.storeId : 0
+      return (this.isMultistoreEnable && this.currentStore) ? this.currentStore.storeId : 1
```
filrak: since storeID is set to '1' by default it can be false therefore the last check is redundant. Am I right?

janmyszkier: @filrak I don't think so.
what the above condition seems to do is:
- if you have multistore and the current store has been correctly set, use `this.currentStore.storeId` (from multistore module)
- else (multistore is disabled or this.currentStore is empty) use cms fallback (not taken from multistore fallback)
there might be some magic logic I cannot find but when we tried, both of those had to be set separately (depending on multistore setup we had). Based on that I think this still require to set BOTH fallbacks this way.

janmyszkier:
@patzick rebased on master and added changelog

PR #2607, opened 2019-03-19
Bugfix/2595 - different caches for storeViews

### LINKED ISSUE #2595: "multisiteCommonCart": true has no effect.
## Current behavior
Setting "multisiteCommonCart": true, has no effect

## Expected behavior
Setting "multisiteCommonCart": true, should force the user to have different carts on the different stores respectively

Setup multistore environment with: 
"multisiteCommonCart": true

Visit each store, observe localStorage in devtools, only "default" non-prefixed cart storage will be visible:
screenshot-1552651974

## Repository
@vue-storefront: v.1.8.3

# Can you handle fixing this bug by yourself?
No, not at the moment

## Additional information
The initializing of $db.cartsCollection is run in the beforeRegistration hook in the cart module. This happens as part of createApp (core/app.ts) and during this the currentStoreView is always initialized with null.
Later during bootstrap, either client-entry og server-entry the storeview is changed, with prepareStoreView.
Only problem is that the $db.cartsCollection isnt updated, since the beforeRegistration hook is not fired again after createApp().

Like with the syncTask storage it should be initialized when the store changes (core/lib/multistore.ts:76). 
There are a couple of modules who would rely on this behaviour: cms, user, cart, compare, checkout.

filrak:
### Related issues

closes #2595
image


README.md
```diff
@@ -268,28 +268,28 @@ Vue Storefront is a Community effort brought to You by our great Core Team and s
         </a>
```
filrak: pulled from master

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
```
filrak: pulled from master

docs/guide/installation/prod.vuestorefront.io
```diff
@@ -38,11 +38,10 @@ server {
 	gzip_types
```
filrak: pulled from master

src/themes/default/css/vendor/_flexboxgrid2.scss
```diff
@@ -83,6 +83,16 @@
   max-width: 100%;
```
filrak: pulled from master

patzick:
@filrak could you resolve conflicts so we could merge that? :)

PR #2617, opened 2019-03-20
Fix error with trailing spaces

Aekal:

Fixed an error during compilation with `yarn dev` on latest develop

### Screenshots of visual changes before

Zrzut ekranu 2019-03-21 o 00 28 23


Aekal:
Not needed anymore, it's already fixed in develop

PR #2618, opened 2019-03-21
core category page - fixed lint issue trailing spaces not allowed

szafran89:
There was lint issue `Trailing spaces not allowed` once I run `yarn dev`


szafran89:
Ok I see there is the same fix provided by @Aekal one PR before :D

Issue #2593, opened 2019-03-14
Payment.vue issue when no address set


pkarw:
## Current behavior

![Screenshot 2019-03-14 09 30 31](https://user-images.githubusercontent.com/211899/54350519-cbf88a00-464d-11e9-8190-0cc53aaddb43.png)

## Expected behavior

No error.

## Steps to reproduce the issue

I've logged in as a user who has no address assigned; then added the product to cart and "Go To Checkout". Then this error appeared.

kziuzial4w:
My issue.

PR #2619, opened 2019-03-21
#2593 - Payment - fixed issue when no address set

szafran89:
### Related issues

#2593 

Fixed console error on checkout page


patzick:
@szafran89 thank you for this PR!
As it is planned as improvements to current RC could you create PR for `release/1.9` branch with this change? This requires to make new branch from there and commit this change.
Also a little bit cleaner would be `if (this.payment.firstName && this.payment.firstName.length) {`

Issue #841, opened 2018-03-10
Adding Google Tag Manager feature


karolbzik:
Let's add Google Tag Manager to the main template.
I prepared account with ID GTM-WGQ7T9C.

Here's official documentation for installation - https://developers.google.com/tag-manager/quickstart.

Can anyone help me with this?

kjugi:
I've dive deep into this issue and searched something like this: https://github.com/mib200/vue-gtm Don't know if this plugin worth to implement but seems to do what we want. The question is which tags and events to trigger.

I've tried to add it on local (pushed one commit with an implement on my fork) but found one problem. In REAMDE author said we must copy/paste GTM tag in our `index.html` and it seems to be ok but how to call the google tag id from config here? Of course, it can be some other idea to deal with it but don't find anything

karolbzik:
@kjugi thank you for the suggestion.

From my point of view the best would be creating an integration with already predefined GTM data layer for ehanced ecommerce for Google Analytics (then we could even skip #839 issue). The shape of the data layer is explained here https://developers.google.com/tag-manager/enhanced-ecommerce.

@pkarw @filrak WDYT about https://github.com/mib200/vue-gtm and what @kjugi and I suggested?

pkarw:
You should create a component for this I belive please take a look at analytics extension as a reference

talalus:
I'll try on this

jorkvist:
Any update on this? Is it paused? I tried the code but it's failing for me. Something with the router that is passed that breaks i think. Would also be nice to get the correct DataLayer implemented on atleast products and category pages.

talalus:
I’ll continue on this in the weekend, yes sure data layer will be implemented

jorkvist:
@talalus thanks for that and thanks for the answer! :) 👍

anders03:
This would be a much-appreciated function for us marketing people. Would be very exciting to see GTM working on Vue-storefront

talalus:
merged with develop

jsxFardin:
Is there any documentation for tag manager ?

talalus:
@jsxFardin 
1. Google Enhanced Ecommerce docs: https://developers.google.com/tag-manager/enhanced-ecommerce
2. Vue-gtm docs https://github.com/mib200/vue-gtm
You can find current tracked events [here](https://github.com/DivanteLtd/vue-storefront/blob/master/src/modules/google-tag-manager/hooks/afterRegistration.ts)

[...]

PR #2624, opened 2019-03-21
Feature/841

talalus:
#841 

Still need to know what is the best place to track
-  product impressions
-  checkout step


talalus:
@patrykpiston changelong updated

Flyingmana:
@talalus for changelog files, please dont replace but only add a new line, then git can do its magic to avoid conflicts

src/modules/google-tag-manager/hooks/afterRegistration.ts
```diff
@@ -0,0 +1,100 @@
+import { currentStoreView } from '@vue-storefront/core/lib/multistore'
+
+export function afterRegistration(Vue, config, store, isServer) {
+  if (config.analytics.id && !isServer) {
```
Flyingmana: this should check for `config.googleTagManager.id`

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.10.0-rc.1] - UNRELEASED
 
 ### Added
--
```
Flyingmana: replacing this line causes a conflict, it should just add a new one

src/modules/google-tag-manager/hooks/afterRegistration.ts
talalus: Good catch, Thanks

CHANGELOG.md
talalus: done

PR #2630, opened 2019-03-21
Event bus import removal

szafran89:
IMO we can remove `EventBus` import and use `$bus` plugin


patzick:
@szafran89 there should be one consistent way of calling EventBus in components, your proposal is okay to me - would you consider changing all uses with this PR? :)

filrak:
Agree, can you check the other imports too?

szafran89:
Yes, I'll check it tomorrow

filrak:
image

szafran89:
Sorry for delay, I'll check it ASAP :D

szafran89:
@patzick please check again, now should be ready to merge ;) There are a few more places with `EventBus` but can't change it b/c it is called before initiating the component.

PR #2656, opened 2019-03-25
Fixed cart sync anomaly

chygoz2:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/2639

Disabled the **Add to Cart** button when clicked until the product has actually been added to the server cart.


chygoz2:
> Thanks! I belive that debounce() should be implemented in the UI (*.vue) components - not in the Vuex action

So i was thinking that if the debounce were to be implemented in the UI, the user experience might not be the best in the sense that when the add to cart button is clicked, the user will not receive any form of feedback until the duration passed to the debounce function has elapsed. This would tend to make the user click the add to cart button more times than actually intended.

pkarw:
You're right in this situation maybe we need to refactor it a little bit deeper - meaning to move the notification outside of Vuex as well; @patzick what's your opinion on that?

patzick:
@pkarw @chygoz2 
I think the flow should goes this way:
1. `cart/addItem` action returns a promise which is resolved after product is added to cart
2. button awaits for this action and till then its state is 'disabled' so there is no possibility to click again

in this case state of "processing" is kept in button. But we can also can have situation, where we have multiple 'add to cart' buttons, so this also should be covered. So in my opinion best way is to have "adding_to_cart" boolean value in cart store state, which is used by getter like `isAddingToCartInProgress` which is passed to addToCart buttons. So:

- when action is called it sets `adding_to_cart` flag to `true`, buttons are disabled then (or even better - they have some processing state on that behavior to show progress); and after adding to cart (or in case of any error) flag is set back to false.

There is no need to use debounce for that and user is perfectly aware of what is going on.

chygoz2:
@patzick This makes perfect sense. I'll make the necessary changes and update the PR. Thanks!

pkarw:
The isAddingToCart flag shouldn’t be global as it can cause fatal errors (no possibility to add any other product if the request wasn’t successfully fulfilled). I belive it would be far better to use the isProductDisabled flag for this purpose as it uses the product.errors collection which is always set for the specific product only

pkarw:
Closed due to long inactivity. Will be fixed in #2887

PR #2746, opened 2019-04-17
Fix range queries to elasticsearch

oskar1233:
Since now, range queries were activated only if `filter.value` contained ALL of `rangeOperators`. This logic is obviously wrong.

This PR inverses the logic - we check if every `filter.value` key is one of `rangeOperators`.


pkarw:
Please just update the Changelog, ok?

oskar1233:
@patzick Thanks for doing that.

Issue #2693, opened 2019-04-08
Switching between login and register form needs 2nd attempt


ArturDivante:
## Current behavior
1. First click on 'or register an account' link is not opening registration form
2. Small validation error 'Field is required' is displayed under email field. 
3. Second click on the link will open correct form

## Expected behavior
1. Registration form should be opened on 1st click

## Steps to reproduce the issue
1. Open https://demo.storefrontcloud.io/
2. Click My Account button to open login  popup
3. Click 'or register an account' link

## Can you handle fixing this bug by yourself?
No

## Which [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html) state this refers to? Info for developer.
Bug is reproducable on every environment, but its easiest on https://demo.storefrontcloud.io/

## Environment details
- Browser: many
- OS: Win10->

## Additional information
1. Reproducing this issue might be tricky. Refreshing page is not always enough. I had highest reproduction rate by closing and lauchning fresh browser instance - but still not 100%
2. Occasionaly, same error appears in diffrent direction - when we use 'or login to your account' link from registration popup. 

![loginRegister](https://user-images.githubusercontent.com/47415585/55733252-ca886a80-5a1d-11e9-9714-ca08b6629a7d.PNG)

rio-karim:
Happy to take a look at this.

SylarRuby:
> Happy to take a look at this.

How you on this? Winning?

SylarRuby:
I'm not an expert here but I've noticed the use of `$touch()` being used on an `input` with `:validation` also with `focus`. To me, all cannot be mixed. I've removed `focus` on both `Register.vue` and `Login.vue`. See my [branch](https://github.com/SylarRuby/vue-storefront/tree/remove-focus-on-model-%232693)

rio-karim:
> I'm not an expert here but I've noticed the use of `$touch()` being used on an `input` with `:validation` also with `focus`. To me, all cannot be mixed. I've removed `focus` on both `Register.vue` and `Login.vue`. See my [branch](https://github.com/SylarRuby/vue-storefront/tree/remove-focus-on-model-%232693)

Trying to observe the change in Auth States but doesn't seem to fire those methods when the issue throws. Does removing focus attr on the Login and Register fix this issue?

SylarRuby:
[...]

PR #2759, opened 2019-04-18
Remove focus on model #2693

SylarRuby:
Emits sends a global event which opens the model. It also listens for a "touch" to close the modal. Both `Register.vue` and `Login.vue`'s form `input`  has an event listener called `$touch()` while using a validation. As it's also being focused, when you touch an element, the validation then triggers. Removing the `focus` prevents the validation from triggering.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [1.8.3] - 2019.04.18
+## [1.8.6] - 2019.04.18
 - Switching between login and register form needs 2nd attempt - @SylarRuby (#2693)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


pkarw:
@patzick please do check this one - the issue with switching from Login to Register (and back) is that even when You don't enter anything in the form and just click "or Register account" button it fires the validation anyway. This PR is preventing this behavior I guess

patzick:
@SylarRuby  I'm not sure if i get this right. We need focus on field for possibility to type email right after modal is open. Blur caused validation check which is also a proper behavior for forms.
I've changed input heigh to avoid form jumping during validation.

Could you provide me some step by step info how to replicate issue with touch event which closes modal?

SylarRuby:
You can't. When the modal is opened, an event is listening to close the model. That event is the culprit. Open your console window, open the model and click any where in the console. The modal disappears. When you use `focus` it assumes that input field is being entered and validation runs on modal close.

> We need focus on field for possibility

I'd say remove event listeners on the modal container. Where is that? I'd have to look and update this PR.

SylarRuby:
Could you tell me what is `$touch()` / how it's set?

pkarw:
Closing due to version incompatibility

PR #2769, opened 2019-04-23
Issue/add custom options to standard system fields

afirlejczyk:
Custom Options are not visible on product page by default.
I add `custom_options` to standardSystemFields and options will be now visible on product page.
screenshot-localhost-3011-2019 04 23-12-18-09

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -177,6 +177,7 @@
           "description",
           "configurable_options",
           "tsk",
+          "custom_options",
           "custom_attributes",
           "size_options",
           "regular_price",


pkarw:
This change is fine however I'm not sure if it could have fixed the custom options issue as `standardSystemFields` property is being used just in one single place in the code to optimize the attribute meta data loader: https://github.com/DivanteLtd/vue-storefront/blob/cde3b9c03eff3deeea7dc0bdd5cc9e311cfafa07/core/modules/catalog/store/product/actions.ts#L618

?

Issue #2573, opened 2019-03-08
Problems with product pictures in offline mode (PLP and PDP)


ArturDivante:
## Current behavior
- After step 4: Product image blinks and disapears, if this product wasnt pre-loaded before going offline
- After step 5: all pictures on PLP are replaced with the picture from last opened PDP.  

## Expected behavior
Pictures are displayed correctly in offline mode

## Steps to reproduce the issue
1. Go to https://test.storefrontcloud.io/
2. Open women category
3. Turn to offline mode
4. Open any product page
5. Go back to category page

_(in attached video, 1st and 4th product were proeloaded - so they load correctly)_
![offline pictures bug](https://user-images.githubusercontent.com/47415585/54026360-d0c8c400-419d-11e9-93fe-618aeb5908fa.gif)

## Repository
https://test.storefrontcloud.io/

## Can you handle fixing this bug by yourself?
no

## Environment details

- Browser: Many
- OS: Win10

pkarw:
@patzick there is still a problem with offline images on next (works just fine on demo) - the low quality offline images are not used which causes two things:
- when loading the image even when user is online there is no this specific proffrsssive loading effect that’s visible on demo (you get pixelated image and immediately afterwards the sharp one)
- images are simply not displayed in the offline mode
![BC9A4C43-961B-4FA4-90F0-B1387536A764](https://user-images.githubusercontent.com/211899/55581126-5fd4e780-571c-11e9-8433-02beead0741e.png)

pkarw:
Another thing - when switching colors it shouldn’t display the placeholder but stock to the thumbnail picture (offline image which is selected in product/single) - compared to demo (1.8.4) where it works exactly like that
![5D1E6007-8486-453C-A400-624DEDEE095B](https://user-images.githubusercontent.com/211899/55581431-1b961700-571d-11e9-97d2-ee3ff240c3ce.png)

pkarw:
Even I’m back online still images won’t work :-(

![C3B4F262-E073-4161-96FC-B60CFFEDE9F2](https://user-images.githubusercontent.com/211899/55581536-6b74de00-571d-11e9-91c8-3ece0f03158f.png)

pkarw:
The same thumbnail should be used in the Microcart - need to be double checked

ArturDivante:
Also, we still have returning issue - some products are not pre-loeaded for offline mode while they should, exactly like in #2543

1. Open any category
2. Turn to offline mode
[...]

PR #2812, opened 2019-04-26
Fix blinking category page

filrak:


patzick:
@pkarw @filrak guys, this breaks offline, so i'm closing for now also #2573, it's strictly related

PR #2860, opened 2019-05-06
FAQ - product page CORS issues

lukeromanowicz:


docs/guide/basics/recipes.md
```diff
@@ -55,6 +55,12 @@ In a case of
 See the discussion in [#137](https://github.com/DivanteLtd/vue-storefront/issues/137).
 Please also check the [Database tool](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/Database%20tool.md)
 
+## HTTP 400 / CORS errors when trying to retrieve detailed product information from API
+
+If all the other requests work without any issues then it's highly possible that your products have too many attributes. All of them are included in ElasticSearch query sent via GET request which might become too long. Try to limit amount of attributes that are indexed and later used for this product view.
+
+This answer is valid until [this issue is resolved](https://github.com/DivanteLtd/vue-storefront/issues/2167).
```
pkarw: The other reason for CORS error is when the `config.elasticsearch.host` has been set without the `http://` either `https://` url schema (eg: `localhost:8080/api/catalog`). In that case the user needs just to update the config setting in order to fix the CORS errror (https://www.dropbox.com/s/1sjmmmuodzkt9n7/Screenshot%202019-05-12%2008.59.55.png?dl=0)

lukeromanowicz: In my opinion, this is a completely different issue. I've pointed out a situation where only 1 specific request throws CORS while in this case, if I understand it correctly, there will be CORS issues all over the catalog part of application.

PR #2889, opened 2019-05-12
Feature/2649

### LINKED ISSUE #2649: Refactor rootStore.state.config to dedicated module

The `rootStore.state.config.entities.attribute.includeFields` kind of paths are all around the code. We should create a dedicated module `ConfigManager` (like [RouteManager](https://github.com/DivanteLtd/vue-storefront/blob/develop/core/lib/router-manager.ts)) to keep the config.

Note: The thing is that the config [needs to be updated runtime](https://github.com/DivanteLtd/vue-storefront/blob/1b51682feb55cc9098d3d36a5a0e47505b827472/core/app.ts#L72)

- the old pattern `rootStore.state.config` should still work for backward compatibility
- all the instances in the core + theme of `rootStore.state.config` replaced by this new fancy module

pkarw:
### Related issues

closes #2649 

I've refactored the way we access the config. The `rootStore.state.config` is still available however - deprecated. I belive we should remove this property starting with Vue Storefront 2.0.

From 1.10.rc-1 the only recommended way to access the config file is:
```js
import { ConfigManager } from '@vue-storefront/core/lib/config-manager'
const config = ConfigManager.getConfig()
```

I've tested the code with the `config.dynamicConfigReload=true` which provides the user with the modified config files within each request.

`ConfigManager.geConfig()` - returns the current, dynamically reloaded configuration
`ConfigManager.getBaseConfig()` - returns the bundled in, static configuration file - it's just required with some of the calls that.. basically set up the dynamic config mechanism :)

The naming is after previously accepter `RouterManager`

**Note to CR** We shouldn't have root module level calls like: 

```
const config = ConfigManager.getConfig()
```

The config should be obtained at the function level to make sure it's the updated reference/object after the runtime changes. Kind of IoC for config. Therefore in many places, I used the `ConfigManager.getConfig().someProperty` accessors that of course violates the Demeter Principle. It could be optimized in the second step by extracting the subsequent getConfig() calls within single JS function to single `const config = ConfigManager.getConfig()`. I haven't had time to optimize and test it properly and because of the V8 optimizations, this optimization is just to .. improve the readability :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed wrong price displayed in instant checkout module - @vishal-7037 (#2884)
 
 ### Changed / Improved
+- Changed the way to access the configuration. Currently the `rootStore.state.config` is deprecated. Please do use the `@vue-storefront/core/lib/config-manager.ts` > `ConfigManager.getConfig()` instead - @pkarw (#2649)
 - Changed the order number (from `entity_id` to `increment_id`) on MyOrders and MyOrder pages - @pkarw (#2743)
 - Disabled the server cart sync in case user is in the checkout - @pkarw (#2749)
 - Improved ProductGalleryCarousel component to handle nonnumeric options id’s - @danieldomurad (#2586)
--- a/config/default.json
+++ b/config/default.json
@@ -10,8 +10,8 @@
       "outputCacheDefaultTtl": 86400,
       "availableCacheTags": ["product", "category", "home", "checkout", "page-not-found", "compare", "my-account", "P", "C", "error"],
       "invalidateCacheKey": "aeSu7aip",
-      "dynamicConfigReload": false,
-      "dynamicConfigContinueOnError": false,
+      "dynamicConfigReload": true,
+      "dynamicConfigContinueOnError": true,
       "dynamicConfigExclude": ["ssr", "storeViews", "entities", "localForage", "shipping", "boost", "query"],
       "dynamicConfigInclude": [],
       "elasticCacheQuota": 3096
--- a/config/default.json
+++ b/config/default.json
@@ -10,8 +10,8 @@
       "outputCacheDefaultTtl": 86400,
       "availableCacheTags": ["product", "category", "home", "checkout", "page-not-found", "compare", "my-account", "P", "C", "error"],
       "invalidateCacheKey": "aeSu7aip",
-      "dynamicConfigReload": true,
-      "dynamicConfigContinueOnError": true,
+      "dynamicConfigReload": false,
+      "dynamicConfigContinueOnError": false,
       "dynamicConfigExclude": ["ssr", "storeViews", "entities", "localForage", "shipping", "boost", "query"],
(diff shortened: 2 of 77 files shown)


pkarw:
I've got no clue why there is some issue with unit tests: https://www.dropbox.com/s/m2mek6e2d3jf7xu/Screenshot%202019-05-12%2011.50.55.png?dl=0 

I mean: it's related to getters, that were fixed in the https://github.com/DivanteLtd/vue-storefront/pull/2828

pkarw:
We probably do need to replace the `this.$config` with a `ConfigManager.getConfig()` as `import { ConfigPlugin } from '@vue-storefront/core/compatibility/plugins/config'` -> it's kind of deprecated/left just due to compatibility issues

patzick:
@pkarw why not just `import {somePropertyWeWanna} from 'config'`?
Is ConfigManager you introduced doing something more with config?

pkarw:
@patzick it allows you to update the config runtime / reload config and not have it bundled at compile time (which is the case with importing config property from static precompiled bundle)

pkarw:
... without this runtime reload option the user must for example rebuild the app with each new node deployed - this is just one case why dynamic config reload is required

filrak:
In addition to what @patzick wrote - I don't see a reason to make it a function. It's executed when code is reloaded exactly the same way object is read. There is no advantage of using a function - only performance and DX disadvantages.

The API should be as simple as this
```
import config from 'config'
```
Where config is defaultly exported object, Every import is using a reference to this object by default (that's how ES modules work)so it's in sync.

filrak:
> ... without this runtime reload option the user must for example rebuild the app with each new node deployed - this is just one case why dynamic config reload is required

Can you give a use case when something like this may be needed on production?

pkarw:
@filrak it was requested both by devops teams and the developers - for example when you’re doing a hot-deploy scaling the up horizontally you need to change the endpoints sometimes. It’s not the often case but having the config static is the worst option.

Maybe an other option is to return the default from config module that lets you update itself - so the object having kind of “update” method + exposing all the other properties 

However I don’t see much a value in it.

You can still use the config as you wrote. The only thing is there is no dynamicUpdate support in this case.

I belive we should go into kind of IoC support - the same way we want with RoutesManagae now we go with ConfigManager etc

pkarw:
We should go with the same approach for the other things we have in Vue.prototype:
https://github.com/DivanteLtd/vue-storefront/search?utf8=%E2%9C%93&q=Vue.prototype&type=

StorageManager - for getting the UniversalStorage instanced instead of $db
EventsManager - for replacing the $bus - however it’s up to discussion

filrak:
> Maybe an other option is to return the default from config module that lets you update itself - so the object having kind of “update” method + exposing all the other properties. However I don’t see much a value in it.
You can still use the config as you wrote. The only thing is there is no dynamicUpdate support in this case.

I disagree - as I mentioned ES modules are singletons so there is no additional code needed (especially not a function) https://k94n.com/es6-modules-single-instance-pattern

alinadivante:
Due to ongoing discussion, I remove label for QA. Please add it again when it will be ready for testing :)

pkarw:
Closing this one in favor od #2899

PR #2710, opened 2019-04-11
#2560 - Redirect to the checkout after adding the items to the cart.

MikeSheward:
### Related issues
#2560

### Short description and why it's useful
After adding items to your cart, redirect to the checkout.

### Which environment this relates to
Check your case. In case of any doubts please read about [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html)

- [x] Test version (https://test.storefrontcloud.io) - this is a new feature or improvement for Vue Storefront. I've created branch from `develop` branch and want to merge it back to `develop`
- [ ] RC version (https://next.storefrontcloud.io) - this is a stabilisation fix for Release Candidate of Vue Storefront. I've created branch from `release` branch and want to merge it back to `release`
- [ ] Stable version (https://demo.storefrontcloud.io) - this is an important fix for current stable version. I've created branch from `hotfix` or `master` branch and want to merge it back to `hotfix`

### Upgrade Notes and Changelog

- [x] No upgrade steps required (100% backward compatibility and no breaking changes)
- [ ] I've updated the [Upgrade notes](https://github.com/DivanteLtd/vue-storefront/blob/develop/doc/Upgrade%20notes.md) and [Changelog](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md) on how to port existing VS sites with this new feature

**IMPORTANT NOTICE** - Remember to update `CHANGELOG.md` with description of your change

### Contribution and currently important rules acceptance

- [x] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)
- [x] I read the [TypeScript Action Plan](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/TypeScript%20Action%20Plan.md) and adjusted my PR according to it
- [x] I read about [Vue Storefront Modules](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md) and I am aware that every new feature should be a module

PR #2937, opened 2019-05-23
#2560   redirect to the checkout after adding the items to the cart

### LINKED ISSUE #2560: Remake order is not remaking the order - misleading behavior
Currently the 'Remake Order' button in profile is only adding the same products to the cart. It's misleading. At bare minimum it should also redirect to checkout.

patzick:
### Related issues

closes #2560

Duplicate of #2710 done by @MikeSheward  which couldn't be merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -174,6 +174,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated cypress dependency for e2e tests - @lukeromanowicz (#2518)
 - Improved styles on recommendation filters, product tile and numeric input - @patzick (#2458)
 - Removed editing mode from My Newsletter section - @aniamusial (#2766)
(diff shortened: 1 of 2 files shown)


alinadivante:
If I use 'Remake order' from dropdown list 
image - it works (I'm redirected to checkout) - **BUT** if I will remove products from cart and I will try again remake the same order then it doesn't work. Im' redirected to Home page and "Shopping cart is empty. Please add some products before entering Checkout" pop-up occurs and at the same time product is added to cart. 
image

What more? If I click on "view order"  from dropdown list and after that I click 'Remake order' link (on order details page) it causes only adding product to cart (without redirecting to checkout) 
image

pkarw:
@alinadivante could you create a new issue for this please? I’m afraid it would get lost just here in the comment :/

alinadivante:
@pkarw here is open issue for that https://github.com/DivanteLtd/vue-storefront/issues/2560
I wrote this comment here by mistake ;)

Issue #2685, opened 2019-04-08
Missing static file causes memory bloat


grimasod:
## Current behavior

A missing static file can cause the server to bloat and crash.

- The default request handler `app.get('*')` is for a Vue request
- The static file request handler, eg `app.use('/dist')` allows fallthrough
- Therefore, when a static (image or JavaScript) file is not found, the request falls through to the default handler
- The default handler attempts to render the page, which results in an error
- The error handler is called and it redirects to the "page-not-found" route
- The Page Not Found request is unnecessarily rendered and returned instead of a standard 404 status

This situation, a simple missing file, causes an increase in memory usage on the server and is worth fixing. However, a more serious condition can arise.

- If a theme file is missing, such as a header logo, the situation described above occurs
- When the Page Not Found is being rendered, the missing file is requested again
- The same steps are repeated and there's another redirect to page-not-found
- The server attempts to render the Page Not Found request again
- This is an infinite loop. Each image request results in a further page-not-found redirect and another request for the missing image and so on.

As this happens, each request uses more memory and very soon the limit is reached. In production mode, PM2 should then kill the server and restart it, but the problem remains.

## Expected behavior

A 404 file-not-found status is returned

## Steps to reproduce the issue

1. Delete /assets/logo.svg from a default installation
2. Open any page in a browser
3. Observe memory usage on the server

## Can you handle fixing this bug by yourself?

- [x] YES
- [ ] NO

Simple fix. Preparing a Pull Request

## Which [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html) state this refers to? Info for developer.
Pick one option.

- [x] This is a bug report for test version on https://test.storefrontcloud.io - In this case Developer should create branch from `develop` branch and create Pull Request `2. Feature / Improvement` back to `develop`.
- [ ] This is a bug report for current Release Candidate version on https://next.storefrontcloud.io - In this case Developer should create branch from `release` branch and create Pull Request `3. Stabilisation fix` back to `release`.
- [ ] This is a bug report for current Stable version on https://demo.storefrontcloud.io and should be placed in next stable version hotfix - In this case Developer should create branch from `hotfix` or `master` branch and create Pull Request `4. Hotfix` back to `hotfix`.

## Environment details

- Browser: Chrome
- OS: Mac
- Node: 8.11.4
- Code Version: 1.8.4, master, develop

patzick:
Thanks @grimasod !

pkarw:
[...]

PR #3002, opened 2019-05-31
Use 404 redirects only for defined cases

yuriboyko:
If some of resources (eg .png, .ico  etc) is missed  404 redirect  produces which cause performance issue as full 404 page render applies in this case 

This update allow to manage loading resource types which could be not allowed to use  404 redirect and as result do not lost performance


core/scripts/server.js
```diff
@@ -125,7 +125,20 @@ app.get('*', (req, res, next) => {
...
-      res.redirect('/page-not-found')
+      const contains = (target, pattern) => {
+        var value = 0
+        pattern.forEach(word => {
+          value = value + target.includes(word)
+        })
+        return (value >= 1)
+      }
+      if (contains(req.url, config.redirect404.notAllowedExtensions)) {
+        console.error(`Resource not found : ${req.url}`)
```
pkarw: I belive we should set the proper response  http header in here as well

config/default.json
```diff
@@ -449,6 +449,9 @@
           }
         ]
       }
+    },
+    "redirect404": {
+      "notAllowedExtensions": [".png", ".gif", ".jpg", ".jpeg", ".woff", ".eot", ".woff2", ".ttf", ".svg", ".css", ".js", ".json", ".ico", ".tiff", ".tif", ".raw"]
     }
```
pkarw: This should be a subsection of “server” section called “server.customErrors.http404.notAllowedExtensions”

core/scripts/server.js
yuriboyko: Done

config/default.json
yuriboyko: Done

yuriboyko:
@pkarw please check last update applied with filter preventing ssrRenderer of static content as we discussed

core/scripts/server.js
pkarw: I like it!

yuriboyko:
@pkarw changelog is added

pkarw:
@yuriboyko I've switched the branch to develop. Please remove the changes in the `docs` folder - they're messing with the current docs

core/scripts/server.js
```diff
@@ -125,10 +125,38 @@ app.post('/invalidate', invalidateCache)
...
+
+  if (contains(req.url, config.server.ssrDisabledFor.extensions)) {
+    return
+  }
+
   const s = Date.now()
   const errorHandler = err => {
     if (err && err.code === 404) {
-      res.redirect('/page-not-found')
+      if (contains(req.url, config.server.customErrors.http404.notAllowedExtensions)) {
```
patzick: here again would be
```suggestion
      if (NOT_ALLOWED_SSR_EXTENSIONS_REGEX.test(req.url)) {
```

or another const if this other setting from config is really needed here

core/scripts/server.js
```diff
@@ -125,10 +125,38 @@ app.post('/invalidate', invalidateCache)
...
+        console.error(`Resource not found : ${req.url}`)
+        res.setHeader('Content-Type', 'text/html')
```
patzick: instead of building response we should use utils:
```js
apiStatus(res, 'Vue Storefront: Resource is not found', 404)
```

config/default.json
```diff
@@ -14,7 +14,15 @@
...
       "dynamicConfigInclude": [],
-      "elasticCacheQuota": 3096
+      "elasticCacheQuota": 4096,
+      "customErrors": {
+        "http404": {
+          "notAllowedExtensions": [".png", ".gif", ".jpg", ".jpeg", ".woff", ".eot", ".woff2", ".ttf", ".svg", ".css", ".js", ".json", ".ico", ".tiff", ".tif", ".raw"]
+        }
+      },
+      "ssrDisabledFor": {
+        "extensions": [".png", ".gif", ".jpg", ".jpeg", ".woff", ".eot", ".woff2", ".ttf", ".svg", ".css", ".js", ".json", ".ico", ".tiff", ".tif", ".raw"]
```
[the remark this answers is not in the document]
yuriboyko: it was suggested by Piotr to have another one filter to prevent loading static content at all for some specific extensions.  onter one filter is related to verify if 404 has to be rendered.
i was thinking about this already and we probably can combine them but i'm not sure as it could be different cases.

core/scripts/server.js
```diff
@@ -125,10 +125,38 @@ app.post('/invalidate', invalidateCache)
...
 app.get('*', (req, res, next) => {
+  const contains = (target, pattern) => {
+    var value = 0
+    pattern.forEach(word => {
+      value = value + target.includes(word)
+    })
+    return (value >= 1)
+  }
+
+  if (contains(req.url, config.server.ssrDisabledFor.extensions)) {
```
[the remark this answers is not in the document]
yuriboyko: I like this suggestion. will do

core/scripts/server.js
```diff
@@ -125,10 +125,38 @@ app.post('/invalidate', invalidateCache)
...
+    return
```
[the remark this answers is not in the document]
yuriboyko: Good point. will do

core/scripts/server.js
yuriboyko: I like this suggestion. will do with another const for now. then they could be merged  later

core/scripts/server.js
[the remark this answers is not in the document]
yuriboyko: Done.

config/default.json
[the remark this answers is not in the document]
yuriboyko: Done.

core/scripts/server.js
yuriboyko: Done.

core/scripts/server.js
yuriboyko: Done.

core/scripts/server.js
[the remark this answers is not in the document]
yuriboyko: Done.

core/scripts/server.js
```diff
@@ -125,10 +125,38 @@ app.post('/invalidate', invalidateCache)
 app.get('/invalidate', invalidateCache)
 
 app.get('*', (req, res, next) => {
+  const contains = (target, pattern) => {
```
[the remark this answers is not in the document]
yuriboyko: Done.

config/default.json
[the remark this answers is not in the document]
patzick: I'd really would stay with single setting of this and possibly extent it that way, when somebody would have a real need to do that kind of stuff.
@pkarw could you add your opinion why to have two separate settings? I'm afraid its harder for user to get what is what that way.

core/scripts/server.js
```diff
@@ -125,10 +127,22 @@ app.post('/invalidate', invalidateCache)
 app.get('/invalidate', invalidateCache)
 
 app.get('*', (req, res, next) => {
+  if (NOT_ALLOWED_SSR_EXTENSIONS_REGEX.test(req.url)) {
+    utils.apiStatus(res, 'Vue Storefront: Resource is not found', 404)
```
patzick:
```suggestion
    apiStatus(res, 'Vue Storefront: Resource is not found', 404)
```

core/scripts/server.js
```diff
@@ -125,10 +127,22 @@ app.post('/invalidate', invalidateCache)
...
+    return
+  }
+
   const s = Date.now()
   const errorHandler = err => {
     if (err && err.code === 404) {
-      res.redirect('/page-not-found')
+      if (NOT_ALLOWED_CUSTOMERROR404_EXTENSIONS_REGEX.test(req.url)) {
+        utils.apiStatus(res, 'Vue Storefront: Resource is not found', 404)
```
patzick:
```suggestion
        apiStatus(res, 'Vue Storefront: Resource is not found', 404)
```

grimasod:
If I understand correctly, this seems to be the same as issue #2685, resolved with PR #2688.
My solution was to add `fallthrough: false` to the Express static file handler, so the following files will never be handled by Vue.js and they'll immediately return a status 404 if missing:

- Any file in /dist/
- Any file in /assets/
- /service-worker.js

I believe that should cover all image and js file requests, unless I'm mistaken.

I see this PR was originally based on the `master` branch, so my PR wouldn't have been included.

config/default.json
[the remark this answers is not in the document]
pkarw: Correct. `ssrDisabledFor` is more than enough. Le'ts keep it simple @yuriboyko

core/scripts/server.js
yuriboyko: Yes. sorry. had opened project which uses older release and it has  utils imported and uses utils.apiStatus ,  looks like i  just copied from the wrong window and did check

core/scripts/server.js
yuriboyko: Yes. sorry. had opened project which uses older release and it has  utils imported and uses utils.apiStatus ,  looks like i  just copied from the wrong window and did check

config/default.json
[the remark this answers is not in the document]
yuriboyko: Done

grimasod:
@yuriboyko @pkarw @patzick 
Guys, I don't mean to be rude, but maybe you missed my point yesterday. Could you please check this against the previous update I mentioned. Please consider:

- All static files (images, css, fonts) are in /assets/ or /dist/
- Requests to those locations are handled by Express
- Fallthrough is disabled, since #2688, so 404 is returned immediately if not found
- Such requests will not reach the default handler, `app.get('*', (req, res, next)`

```
const serve = (path, cache, options) => express.static(resolve(path), Object.assign({
  maxAge: cache && isProd ? 2592000000 : 0, // 1 month in milliseconds = 1000 * 60 * 60 * 24 * 30 = 2592000000
  fallthrough: false
}, options))
...
app.use('/dist', serve('dist', true))
app.use('/assets', serve(themeRoot + '/assets', true))
```

So, unless I'm mistaken, the Regex in this new PR will never match (as long as static files are only in assets and dist). You can test it on our VS site https://wonect.com/assets/no-such-file.png

pkarw:
@grimasod I right. This PR is probably not required. For sure not for the statics. However, it can still be useful for the paths that are not covered with the static middleware (so outside the `/assets`). I'd keep it merged

PR #3026, opened 2019-06-06
Update database tool README to use Yarn

Tjitse-E:
[The docs](https://docs.vuestorefront.io/guide/data/database-tool.html) on the database still use `npm` instead of `yarn`


patzick:
Thanks @Tjitse-E ! :)

PR #3061, opened 2019-06-17
rfc template improvements

lukeromanowicz:
Current docs template has some Rust language references. This PR fixes it


kkdg:
Thanks @lukeromanowicz

PR #3064, opened 2019-06-17
Storefront Cli

filrak:
PoC of Storefront CLI for scaffolding VS apps.

TODO:

- [x] Rename `create-storefront-app` to `vsf init`

TODO in next releases:
- Error handling
- `vsf upgrade`


src/modules/index.ts
```diff
@@ -73,7 +73,7 @@ export const registerModules: VueStorefrontModule[] = [
   PaymentBackendMethods,
   PaymentCashOnDelivery,
   RawOutputExample,
-  AmpRenderer,
+  // AmpRenderer,
```
pkarw: Why this is commented out?

packages/cli/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Installation
+
+``` 
+npm i -g @vue-storefront/cli 
+```
+## Commands
+
+After installing CLI you'll get access to following commands:
+- `create-storefront-app` - Use it to set up new VS project. The CLI will ask you a few questions and set up new, ready to work with Vue Storefront instance.
```
patzick: i've seen that you want another name, so needs to be changed here also

packages/cli/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@vue-storefront/cli",
+  "version": "0.0.4",
```
patzick: shouldn't we keep versioning compatible for easier maintaining?
```suggestion
  "version": "1.10.0-rc.1",
```

packages/cli/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "init": "./index.js"
+  },
+  "bin": {
+    "create-storefront-app": "./bin/createStorefrontApp.js"
```
patzick: script is init, shouldn't here be the same?

.eslintignore
```diff
@@ -1,2 +1,3 @@
 core/build/*.js
 node_modules
+packages/theme-capybara
```
patzick: no need to add this

.gitignore
```diff
@@ -29,6 +29,8 @@ core/resource/i18n/nl-NL.json
 core/resource/i18n/pl-PL.json
 core/resource/i18n/pt-BR.json
 core/resource/i18n/ru-RU.json
+_alpaca.scss
```
patzick: no need to add this

PR #3090, opened 2019-06-20
Change product color or size when product is in the cart

### LINKED ISSUE #2346: Lack of possibility to change color or size, when product is in the Cart
## Current behavior
Lack of possibility to change color or size, when product is in the Cart

## Expected behavior
User should has possibility to change color or/and size, when product is in the Cart

1. Go to the Cart, when user has at least one product inside.
2. Click on Edit next to this product

andrzejewsky:
### Related issues
closes #2346 

Hello VS team!
I've been learning Vue.js recently and this PR is my very first contact of it (besides some small things). I introduced possibility to change the color or size of the product that is in the cart. I spent some time with acquainting of this product and how it works and... I hope i've done everything correctly because I had many doubts about API, what should I use etc. Please have look at it.

Cheers 🎉

### Screenshots
Screenshot 2019-06-20 at 22 12 37

- [x] Test version (https://test.storefrontcloud.io) 
- [ ] RC version (https://next.storefrontcloud.io)
- [ ] Stable version (https://demo.storefrontcloud.io)


filrak:
Hello! This is a long-requested feature, thanks for proposing it! I'm leaving the review to @patzick who has recently worked with cart module

pkarw:
Thanks! I like the implementation the only question is the UI - not sure about it probably needs to be redesigned - @alinadivante can you take a look and consult it with our UX team please?

core/modules/cart/store/actions.ts
```diff
@@ -295,7 +296,7 @@ const actions: ActionTree<CartState, RootState> = {
         continue
       }
       if (config.entities.optimize && config.entities.optimizeShoppingCart) {
-        product = omit(product, ['configurable_children', 'configurable_options', 'media_gallery', 'description', 'category', 'category_ids', 'product_links', 'stock', 'description'])
```
pkarw: This change will increase the size of the order json passed to server endpoint api/order/create. We should remove these properties in order/placeOrder action just before passing them to the server

core/modules/cart/store/actions.ts
```diff
@@ -387,6 +388,20 @@ const actions: ActionTree<CartState, RootState> = {
       return diffLog
     }
   },
+  updateVariant (context, { product, configuration }) {
```
pkarw: Rename it to “configureItem” to make it coherent with “addIten”, “renoveItem” and also with “product/configure” actiona

core/modules/cart/store/actions.ts
```diff
@@ -572,12 +587,14 @@ const actions: ActionTree<CartState, RootState> = {
...
 
     /** helper - sub method to react for the server response after the sync */
-    const _afterServerItemUpdated = async function ({ dispatch, commit }, event, clientItem = null) {
+    const _afterServerItemUpdated = async function ({ dispatch, commit }, event, clientItem = null, serverItem = null) {
       Logger.debug('Cart item server sync' + event, 'cart')()
       diffLog.serverResponses.push({ 'status': event.resultCode, 'sku': clientItem.sku, 'result': event })
       if (event.resultCode !== 200) {
         // TODO: add the strategy to configure behaviour if the product is (confirmed) out of the stock
-        if (clientItem.server_item_id) {
+        if (!serverItem) {
```
pkarw: I’m not sure why this line has been added? There is another place where removeItem is executed

pkarw: I mean line 610: when there is no serverItem, calling the “removeItem” which executed the recursive “sync” + “merge” is pointless and could cause some side effects. We should just remove the item client side as it’s done in 610 I belive. Please let me know if there is some other reason why it was done differently?

src/themes/default/components/core/blocks/Microcart/Product.vue
```diff
@@ -95,17 +128,28 @@ import config from 'config'
...
   components: {
     RemoveButton,
     BaseInputNumber,
-    ProductImage
+    ProductImage,
+    ColorSelector,
+    SizeSelector
+  },
+  mixins: [Product, ProductOption],
+  beforeMount () {
```
pkarw: Would be cool to remove the event bus call - probably it requires a change to GenericSelector to call some event instead - but probably we should just keep it as it is and have it in mind with Product page refactor @filrak and @patzick - #3093

core/modules/cart/store/actions.ts
andrzejewsky: Yeah, the reason why I added this line it's because it doesn't deleting the item - it's the case when product is "out of stock".

Considering following steps:
1. Add item to the cart (size:S, color: black)
2. Change size to XS (that size is out of stock)
3. You receive the notification from the server "out of stock" but the product is still in the cart

I've noticed, it wasn't deleted because we have _server_item_id_ set (mutation: CART_DEL_NON_CONFIRMED_ITEM)

src/themes/default/components/core/blocks/Microcart/Product.vue
andrzejewsky: Agree, but I think it's a separated task because would be nice to change it everywhere  _GenericSelector_ occurs and get rid of this event bus inside

core/modules/cart/store/actions.ts
pkarw: That's a good point! OK, so `removeItem` should be fine (as it won't sync when there is `server_item_id` set).

src/themes/default/components/core/blocks/Microcart/Product.vue
pkarw: OK

pkarw:
Thanks for the comments - could you update the code regarding the other comments posted?

andrzejewsky:
@pkarw Just posted everything (look at file diff tab, squashed commits to just one), this is how it looks now:

Screenshot 2019-06-22 at 17 46 01

pkarw:
Cool. Regarding the UX - our design team will probably propose something, But wouldn't it be cool to have the controls much smaller and put along side the selected options?

Color: Blue [ ] [ ] [ ] [x] [ ] -> so you can switch the color right next to the selected onel the controls should be in that case like 8px in height.

I think that it would be also cool to have this feature as an option (enabled by default in the `config.cart.productsAreReconfigurable`

core/modules/order/store/actions.ts
```diff
@@ -19,6 +20,7 @@ const actions: ActionTree<OrderState, RootState> = {
    * @param {Order} order order data to be send
    */
   async placeOrder ({ commit, getters, dispatch }, order: Order) {
+    order.products = order.products.map(product => omit(product, ['configurable_options', 'configurable_children'])) as Order['products']
```
pkarw: Please just add the check `if (config.entities.optimize && config.entities.optimizeShoppingCart) {`

andrzejewsky:
@pkarw Update:

Screenshot 2019-06-22 at 19 45 21

alinadivante:
Thank you for your suggestion @andrzejewsky  ! :)
I think that Size can be displayed as drop-down list, but I will consult UX solution with @StasiekDivante

StasiekDivante:
@alinadivante to be honest I don't think people will use color / size change at cart page that often so we shouldn't risk complicating UI for the sake of usability here. I would go with simple "Edit" option next to "remove"

alinadivante:
@StasiekDivante thanks for your feedback.
My question is: will **_Edit_** activate the possibility of changing both parameters? or how will it work?

StasiekDivante:
Sure, edit would activate possibility to edit both options, it would look something like this:

Screenshot 2019-07-01 at 13 25 19

alinadivante:
@stasiekDivante interesting solution :)
what do you think @pkarw? In my opinion, it's fine!

pkarw:
Looks great! @andrzejewsky can you please change the UI regarding this design? Please do also resolve the conflicts in thie `core/pages/Product.js`

andrzejewsky:
@pkarw yeah, will be working on it this week.

pkarw:
Hi there! is it ready to be merged in? :)

andrzejewsky:
@pkarw not yet, I'm introducing these new changes, I'll finish tomorrow

andrzejewsky:
@pkarw Done, this is how it looks now:
Screenshot 2019-07-07 at 16 09 58
Screenshot 2019-07-07 at 16 10 07
demo

Please let me know before the merge, I'll squash commits to just one.

pkarw:
Looks great, thanks!

core/modules/cart/store/actions.ts
```diff
@@ -751,6 +769,15 @@ const actions: ActionTree<CartState, RootState> = {
   },
   toggleMicrocart ({ commit }) {
     commit(types.CART_TOGGLE_MICROCART)
+  },
```
pkarw: These methods are related to UI and therefore should be in the theme / Vuex store in the theme not in the core. Please move them there please

core/modules/cart/store/getters.ts
```diff
@@ -127,6 +127,15 @@ const getters: GetterTree<CartState, RootState> = {
...
       return itm.type_id === 'downloadable' || itm.type_id === 'virtual' // check for downloadable & virtual products
     })
+  },
+  isEditMode (state) {
+    return state.editMode !== null
+  },
+  editingProductId (state, getters) {
+    return getters.isEditMode && state.editMode.productId
+  },
+  selectedFilters (state, getters) {
```
pkarw: Should be moved to the theme

core/modules/cart/store/mutations.ts
```diff
@@ -86,6 +87,15 @@ const mutations: MutationTree<CartState> = {
...
     state.isMicrocartOpen = !state.isMicrocartOpen
+  },
+  [types.CART_OPEN_EDIT_MODE] (state, { productId, selectedOptions }) {
+    state.editMode = { productId, selectedOptions }
+  },
+  [types.CART_EDIT_MODE_SET_FILTERS] (state, { filterOptions }) {
+    state.editMode.selectedOptions[filterOptions.type] = filterOptions
+  },
+  [types.CART_CLOSE_EDIT_MODE] (state) {
+    state.editMode = null
```
pkarw: Should be moved to the theme

core/modules/cart/types/CartState.ts
```diff
@@ -10,5 +10,6 @@ export default interface CartState {
   cartItemsHash: string,
   cartServerLastSyncDate: number,
   cartServerLastTotalsSyncDate: number,
-  cartItems: any[]
+  cartItems: any[],
+  editMode: any
```
pkarw: Should be moved to the theme

core/modules/catalog/components/ProductOption.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+  },
+  methods: {
+    isOptionAvailable (option) { // check if the option is available
+      let currentConfig = Object.assign({}, this.configuration)
+      currentConfig[option.type] = option
+      return isOptionAvailableAsync(this.$store, { product: this.product, configuration: currentConfig })
+    }
+  }
+}
```
pkarw: @patzick it seems like a product page refactor as a side effect :)

andrzejewsky:
@pkarw done, but I'm not sure if registered module correctly. I was trying to extend the cart module but I've noticed that extendStore (found in docs) was removed, so used `registerModule` in `src/themes/default/index.js` maybe is there any other solution to extend vuex modules?

alinadivante:
@andrzejewsky it looks very nice!
But sometimes I get "{"code":500,"result":"Cart 617857 does not contain item 123798"}" when I update my product and I'm not sure why..
And what about bundle products? Don't we want to edit them from the cart? Only configurable products, right?

andrzejewsky:
@alinadivante yes, you can configure only the _configurable_ products. In terms of 500 error.. hmm looks weird, it's something like you haven't had cart synchronised with the backend yet, but you are changing variant - this is the first thing that came to my mind if i see the error message saying that your product isn't in the cart. I'll check that.

alinadivante:
I also have errors in the dev console on your branch @andrzejewsky 
 when I added group/bundle products
image

andrzejewsky:
@alinadivante ahhh yess.. because `productsAreReconfigurable` is the new property in he config file, so just add it in the `cart` section or just call:

`cp config/default.json config/local.json`

src/themes/default/index.js
```diff
@@ -17,6 +18,7 @@ function initTheme (app, router, store, config, ssrContext) {
   // You can do it on your own and then be able to customize the components used for example for German storeView checkout
   // To do so please execlude the desired storeView from the config.storeViews.mapStoreUrlsFor and map the urls by Your own like:
   // { name: 'de-checkout', path: '/checkout', component: CheckoutCustomized },
+  store.registerModule('themeCart', cartModule)
```
filrak: you can also just extend `ui-store` (use `extendModule` with module that has `ui` key)

andrzejewsky: @filrak Should I do this inside of `src/modules/ui-store` (like these comments says in this file)? In that case I'd have to import store configuration from `themes/default/store`, is that good idea? Or maybe should I create a new module (eg. `ui-store-cart`) with extending ui-store and move there all of store configuration that currently I have created in the `themes/default/store`?

andrzejewsky:
I've just made some changes (+ applied review) & editing mode for quantity as well, because when you are in the edit mode you shouldn't update anything until you click `update item` button for confirmation your changes. It solves also 500 errors @alinadivante and avoids the unnecessary requests.

pkarw:
Thanks! @andrzejewsky can you just please apply the changes suggested by @patzick ?

andrzejewsky:
Updated. I did some refactoring stuff, I moved options/filter logic to the helpers and tested it @patzick . Basically, we would be able to remove` ProductOptions.ts` and import just these helpers but.. maybe let's keep this for now... 
One more question... what about config option `productsAreReconfigurable` do we want to get rid of this? For now I just updated docs with information about this param.

alinadivante:
@andrzejewsky  In my humble opinion, there is one more thing to improve.
If I open the edit mode and change something, but I will use the "close" button instead of "Update item" and open the edit mode again, I see this new color / size.

image

I also noticed one problem ...
when I added products (as a logged in user) in the Chrome browser, then I opened Firefox browser and as a guest I added new products and after re-logging I saw that some products don't have editing options..
image

andrzejewsky:
@alinadivante I fixed this issue with current selected filters when you close edit mode and open it again.

The problem is with the second one... first of all if you don't see edit button it means that you don't have configuration options for that product (there is special condition for that).

Also, as I can see on your screenshot, the name of that product is weird. It says that there is no option besides **M** size and the **red** color: _Karissa V-Neck Tee-**M**-**Red**_. I don't even have that product in my catalog, I can find only `Karissa V-Neck Tee`.

So, it's maybe the last variant of that product? which is explainable why there are no configuration options and that means edit button shouldn't appear. @pkarw am I right? is that possible?

I'll try to reproduce that case (when there is just one option) later.

pkarw:
@andrzejewsky I think this is not the case - this product has more than 1 configurable_children for 99%

andrzejewsky:
so I'll try to find out why it happens, thanks @pkarw

alinadivante:
@andrzejewsky when I added this product (Karissa V-Neck Tee) there was Edit option.
It was only when I opened in the second browser and I logged in again, I didn't have edit option for this product.

alinadivante:
1. Log in to your account in VSF on one browser
2. Add few products
3. Open another browser
4. Add new product as guest
5. Log in to your account (wait for cart synchronised )

image

andrzejewsky:
@alinadivante this is exactly what I did and for me the edit option is always available, and displays immediately. I'll be working on it, maybe we have different configuration and i has some impact. Can you paste here (or somewhere else) your config `local.json`? I don't know if it will help me but I might try.

andrzejewsky:
It seems like it's a problem with synchronisation. I went thorough logic responsible for that and I came across for these lines:

https://github.com/DivanteLtd/vue-storefront/blob/6a9b581b145321eec4cc7921604703494433b3c1/core/modules/catalog/store/product/actions.ts#L414-L418

So when the product is simple, it removes the configuration options. @alinadivante sent me a cart state from vuex and it confirmed my thoughts - products that don't have edit options are marked as `simple`.

Screenshot 2019-07-11 at 13 45 08

I'm unable to reproduce it, looks like everything works for me but later I'll try to find out why some product are `simple` for her but not for me.

pkarw:
The problem is that with the `cart/sync` method we're adding the simple products to the shopping cart - Magento is providing us witt the `simple` products only (MH01-XL-Gray) for example  not MH01 (configurable) So this line:https://github.com/DivanteLtd/vue-storefront/blob/9b6368ccdc1f9e61c1e9f1349b5cb3e0bae991f7/core/modules/cart/store/actions.ts#L709 is adding the `simple` products to the cart.

There is no difference from the user point of view. But there is a huge difference from this feature point of view :)

The workaround - we would have to find the parent product in here: https://github.com/DivanteLtd/vue-storefront/blob/9b6368ccdc1f9e61c1e9f1349b5cb3e0bae991f7/core/modules/cart/store/actions.ts#L712 -> configure it ([configureProductAsync](https://github.com/DivanteLtd/vue-storefront/blob/9b6368ccdc1f9e61c1e9f1349b5cb3e0bae991f7/core/modules/catalog/helpers/index.ts#L482) and just after that add to the cart.

How to find the parent product? https://github.com/DivanteLtd/vue-storefront/blob/9b6368ccdc1f9e61c1e9f1349b5cb3e0bae991f7/core/modules/catalog/store/product/actions.ts#L195

I don't like this solution as it will add another network call to the cart sync :( So if we're to handle this case it must be optional (just in case of `config.cart.productsAreConfigurable = true`)

OK, there could be just one network call: instead of `product/single` we can use just the query above - modified to support kind of logick: `configurable_children.sku = <serverItem.sku> || sku = <serverItem.sku>`. The downside is that `product/single` is using cache so most of the times it doesn't run the network call at all.

But i can't find better option to handle this

pkarw:
Another option - optimization - is to add the `parentSku` field to the indexer: https://github.com/DivanteLtd/mage2vuestorefront/issues and https://github.com/DivanteLtd/magento2-vsbridge-indexer and https://github.com/DivanteLtd/magento1-vsbridge-indexer

But - we have also other indexers (different platforms) so we can't assume this field is always there. Please do support it with the 3rd option presented - and this optimization could be a option for development

StasiekDivante:
Hey @andrzejewsky could we add a small animation to transition between states? Right now it looks a bit rough when you go to (or exit) edit mode.

If you need I can create animation for you

andrzejewsky:
@StasiekDivante There is a small transition when you open and close edit mode.

> If you need I can create animation for you

yeah, if you have some ideas, feel free 😛

andrzejewsky:
@pkarw Ok I understand the problem, besically I have found exactly the same lines, but first of all, why I'm unable to achieve the same result as @alinadivante did? In my case I can't see any `simple` products, all of them are configurable even if I add product as a guest and then login and wait for the sync. I even have the same configuration file. Is there any reason?

pkarw:
You must have the cart saved on your magento2 account so you must create account (in vs) add products to cart logout then login back so the cart will be back synced with  Magento and products from magento cart restored to vs cart

andrzejewsky:
@pkarw this is exactly what I did, result the same - all of products always are configurable instead of `simple`. Tested with `test.storefrontcloud.io` as a backend, same with `demo.vuestorefront.io` even with `vue-storefront-api` locally. So if it works for you, maybe I made something wrong with my setup.

pkarw:
@alinadivante what do you think?

alinadivante:
@andrzejewsky add more configurable products (5-10), being logged in, and in the second browser window, as a guest, also add  about 5 products first, then log in.

I hope that with more products in the cart you will reproduce it! :)

andrzejewsky:
@pkarw Ok finally I've got the same result. So, I'll implement some workaround for simple products as well (based on your comments). Thanks to @alinadivante for sending me config files!

andrzejewsky:
Updated. I've made a little bit refactoring because of lack of possibility to fetch parent product when I have only simple one. As the result `catalog/store/product/actions.ts` is smaller, because I extracted some logic from `list` action to helpers. Now two new (reusable) functions are available:

`findProductsByQuery` - it was used in the `list` action for loading products in the catalog
`findConfigurableParent` - for loading parent product (+configured if configuration is passed)

`createProductListHelpers` - is just a extracted logic from `list` action

Now, editing mode works even with the simple products. When you click "Edit" the configurable parent will be loaded.

Also I covered the case when there are two products with the same `sku` - it's possible because in the edit mode you can configure the same filters as some product in the cart already has. In that case the next product with same `sku` will be deleted and the edited one will increase `quantity` by the `quantity` of deleted one.

pkarw:
Good job with the refactoring! @patzick just takie a look - product/list just got much much clear and it’s something that might be pretty useful for you for the Product.vue refactoring

core/modules/order/store/actions.ts
```diff
@@ -19,6 +20,9 @@ const actions: ActionTree<OrderState, RootState> = {
    * @param {Order} order order data to be send
    */
   async placeOrder ({ commit, getters, dispatch }, order: Order) {
+    if (config.entities.optimize && config.entities.optimizeShoppingCart) {
+      order.products = order.products.map(product => omit(product, ['configurable_options', 'configurable_children'])) as Order['products']
```
pkarw: Related to: https://github.com/DivanteLtd/vue-storefront/pull/3222

alinadivante:
Fixed :) 
image

CHANGELOG.md
```diff
@@ -72,6 +72,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Products column change functionality - @vishal-7037 (#3017)
 - New Module order-history this provides the pagination via lazy laod - @hackbard (#2810)
 - OrderNumber on ThankYouPage - @Flyingmana (#2743)
+- Added possibility to change color or size of the product that is already in the cart - @andrzejewsky (#2346)
```
patzick: this is PR to develop, so to newest version. You added it to 1.10 which was already released.
Please change it to 1.11 and read about [Releasing cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html)

core/modules/catalog/helpers/search.ts
```diff
@@ -0,0 +1,129 @@
...
+          configurableChild[opt.attribute_code] = opt.value
+        }
+      }
+    }
+  }
+
+  return product
+}
+
+const configureTaxes = (context, { resp, isCacheable, cacheByKey }) => {
```
patzick: basically if helper needs a vuex context, then it should be an action not helper. Otherwise helpers instead of helping they are distracting vuex logic to parts.
Helpers should do computations etc. - take data and return result, which is also easy to test by unit tests.
But vuex actions should be the ones, which are dispatching other actions.
Let me know if you understand this point of view - it is really important to have this understanding of helpers.
cc @pkarw @filrak @przspa @lukeromanowicz

core/modules/catalog/types/ProductConfiguration.ts
```diff
@@ -0,0 +1,10 @@
+export interface Configuration {
```
patzick: i'd change it to ProductOption. Configuration is more reserved to application Configuration.

src/themes/default/store/cart/actions.ts
```diff
@@ -0,0 +1,35 @@
+import * as types from './mutation-types'
+import { findConfigurableParent } from '@vue-storefront/core/modules/catalog/helpers/search'
+import getCurrentConfigurationFromTotals from './helpers/getCurrentConfigurationFromTotals'
+
+const actions = {
+  configureProduct (context, { product }) {
+    return new Promise((resolve, reject) => {
```
patzick: please use async/await instead of Promise syntax

core/modules/catalog/helpers/search.ts
andrzejewsky: @patzick completely agree, basically I wanted to ask about this point just to clarify what helpers means in this project. My understanding was exactly like yours but I saw context-bound in the other helpers so I thought that sometimes you are using it in that way and that's allowed. Ok I'll try to refactor it as much as it is possible, because probably some of them would be useful later.

patzick: yes, we have that kind of helpers, but we're trying to refactor them in time with this conception. That's cool that you see this the same way. I agree that some of them are reusable, but only with logic part - with every dispatch there should be an action, which can also be reusable if is doing one specific thing at the time not set of things :)

andrzejewsky:
@patzick I applied your comments + refectored actions/helpers - everything that I could, was extracted to the helpers and one big `list` action was split to the smaller ones. I think some of them could be refactored even more + would be nice to have helpers tested, but I think maybe it's something for different taks because this feature becoming gradually more complex.

pkarw:
@patzick for me it's ready to be merged in :)
@andrzejewsky - great job

andrzejewsky:
Any update of that? Could it be merged? Because it's only waiting and I have to resolve conflict sometimes.

PR #3092, opened 2019-06-21
Robots.txt from theme

shkodasv:
### Related issues
https://github.com/DivanteLtd/vue-storefront/issues/2110

It is more flexible to have robots.txt in the theme instead of src/server/robots.js

### DIFF
--- a/src/server/robots.js
+++ b/src/server/robots.js
@@ -1,6 +1,6 @@
 const fs = require('fs')
 const themePath = require('../../core/build/theme-path')
-const robots = `${themePath}/assets/robots.txt`
+const robots = `${themePath}/robots.txt`
 
 module.exports = expressApp => {
   expressApp.get('/robots.txt', (req, res) => {
--- a/src/server/robots.js
+++ b/src/server/robots.js
@@ -1,6 +1,14 @@
+const fs = require('fs')
+const themePath = require('../../core/build/theme-path')
+const robots = `${themePath}/assets/robots.txt`
 
-module.exports = (expressApp) => {
+module.exports = expressApp => {
   expressApp.get('/robots.txt', (req, res) => {
-    res.end('User-agent: *\nDisallow: ')
+    if (fs.existsSync(robots)) {
+      res.type('text/plain')
+      res.sendFile(robots)
+    } else {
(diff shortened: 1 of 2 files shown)


src/server/robots.js
```diff
@@ -1,6 +1,14 @@
...
 
-module.exports = (expressApp) => {
+module.exports = expressApp => {
   expressApp.get('/robots.txt', (req, res) => {
-    res.end('User-agent: *\nDisallow: ')
+    if (fs.existsSync(robots)) {
+      res.type('text/plain')
+      res.sendFile(robots)
+    } else {
+      res.end('User-agent: *\nDisallow: ')
```
pkarw: I believe it might be even a simpler solution just adding a static file server for robots.txt in the `core/scripts/server.js` (https://github.com/DivanteLtd/vue-storefront/blob/e14a2a0abde6f71a48e21c8e694b5502e50d3aa3/core/scripts/server.js#L112) + creating a default file with the `User-agent: *\nDisallow`)

filrak: @pkarw don't you think such case-specific things shouldn't land in core?

filrak: Maybe we should have a directory in theme that maps directly to server root path so people can put there whatever they want (we should also place assets in there)

filrak:
Could you please tell motivation why it's better to have it theme? From my PoV it adds unnecessary complexity

src/server/robots.js
pkarw: Sure, there should be assets server bound to robots.txt from these exactly as we do have with the assets  (see example above)

shkodasv:
@filrak 
Hello!
Because the theme is safe place to put everything specific to project. When we have robots.txt in src/server/robots.js it is hard to have different robots.txt for different themes, hard to update it, also a possibility to override the file during upgrade.

filrak:
@shkodasv I'm not convinced. can you please give a real-world example when such thing is needed?
It's something that could potentially break others stores in a future if refactored/changed so new api should be included only if absolutely necessary.

pkarw:
Closing becuase there are better ways to handle this issue (static file hosting via proxy or just assets + alias)

PR #3152, opened 2019-06-27
Modules 2.0

filrak:
### Related issues

closes https://github.com/DivanteLtd/vue-storefront/issues/3144

New modules API. Motivations can be found on https://github.com/DivanteLtd/vue-storefront/issues/3144

### Scope

- [x] New modules API
- [x] Create independent lerna package for `@vue-storefront/module`
- [x] Module template
- [x] Module npm package boilerplate
- [x] Hooks API
- [ ] Define Hooks
- [ ] Unit tests
- [ ] Rewrite core modules to new API
- [x] Add depreciation note on old modules via `Logger`
- [x] Build `@vue-storefront/module` as npm package in vanilla js and publish


pkarw:
I like this idea a lot! Really! Let's make it happen in 1.11.

Some insights from my side:
- @patzick suggested to have the events definitions on the module level - I like this idea,
- we don't have any mechanism to unbind the events - isn't there a risk that one function will be added and then executed multiple time if the user won't be avare of this consequence? I mean when you add the listener in the `mounted()` - it will be added each time the component is mounted. I think it would be safe to add kind of hashing / safety key for that (maybe an optional key like `listenerUniqueIdentifier` so the user could control - for the selected event listeners if they're added always once or multiple times) ?

filrak:
>  @patzick suggested to have the events definitions on the module level - I like this idea,

You mean to let people add their own hooks inside module? I think making factory methods for public hooks would be sufficient in tat case. @patzick WDYT?

@pkarw regarding your concerns about hooks. Imho we shouldn't have any under the hood checks like this:
- first: hooks are generally not meant to be used in components (but ofc they can be)
- second: depending on situation user might want to run hook from a given component multiple times (e.g. while placing an order). If he/she wants to make it a one-timer it's straightforward to add some flag inside hook function (`onHook(() => { if(flag) something() })`) to have this feature. Imho iImho we shouldn't expand our API for things that are easily doable for the users anyways It's another thing to maintain and think about in terms of breaking changes yet gives us nothing in return.

pkarw:
@filrak OK, thanks for the comment

filrak:
Thank you! ;)

filrak:
Most of the work has been done. You can start reviewing :)

core/modules/catalog/index.ts
```diff
@@ -1,20 +1,33 @@
...
-    { key: 'tax', module: taxModule },
-    { key: 'category', module: categoryModule }
-  ] },
-  beforeRegistration
-})
+export const CatalogModule: StorefrontModule = function (app, store, router, moduleConfig, appConfig) {
+  const storeView = currentStoreView()
+  const dbNamePrefix = storeView.storeCode ? storeView.storeCode + '-' : ''
+
+  initCacheStorage('categories')
```
pkarw: We should refactor this to ‚StorageManager.set’ - to have the same syntax like above (this function does exactly the same just for different storeName). The attributes store is the only one still needed (we can remove the other ones)

packages/module/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.js
+*.d.ts
```
patzick: why type definition files should be ignored here?

packages/module/tests.spec.ts
```diff
@@ -0,0 +1,18 @@
+// import { refs, injectReferences } from "./index"
```
patzick: remove or update :)

filrak: I commented to update in next iteration (i want to merge it asap bc of conflicts) but If this hurts your eyes i can delete and copy from older commits

packages/module/.gitignore
filrak: All. During a build step and publishing step i compile ts to js/d.ts files. It's just to prevent autocommiting them.

core/modules/catalog/index.ts
pkarw: We need to remove `categories` and `products` keep only `attribbutes`

patzick:
@filrak please fix tests so we could merge this PR :)

filrak:
@patzick fixed by @lukeromanowicz ;)

PR #3176, opened 2019-07-04
Fix #3102 - Date fitler ignoring format. Use format object

### LINKED ISSUE #3102: Date filter is ignoring the format parameter
## Current behavior

I'm trying to change the format of a date within a template using the `date` filter:
```
<p>{{ item.created_at | date('format as a parameter here') }}</p>
```

But the date filter is in fact not using the parameter at all:
https://github.com/DivanteLtd/vue-storefront/blob/d40ba3b958c26d77d2d4c98bdf403af1ffd19275/core/filters/date.js

And therefore the date is always returned in the same format (e.g. `25.6.2019, 11:50:32`).

## Expected behavior
I should be able to choose the format when using the filter.

Try to pass any argument to the `date` filter within a template:
```
<p>{{ item.created_at | date('format as a parameter here') }}</p>
```
No matter what you pass in, nothing happens.

## Repository
The latest commit for the file in the current `master` branch: https://github.com/DivanteLtd/vue-storefront/blob/d40ba3b958c26d77d2d4c98bdf403af1ffd19275/core/filters/date.js

- [ ] YES
- [x] NO

Pick one option.

- [x] This is a bug report for current Release Candidate version on https://next.storefrontcloud.io - In this case Developer should create branch from `release` branch and create Pull Request `3. Stabilisation fix` back to `release`.

## Environment details
- Browser: Any
- OS: Any
- Node: Any
- Code Version: v1.9.1

grimasod:
### Related issues

closes #3102

Date filter was not using format. Now uses a format object. Applied to default theme Order details view.

Before
Screen Shot 2019-07-04 at 11 11 11 AM

After
Screen Shot 2019-07-04 at 11 10 07 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.10.0-rc.2] - UNRELEASED
 
 ### Fixed
-- Fixed wrong meta description attribute by page overwrite - @przspa (#3091) 
+- Fixed wrong meta description attribute by page overwrite - @przspa (#3091)
+- Fixed date filter ignoring format param - @grimasod (#3176)
 
 ## [1.10.0-rc.1] - 2019.06.19
 
--- a/core/filters/date.js
+++ b/core/filters/date.js
@@ -3,9 +3,9 @@ import { currentStoreView } from '@vue-storefront/core/lib/multistore'
 /**
  * Converts date to format provided as an argument or defined in config file (if argument not provided)
  * @param {String} date
- * @param {String} format
+ * @param {Object} format
  */
-export function date (date, format) {
+export function date (date, format = {}) {
   const storeView = currentStoreView()
-  return new Date(date).toLocaleString(storeView.i18n.defaultLocale)
+  return new Date(date).toLocaleString(storeView.i18n.defaultLocale, format)
 }
--- a/core/filters/date.js
+++ b/core/filters/date.js
@@ -3,9 +3,46 @@ import { currentStoreView } from '@vue-storefront/core/lib/multistore'
 /**
  * Converts date to format provided as an argument or defined in config file (if argument not provided)
  * @param {String} date
- * @param {Object} format
+ * @param {String} format
  */
-export function date (date, format = {}) {
+export function date (date, format = '') {
   const storeView = currentStoreView()
-  return new Date(date).toLocaleString(storeView.i18n.defaultLocale, format)
+  let options = {}
+  switch (format) {
+    case 'LT':
+      options = {hour: 'numeric', minute: 'numeric'}
+      break
+    case 'LTS':
+      options = {hour: 'numeric', minute: 'numeric', second: 'numeric'}
+      break
+    case 'L':
+      options = {year: 'numeric', month: '2-digit', day: '2-digit'}
+      break
+    case 'l':
+      options = {year: 'numeric', month: 'numeric', day: 'numeric'}
+      break
+    case 'LL':
+      options = {year: 'numeric', month: 'long', day: 'numeric'}
+      break
+    case 'll':
+      options = {year: 'numeric', month: 'short', day: 'numeric'}
+      break
+    case 'LLL':
+      options = {year: 'numeric', month: 'long', day: 'numeric', hour: 'numeric', minute: 'numeric'}
+      break
+    case 'lll':
+      options = {year: 'numeric', month: 'short', day: 'numeric', hour: 'numeric', minute: 'numeric'}
+      break
+    case 'LLLL':
+      options = {year: 'numeric', month: 'long', day: 'numeric', weekday: 'long', hour: 'numeric', minute: 'numeric'}
+      break
+    case 'llll':
+      options = {year: 'numeric', month: 'short', day: 'numeric', weekday: 'short', hour: 'numeric', minute: 'numeric'}
+      break
+    default:
+      // use empty object
+  }
+  console.log(format)
+  console.log(options)
+  return new Date(date).toLocaleString(storeView.i18n.defaultLocale, options)
 }
--- a/src/themes/default/components/core/blocks/MyAccount/MyOrder.vue
+++ b/src/themes/default/components/core/blocks/MyAccount/MyOrder.vue
@@ -17,7 +17,7 @@
     <!-- My order body -->
     <div class="row fs16 mb20">
       <div class="col-xs-12 h4">
-        <p>{{ order.created_at | date('MMMM D, YYYY') }}</p>
+        <p>{{ order.created_at | date({year: 'numeric', month: 'long', day: 'numeric', hour: 'numeric', minute: 'numeric'}) }}</p>
         <p class="mt35">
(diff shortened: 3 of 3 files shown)


grimasod:
@patzick Thanks for the feedback. I know what you mean, a string would be easier and more familiar. There are 2 reasons why I used an object:

1. That's what JavaScript uses in Date.toLocaleString(), so it can be passed in directly. Of course, it's possible to parse the string and map it into an object, with a little effort, but...

2. The string format would not be used for the output and could be misleading. For example, "YYYY-MM-DD" will not output "2019-06-25", because it depends on the locale in the store view i18n setting. In this example, it could output as "25/06/2019", or "2019. 06. 25".

I think it's important to keep the local setting, at least as an option. We could have a second filter, "localeDate" or something like that. What do you think?

> Also please create branch from release/v1.10 because it's stabilization fix and should go there as well :)

Sure :) So, should I close this PR and create a new one? I'm still learning the finer details of GH!

patzick:
@grimasod great discussion and arguments :)

In this case i'd propose not to use `Date(date).toLocaleString` and use `Moment.js`, we already have it in dependencies, and it fully supports locales https://momentjs.com/docs/#/i18n/instance-locale/

then output would be with locale as default ( format as `L`), and custom if developer passes format `{{ order.created_at | date('YYYY-MM-DD') }}` then should use this string format.

How do you think about that solution?

And yes - unfortunately in this case we need new PR, because branch needs to be created from `release/v1.10` :)

grimasod:
@patzick yeah :)

Great idea! There's no need to reinvent the wheel. It would give maximum flexibility for developers.

Just once minor point - currently the default includes the time, so I think we'd need to output the LT format as well as L.v I'm not so familiar with moment.js, just looking into it now.

Ok, no problem. I'll leave this open until the new one is ready

patzick:
Sure, if with time docs will tell you how to properly do this.
Good luck, can't wait to see it :)

pkarw:
@patzick are you sure about moment.js? I remember that I removed it because it weighed way tooooo much as for bundle size we need always to minimize. Almost sure we’re not using moment anywhere in the code (not importing it therefore it’s not a part of the bundle even when it left in package.json). I agree on string as @patzick wrote but not specifically about moment library

grimasod:
Right. In a quick test it added 265 KB to app.js.

However, writing a fully featured date formatter is quite complicated in JS. And, as mentioned above, we need to consider the locale.

So how about only implementing the named masks for localized formats that moment.js uses: "L", "LL", "LLL" etc.

Or something a bit more verbose like "shortDate", "mediumDate", "LongDate", etc, (similar to the Steven Levithan formatter  http://blog.stevenlevithan.com/archives/date-time-format)

grimasod:
@patzick @pkarw See updated file for a possible solution

It's not fully optimized or tested. If it's agreed to, I'll make a new PR

patzick:
@pkarw yopu're absolutely right, i've always included it lazily.
We have two possible alternatives - dayjs (96% smaller - 2.74kB, and with API similar to moment.js) and date-fns (90% smaller - 6.54kB)
I would go with `dayjs`

@grimasod i'm strongly against implementing it and not using good existing library; there will be more and more edge cases and we will end up having custom date formatters.
What are you thinking about day.js? Maybe with implementing it you'd completly remove moment.js from dependencies and replaces all usages of it? :)

pkarw:
Cool; day.js seems just like a right fit

grimasod:
@patzick I had a look a dayjs. For specified formatting it's nice and easy to use, but for i18n it's not so simple.

It has locale extensions, that have to be individually imported. Unfortunately it has a mixture of extensions using 2-character language codes (e.g. "en" for en-US) and ones using full locale codes, (like "en-NZ"). This makes it difficult to match up with VS store locales.

Of course it's possible to add some logic for figuring that out. We would only want to run it once - not every time the function is called. So where would be the best place to do that? Maybe in `app.ts`, where the filters are imported.

I'm pretty busy, so it might take a few days

patzick:
@grimasod worst case we could have simple mapping for that, that's not a problem:)
going further - date filter should lazily load locale, then set it and display result. Lazy import will not be invoked again so we have that covered, and even better - won't be invoked if not needed as it would be in app.ts :)

grimasod:
@patzick Ok. I'll look into it and make a new PR :)

patzick:
@grimasod i've found that day.js has same locale as moment.js - https://github.com/iamkun/dayjs/pull/594#issuecomment-491663275
 so from moment docs:
```
Moment will also try locale specifier substrings from most-specific to least-specific until it finds a locale it knows. This is useful when supplying Moment with a locale string pulled from the user's environment, such as window.navigator.language.

moment.locale('en-NZ'); // 'en'
Finally, Moment will search intelligently through an array of locales and their substrings.

moment.locale(['en-NZ', 'en-AU']); // 'en-au', not 'en'
```

i believe we can assume, we don't need any mapping, this should be somehow resolved as well. We need to dive deeper, and for sure there is a way to match our locales

grimasod:
@patzick I tried this out, but unfortunately day.js doesn't have the same locale matching system as moment.js. So, for example, setting the local to 'en-US' will fail and cause errors.

You're right though, we don't need a mapping. We just need to add some logic like moment.js has.

For example, if the store locale is 'xx-YY'

- Test if 'xx-YY' available
- If available import and apply
- Else test if 'xx' is available
- If available import and apply
- Else do nothing (uses default 'en' locale)

Or something like that. What do you think?

patzick:
@grimasod this sounds good, but this will be lazy import then, right? Should be tested with some kind of POC :)
Other possibility would be to add PR to day.js adding locales from config ` "availableLocale": ["en-US","de-DE","fr-FR","es-ES","nl-NL", "jp-JP", "ru-RU", "it-IT", "pt-BR", "pl-PL", "cs-CZ"],` but i'm not sure if they accept and also if that covers our use cases.
So your idea sounds really good, but we need to ensure this works as expected:)

grimasod:
@patzick This is proving to be quite complicated! I've been trying to create a POC, but struggled with async functions in the filter, for the lazy imports. Actually it seems to be impossible. See:
https://github.com/vuejs/vue/issues/5217

So, alternative options could be:

- Initialize data.js in app.js. As you pointed out, this means it always gets invoked, so not ideal
- Don't use a filter. Use computed values instead.

Any other ideas?

patzick:
@grimasod i'd go with computed property (or even method - lazy loading locals is worth it), even further - helper which can be used in computed property; then we have flexibility (can be used anywhere) and caching provided by computed property.
Are we ready to close this one and wait for new proposition? :)

grimasod:
@patzick I'm trying to get this working with a helper, but without success so far. TBH, I'm not so familiar with lazy loading, apart from using lifecycle events like created or mounted. I'm also pretty busy, so don't have that much time this week.

patzick:
@grimasod understand, so maybe we can collaborate on this? :) just create next WIP PR with helper and simple case with normal loading locale, and we will just change this together. Remember to check possibility to make changes in branch:
image


we will get through this together :)

grimasod:
@patzick ok, thanks! I'll close this one

Issue #2985, opened 2019-05-28
Clear parts of local storage after deployment


nhp:
## What is the motivation for adding / enhancing this feature?
There should be a possibility for targeted local storage flushing. Like after a deployment or rebuild this could be needed to have the current information available at the client side. Right now it is up to the user to do so and that is not really helpful for getting updated stuff out to the browser.
If the user completely throws their local storage we need to take care of not accidentally clearing the cart information.

## What are the acceptance criteria 
* remove targeted stuff from local storage
* remove everything beside session information

## Can you complete this feature request by yourself?
* [ ] YES
* [x] UNSURE
* [ ] NO

## Which [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html) state this refers to? Info for developer.
Pick one option.

* [x] This is a normal feature request. This should be available on https://test.storefrontcloud.io and then after tests this can be added to next Vue Storefront version. In this case Developer should create branch from `develop` branch and create Pull Request `2. Feature / Improvement` back to `develop`.
* [ ] (**Pick this option only if you're sure**) This is an important improvement request for current Release Candidate version on https://next.storefrontcloud.io and should be placed in next RC version. In this case Developer should create branch from `release` branch and create Pull Request `3. Stabilisation fix` back to `release`.
* [ ] (**Pick this option only if you're sure**) This is a critical improvement request for current Stable version on https://demo.storefrontcloud.io and should be placed in next stable version. In this case Developer should create branch from `hotfix` or `master` branch and create Pull Request `4. Hotfix` back to `hotfix`.

## Additional information
According to w3c this is intended behavior as the user or the application is responsible for the local storage only. There is no TTL intended, but there are different options we found: https://stackoverflow.com/questions/2326943/when-do-items-in-html5-local-storage-expire but nothing seems "ready to use"

dennisotugo:
+1

grimasod:
I'm going to look into this. After a release, we often get enquiries from users who can't add items to the cart and experience other issues. Sometimes it's fixed by simply reloading the page. Other times we have to instruct them to clear their browser cache, which is not ideal.

As a first step, I'm thinking of showing an error alert with a button to reload the page. The message would be something like:

_A new version of this website is available. Please click the button below to reload and update to the latest version._

dennisotugo:
@pkarw How do you think this should be handled?

grimasod:
This is a complicated issue. Even for just updating the service worker. See:
[...]

PR #3180, opened 2019-07-04
Feature/2985 - a.k.a "localStorage saver"

pkarw:
### Related issues

This is related to #2985 

I've decreased the `localStorage` quota usage + error handling by introducing new config variables: 

- `config.products.disablePersistentProductsCache` to not store products by SKU (by default it's on). Products are cached in ServiceWorker cache anyway so the `product/list` will populate the in-memory cache (`cache.setItem(..., memoryOnly = true)`); 
- `config.seo.disableUrlRoutesPersistentCache` - to not store the url mappings; they're stored in in-memory cache anyway so no additional requests will be made to the backend for url mapping; however it might cause some issues with url routing in the offline mode (when the offline mode PWA installed on homescreen got reloaded, the in-memory cache will be cleared so there won't potentially be the url mappings; however the same like with `product/list` the ServiceWorker cache SHOULD populate url mappings anyway); 
- `config.syncTasks.disablePersistentTaskQueue` to not store the network requests queue in service worker. Currently only the stock-check and user-data changes were using this queue. The only downside it introuces can be related to the offline mode and these tasks will not be re-executed after connectivity established, but just in a case when the page got reloaded while offline (yeah it might happen using ServiceWorker; `syncTasks` can't be re-populated in cache from SW)

If by some reasons you wan't to have the `localStorage` back on for `Products by SKU`, `Url Routes` and `SyncTasks` - please juset set these variables back to `false` in your `config/local.json`.

- [ ] No upgrade steps required (100% backward compatibility and no breaking changes)


patzick:
Looks good, i have nothing to add:)
i set label to test it by QA on branch before merge to be sure

pkarw:
I've just changed the config property names changing the `LocalStorage` to `Persistent`. The `UniversalStorage` supports also IndexedDb, WebSQL so having LocalStorage in the property names could have been missleading

alinadivante:
hi @pkarw I started vsf with devServiceWorkerr=true ( `yarn dev:sw` ) and comparing your branch to Demo I see that less items are now stored in Local Storage or Cache Storage (i suppose that it is OK). Please write me if I should check something else.


But please look also at this error from my dev console:
image

pkarw:
@alinadivante thanks for testing this out! Regarding this error - it's not related to the changes included in this PR. It's the reason of "QA rejected"? :-(

alinadivante:
@pkarw yes, it was the reason :P I didn't notice this error before, so I assumed that your changes are causing it. If not, I think the rest is fine.

PR #3217, opened 2019-07-11
Refactor Product Page

### LINKED ISSUE #3093: Product page refactoring for theme 2.0
This is the initial plan to refactor the `Product.vue` page accordingly to #2696 and after the initial learnings we got from the `Category.vue` refactoring (#2890)

The goal is to prepare the `Product.vue` for Theme 2.0 - based on Storefront UI but without dramatic changes, keeping the business logic as is (it's very important assumption) and share it between `product` and `product-next` modules. The same time: improving the Vuex readability, preparing the Product page data interface for the 2.0

Stage 1: to be finished till end of July (preferred date: 3rd week of July)

- [ ] Create the `product-next` module for the consistency with `category-next` - use only those new one’s in the new Product.vue and Category.vue
- [ ] We should re-use as much the original logic of `product` module as we can: via helpers or just via importing the actions directly from the old module (doing that way we don’t have to create additional dependencies and can just do like: `import { list } from `@vue-storefront/store/modules/product/actions`
- [ ] Add suitable getters with the proper naming convention -and don’t use the state directly in the theme (just only thru getters)
- [ ] We don’t touch the business logic: `product/helpers` and the actions like `setupBreadcrumbs`, `setupAssociated`, actions related with bundle etc. Maybe just some minor refactoring like:
    - [ ] avoiding state mutations directly in actions (move to mutations)
    - [ ] remove `rootStore` references,
    - [ ] limit the logic based on events
- Don’t change the state much - to keep the single code base with product 1.0 module as far as we can:
    - [ ] we must keep `current`, `original`, `parent`
    - [ ] we must keep `current_bundle_options`, `current_custom_options`, `custom_options_validators`
    - [ ] we can remove (from both - original and new module): `productLoadStart`, `productLoadPromise`, `productGallery` (or if not, rename it to keep the convention which is `under_scored`)
    - [ ] remove `color` and `size` from `current_options` - change the theme to not access this property ditrectly but via getter
    - [ ] we can remove `list` -> but we must refactor the action `product/list` (extract to helper which returns the data + mutate the state after helper is executed in the original `product/list`; in the `product-next/list` (or whatever different name it will have) we can remove this state property as the list ist now in the category)
- Rename getters to the proper naming convention:
    - [ ] `getParentProduct`,
    - [ ] `getCurrentProduct`
    - [ ] `getCurrentProductConfiguration`
    - [ ] `getOriginalProduct`
    - [ ] `getCurrentProductOptions`
    - [ ] `getCurrentProductGallery`
    - [ ] `getCurrenntProductBreadcrumbs`
    - ...
- Refactor Product.vue + remove the `core/pages/Product.js` dependency
    - [ ] map getters with the right getters- https://github.com/DivanteLtd/vue-storefront/blob/70543b3cfa379971220f6514348fe0386a108878/core/pages/Product.js#L25
    - [ ] `attributesByCode` and `attributesByid` seems to not being used anymore and can be removed
    - [ ] shall we have `this.product` or `this.getProduct` (as getter suggest)?
    - [ ] `productName`, `productId` - should be probable replaced with just `product.name` and `productid` in the theme
    - [ ] `customAttributes` ->replaced with vex getter `getCurrentProductCustomAttributes` or just `getCustomAttributes` mapped to `custom_attributes` (as it’s related to the page context only)
    - [ ] Wishlist and Compare should be moved to smart components - the one’s that uses the Vue to perform the logic, but product page shouldn’t be aware of them at all
    - [ ] `isOptionAvailable` could be moved to helper (called just from theme) or can stay here
    - [ ] `onAfterCustomOptionsChanged`, `onAfterBundleOptionsChanged` logic - to be moved to Vuex handler
    - [ ] `onStateCheck` to be removed - the parent check moved to the Vuex
    - [ ] check when `onAfterPriceUpdate` is called (tier prices probably) and check if we can remove it (probably we can move it to vuex + modify tier prices to call this vex action instead)
    - [ ] Not sure if we can move `onAfterFilterChanged` to vex as it replaces parts of url but probably we can
    - [ ] Remove `onUserPricesRefreshed` - I mean call `asyncData` or other vex method instead
- Refactor the `product/actions` and `product/helpers`
    - [ ] replace `state` read operations with getters
    - [ ] replace `state` write operations with mutations
    - [ ] Extract network call from `syncPlatformPricesOver` to helper and then data resolver - remove the `callback_event` (it’s not used)
    - [ ] In the new module `product-next` remove the `setCurrentProduct` and `selectDefaultVariant` from `product-next/single` or its equivalent - the logic of `product-single` should be shared between old and new product modules helper or import; old action should emulate these parameters new one - should just return product data (setting current product should be in different vex action - related to the product page only)
    - [ ] `product/list` should not `updateState` - i mean the new version should not update the state, just return the final data + do caching the `category-next` current call to `product/list` should be replaced with the new method
    - [ ] `related`, `setCurrentOption`, `setCustomOptions`  and `setCurrentErrors` are not used and can be removed
    - [ ] `fetch` and `fetchAsync` should be merged into one call `loadProduct` or `fetchProduct` that will be used in the `Product.vue`

patzick:
### Related issues

closes #3093

This is a plan for refactor product page with full backward compatibility, things to change:

#### Product module Vuex
- [x] rename and refactor getters
- [x] mutations refactor
- [x] use breadcrumbs from category-next, probably leave old breadcrumbs for compatibility, or remove them if it's not an issue
- [x] add simplified versions of actions for easier cleaned Product.vue page design

#### core/pages/Product.js
We want to get rid of dependencies to this file. So we need to extract logic from it

- [ ] move EventBus listeners to special Mixin, which can be removed with removing EventBus
- [ ] move logic from EventBus listener actions to helpers, there are couple of them, so this is one of the most important and consuming elements

#### Product.vue
- [x] remove Product.js usage
- [x] use simplified actions in asyncData
- [x] use proper refactored getters
- [x] add EventBus listeners mixin
- [x] move filters to separate component


core/modules/catalog/store/product/actions.ts
```diff
@@ -686,6 +686,7 @@ const actions: ActionTree<ProductState, RootState> = {
   /**
    * Load the product data - async version for asyncData()
    */
```
pkarw: Plus - we need to simplify the asyncData/validateRoute calls to this method to simplify it as it was done on the category page 

If we add references to category-next they could be done just from Product.vue in case of somebody using Product.js (backward compatibility mode) to not add any reference to category-next in this case

I like this approach, let’s go this way.

Do you plan to add some simplified / refactored version of product/list (maybe decoupled into set of 2-3 helpers) and product/single by this same refactoring effort? Or just do this as a next step? (This staging approach makes perfect sense to me)

filrak:
Please keep in mind that vuex actions are also extension points therefore they should be easy to swap and granular without much side-effects

pkarw:
@patzick is this ready to be merged in?

core/modules/catalog/store/product/actions.ts
```diff
@@ -585,76 +585,58 @@ const actions: ActionTree<ProductState, RootState> = {
...
+      filterValues: config.entities.product.useDynamicAttributeLoader ? productFields : null,
+      only_visible: config.entities.product.useDynamicAttributeLoader === true,
+      only_user_defined: true,
+      includeFields: config.entities.optimize ? config.entities.attribute.includeFields : null
+    }, { root: true })
+    await Promise.all([
+      dispatch('setupBreadcrumbs', { product: product }),
+      dispatch('filterUnavailableVariants', { product: product }),
+      dispatch('setProductGallery', { product: product })
+    ])
```
pkarw: By putting await on attributePromise and then adding `setProductGallery`  and `filterUnavailableAttributes` to Promise.all we're slowering down the CSR significantly as each of these calls executes network call. It was intentionally splited by this if condition (`if(isServer)`)

pkarw:
What's left in order to finish this refactor @patzick?

patzick:
@pkarw it's ready for review. there is no more Product.js in usage, mutations, getters and some actions are better organized. For simplified methods we'll wait specifically for new product page version to see if we need something more there

core/modules/catalog/index.ts
```diff
@@ -17,4 +22,113 @@ export const CatalogModule: StorefrontModule = function (app, store, router, mod
...
+        },
+        skipCache: true
+      })
+    }
+  }
+
+  if (config.usePriceTiers || isUserGroupedTaxActive()) {
+    EventBus.$on('user-after-loggedin', onUserPricesRefreshed)
+    EventBus.$on('user-after-logout', onUserPricesRefreshed)
+  }
```
pkarw: The only suggestion to changes in this file is that maybe we could split these event handlers into separate helpers so they'd be easier to test / re-use?

core/modules/catalog/store/product/actions.ts
```diff
@@ -664,54 +646,27 @@ const actions: ActionTree<ProductState, RootState> = {
...
-    return context.state.productLoadPromise
+  async fetchAsync (context, { parentSku, childSku = null, route = null }) {
+    Logger.info('Fetching product data asynchronously', 'product', {parentSku, childSku})()
+    EventBus.$emit('product-before-load', { store: rootStore, route: route })
+    await context.dispatch('reset')
+    const product = await context.dispatch('fetch', { parentSku: parentSku, childSku: childSku })
+    await EventBus.$emitFilter('product-after-load', { store: rootStore, route: route })
+    return product
   }
 }
```
pkarw: do we still need `fetchAsync` if it got so much simpler? I think it's pretty much misleading with `fetch` and should be reduced to single method

src/themes/default/pages/Product.vue
```diff
@@ -260,85 +266,84 @@ export default {
...
-          const availableOptions = filterVariants.filter(option =>
-            this.isOptionAvailable(option)
-          )
-          filtersMap[type] = availableOptions
-        })
+    offlineImage () {
+      return {
+        src: this.getThumbnail(this.getCurrentProduct.image, config.products.thumbnails.width, config.products.thumbnails.height),
+        error: this.getThumbnail(this.getCurrentProduct.image, config.products.thumbnails.width, config.products.thumbnails.height),
+        loading: this.getThumbnail(this.getCurrentProduct.image, config.products.thumbnails.width, config.products.thumbnails.height)
```
pkarw: shouldn't it be like `getOfflineImage`?

src/themes/default/pages/Product.vue
```diff
@@ -260,85 +266,84 @@ export default {
...
       }
-      return filtersMap
+    },
+    customAttributes () {
```
pkarw: shouldn't it be like `getCustomAttributes`?

src/themes/default/pages/Product.vue
```diff
@@ -260,85 +266,84 @@ export default {
...
+      return this.$i18n.t('Quantity')
     }
   },
   created () {
     this.getQuantity()
   },
+  mounted () {
+    this.$store.dispatch('recently-viewed/addItem', this.getCurrentProduct)
+  },
+  async asyncData ({ store, route }) {
```
pkarw: Please make sure it works well with navigating thru related products (so on the same `Product` component - just switching the router parameters; I'd encountered a lot of issues with this behavior when working on url dispatcher0

core/modules/catalog/index.ts
patzick: i think this should be next step - get all there Listeners in this file, then find when events are emitted and change them to actions to to this job. WDYT?

core/modules/catalog/store/product/actions.ts
patzick: good point, will change that!

src/themes/default/pages/Product.vue
patzick: i think it shouldn't. It's theme-specific and doing it this way we're allowing users to easily modify or even remove it.

patzick: oh, you mean computed name?
i agree if so ;D

core/modules/catalog-next/store/category/actions.ts
```diff
@@ -140,6 +141,13 @@ const actions: ActionTree<CategoryState, RootState> = {
     const categoryHierarchyIds = _prepareCategoryPathIds(category) // getters.getCategoriesHierarchyMap.find(categoryMapping => categoryMapping.includes(category.id))
     const categoryFilters = { 'id': categoryHierarchyIds }
     await dispatch('loadCategories', {filters: categoryFilters})
+  },
+  async loadProductCategories ({ dispatch }, product: Product) {
+    let categoryIds = []
+    if (!product) return categoryIds
+    if (product && product.category_ids) categoryIds = product.category_ids
+    const categoryFilters = { 'id': categoryIds }
+    return dispatch('loadCategories', {filters: categoryFilters})
```
andrzejewsky: I think it's same as:
```javascript
 return dispatch('loadCategories', { filters: product.category_ids })
}

return []
```

core/modules/catalog/store/product/actions.ts
```diff
@@ -534,29 +534,29 @@ const actions: ActionTree<ProductState, RootState> = {
...
     if (productOption && typeof productOption === 'object') { // TODO: this causes some kind of recurrency error
-      context.commit(types.CATALOG_SET_PRODUCT_CURRENT, Object.assign({}, context.state.current, { product_option: productOption }))
+      context.commit(types.PRODUCT_SET_CURRENT, Object.assign({}, context.state.current, { product_option: productOption }))
     }
   },
 
   setCurrentErrors (context, errors) {
     if (errors && typeof errors === 'object') {
-      context.commit(types.CATALOG_SET_PRODUCT_CURRENT, Object.assign({}, context.state.current, { errors: errors }))
+      context.commit(types.PRODUCT_SET_CURRENT, Object.assign({}, context.state.current, { errors: errors }))
```
andrzejewsky: You have created getter:
```javascript
getters.getCurrentProduct
```

core/modules/catalog/index.ts
pkarw: Yeah, sounds OK but at this point, we can just move the functions (handlers) into separate helpers - that's all, to keep this index.ts smaller and more readable

src/themes/default/pages/Product.vue
pkarw: yes, computer property name

pkarw:
@patzick i just wanted to check the status of this one?:)

core/modules/catalog/store/product/actions.ts
```diff
@@ -282,21 +282,21 @@ const actions: ActionTree<ProductState, RootState> = {
...
-    const products = await context.dispatch('findProducts', { query, start, size, entityType, sort, cacheByKey, excludeFields, includeFields, configuration, populateRequestCacheTags })
-
-    await context.dispatch('preConfigureAssociated', { products, prefetchGroupProducts })
+  async list ({ dispatch, commit }, { query, start = 0, size = 50, entityType = 'product', sort = '', cacheByKey = 'sku', prefetchGroupProducts = !isServer, updateState = false, meta = {}, excludeFields = null, includeFields = null, configuration = null, append = false, populateRequestCacheTags = true }) {
+    const searchResult = await dispatch('findProducts', { query, start, size, entityType, sort, cacheByKey, excludeFields, includeFields, configuration, populateRequestCacheTags })
+    await dispatch('preConfigureAssociated', { searchResult, prefetchGroupProducts })
 
     if (updateState) {
-      context.commit(types.CATALOG_UPD_PRODUCTS, { products, append: append })
+      if (append) commit(types.PRODUCT_ADD_PAGED_PRODUCTS, searchResult)
```
defudef: What do you think about using ternary operator?
```TS
if (updateState) {
  commit(append ? types.PRODUCT_ADD_PAGED_PRODUCTS : types.PRODUCT_SET_PAGED_PRODUCTS, searchResult);
}
```

Or even one liner:
```TS
updateState && commit(append ? types.PRODUCT_ADD_PAGED_PRODUCTS : types.PRODUCT_SET_PAGED_PRODUCTS, searchResult);
```

core/modules/catalog/store/product/actions.ts
```diff
@@ -282,21 +282,21 @@ const actions: ActionTree<ProductState, RootState> = {
...
+      else commit(types.PRODUCT_SET_PAGED_PRODUCTS, searchResult)
     }
 
-    EventBus.$emit('product-after-list', { query: query, start: start, size: size, sort: sort, entityType: entityType, meta: meta, result: products })
+    EventBus.$emit('product-after-list', { query: query, start: start, size: size, sort: sort, entityType: entityType, meta: meta, result: searchResult })
```
defudef: You can use the power of ES6 and simplify it a little :)
```TS
EventBus.$emit('product-after-lit', { query, start, size, sort, entityType, meta, result: searchResult })
```

core/modules/catalog/store/product/actions.ts
```diff
@@ -583,76 +583,58 @@ const actions: ActionTree<ProductState, RootState> = {
...
-
-        context.dispatch('setProductGallery', { product: product })
-
-        if (config.products.preventConfigurableChildrenDirectAccess) {
-          subloaders.push(context.dispatch('checkConfigurableParent', { product: product }))
-        }
-      } else { // error or redirect
+    const product = await dispatch('single', { options: productSingleOptions })
+    if (product.status >= 2) {
+      throw new Error(`Product query returned empty result product status = ${product.status}`)
```
defudef: I have a proposition in the future. What do you guys think of creating separate Error classes and define a structure for them, so the code could be shorter and more readable I think?

For example:
```TS
if (product.status >= 2) {
  throw new ProductQueryEmptyStatusError(product.status); // which will produce "Product query returned (...)"
}
if (product.visibility === 1) {
  throw new ProductQueryEmptyVisibilityError(product.visibility);
}
```

core/modules/catalog/store/product/actions.ts
```diff
@@ -583,76 +583,58 @@ const actions: ActionTree<ProductState, RootState> = {
...
-        } */
-
-        context.dispatch('setProductGallery', { product: product })
-
-        if (config.products.preventConfigurableChildrenDirectAccess) {
-          subloaders.push(context.dispatch('checkConfigurableParent', { product: product }))
-        }
-      } else { // error or redirect
+    const product = await dispatch('single', { options: productSingleOptions })
+    if (product.status >= 2) {
```
defudef: Would be great if you could create a constant for this value "2"

core/modules/catalog/events.ts
```diff
@@ -0,0 +1,109 @@
+import EventBus from '@vue-storefront/core/compatibility/plugins/event-bus'
```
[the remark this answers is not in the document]
patzick: we want to get rid of it and make hooks, but this is outside of this scope. The goal was just not to use Product.js anymore.

core/modules/catalog/events.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (store.getters['product/getCurrentProduct'] && product.sku === store.getters['product/getCurrentProduct'].sku) {
+    // join selected variant object to the store
+    await store.dispatch('product/setCurrent', omit(product, ['name']))
+  }
+}
+
+export const filterChangedProduct = async (filterOption, store, router) => {
+  EventBus.$emit('product-before-configure', { filterOption: filterOption, configuration: store.getters['product/getCurrentProductConfiguration'] })
+  const prevOption = store.getters['product/getCurrentProductConfiguration'][filterOption.attribute_code]
+  let changedConfig = Object.assign({}, store.getters['product/getCurrentProductConfiguration'], {[filterOption.attribute_code]: filterOption})
```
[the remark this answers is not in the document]
patzick: I didn't refactor these methods, just moved from product.js, will be done on another stage with changing to hooks.

core/modules/catalog/events.ts
```diff
@@ -0,0 +1,109 @@
...
+        {
+          [filterOption.attribute_code]: prevOption
+        }
+      ))
+    } else {
+      store.commit(PRODUCT_SET_CURRENT_CONFIGURATION, Object.assign(
+        {},
+        store.getters['product/getCurrentProductConfiguration'],
+        {
+          [filterOption.attribute_code]: undefined
```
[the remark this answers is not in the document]
patzick: as above

core/modules/catalog/events.ts
```diff
@@ -0,0 +1,109 @@
...
+        price: priceDelta,
+        price_incl_tax: priceDeltaInclTax
+      }
+    ))
+  }
+}
+
+export const onUserPricesRefreshed = async (store, router) => {
+  if (router.currentRoute.params.parentSku) {
+    await store.dispatch('product/reset')
```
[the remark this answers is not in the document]
patzick: disagree, especially in the case of readability. We're not aiming in older node versions.
We have this discussion here: https://github.com/DivanteLtd/vue-storefront/pull/3512#discussion_r323681368 let's keep it in one place

core/modules/catalog/index.ts
```diff
@@ -17,4 +20,15 @@ export const CatalogModule: StorefrontModule = function (app, store, router, mod
   store.registerModule('stock', stockModule)
   store.registerModule('tax', taxModule)
   store.registerModule('category', categoryModule)
+
+  // Things moved from Product.js
+  EventBus.$on('product-after-priceupdate', product => productAfterPriceupdate(product, store))
```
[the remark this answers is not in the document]
pkarw: @lukeromanowicz good point! we need to have them `$off` somewhere OR just bind in the CSR. Going this way - these events are used because of how the theme works. It would be probably better to move them out from Vuex from the architecture point of view I guess. I'm not sure about this second statement as they're in the end not touching UI directly (but working on data models; therefore it was fine with me to keep it here) but for sure we need to take care of `off`

core/modules/catalog/store/product/mutations.ts
```diff
@@ -1,64 +1,121 @@
...
-    } else {
-      (state.list as PagedProductList).start = products.start as number
-      (state.list as PagedProductList).perPage = products.perPage as number
-      (state.list as PagedProductList).items = (state.list as PagedProductList).items.concat(products.items)
-    }
+    console.error('Deprecated mutation CATALOG_UPD_PRODUCTS - use PRODUCT_SET_PAGED_PRODUCTS or PRODUCT_ADD_PAGED_PRODUCTS instead')
   },
   [types.CATALOG_SET_PRODUCT_CURRENT] (state, product) {
-    state.current = product
+    console.error('Deprecated mutation CATALOG_SET_PRODUCT_CURRENT - use PRODUCT_SET_CURRENT instead')
```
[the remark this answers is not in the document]
pkarw: I agree with @lukeromanowicz  - error + backward compatibility will be more apprecciated

core/modules/catalog/index.ts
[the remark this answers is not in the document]
patzick: good point - on client-side it's invoked once, i'll change this not to be invoked on Server side :)

alinadivante:
@patzick 
I had problems during tests, when I had the **test** and **demo** endpoints set.

→  "api": {    "url": "https://test.storefrontcloud.io"
[ ] I received `An error occurred validating the nonce` message two times, but I have no idea why and I can't find scenario to reproduce..
[ ] problem with opening PDP from Home page, wishlist, compare list
[ ] `HTTP ERROR undefined` message occurs all the time.. when I'm trying open some pages
[ ] This product SKU: WH04-M-ORANGE has 95 available items http://localhost:3000/women/tops-women/hoodies-and-sweatshirts-women/hoodies-and-sweatshirts-24/miko-pullover-hoodie-1097.html?childSku=WH04-M-Orange but after adding to the cart there is `Out of the stock!` info 
[ ] There is no breadcrumbs 


→  "api": {    "url": "https://demo.storefrontcloud.io"
[ ] There is no breadcrumbs 
[ ] Problem with unavailable variants. *SOMETIMES* I can't click on specific color/size.
Scenario: add T**iffany Fitness Tee** product from Home page to wishlist or compare list. Click link from wishlist/compare list (there is also wrong photo displayed) and on PDP try to choose variant Blue/XS
[ ] There are no prices

pkarw:
@patzick please take care of the fixes some of them seems like a critical one's

src/themes/default/pages/Product.vue
```diff
@@ -273,79 +278,86 @@ export default {
...
+    const product = await store.dispatch('product/loadProduct', { parentSku: route.params.parentSku, childSku: route && route.params && route.params.childSku ? route.params.childSku : null })
+    // setup breadcrumbs
+    const productCategories = await store.dispatch('category-next/loadProductCategories', product)
+    const category = store.getters['category-next/getCategoryFrom'](route.path)
+    await store.dispatch('category-next/loadCategoryBreadcrumbs', category)
```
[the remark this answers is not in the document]
patzick: asyncData is not returning anything, shouldn't category also. Should just finish after it's loaded :)

patzick:
@alinadivante i've checked points, which has been corrected. These which are not marked I couldn't reproduce. Could be caused by develop branch problems. Please do re-tests and @pkarw please re-review :)

patzick:
@alinadivante please add comments into #3093 if something would be not okay with PDP

PR #3209, opened 2019-07-10
Get also none product image thumbnails via API

cewald:
### Related issues
closes #3207

### Short description and why it's useful

Adds the possibility to get also none product image thumbnails via API using a special parameter in `getThumbnail()` or using `getMediaThumbnail()`.

It's enabled by default, which I'm not sure about because of the needed change of the API image-url. But it is backwards compatible if you disable it in configuration.

I also added some types to the `getThumbnailPath()` method parameters in `core/helpers/index.ts`.

### Which environment this relates to
Check your case. In case of any doubts please read about [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html)

- [x] Test version (https://test.storefrontcloud.io) - this is a new feature or improvement for Vue Storefront. I've created branch from `develop` branch and want to merge it back to `develop`
- [ ] RC version (https://next.storefrontcloud.io) - this is a stabilisation fix for Release Candidate of Vue Storefront. I've created branch from `release` branch and want to merge it back to `release`
- [ ] Stable version (https://demo.storefrontcloud.io) - this is an important fix for current stable version. I've created branch from `hotfix` or `master` branch and want to merge it back to `hotfix`

### Upgrade Notes and Changelog

- [x] No upgrade steps required (100% backward compatibility and no breaking changes)
- [x] I've updated the [Upgrade notes](https://github.com/DivanteLtd/vue-storefront/blob/develop/doc/Upgrade%20notes.md) and [Changelog](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md) on how to port existing VS sites with this new feature

**IMPORTANT NOTICE** - Remember to update `CHANGELOG.md` with description of your change

### Contribution and currently important rules acceptance

- [x] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)
- [x] I read the [TypeScript Action Plan](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/TypeScript%20Action%20Plan.md) and adjusted my PR according to it
- [x] I read about [Vue Storefront Modules](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md) and I am aware that every new feature should be a module

PR #3224, opened 2019-07-12
Changed getThumbnailPath to accept the image type

ResuBaka:
### Related PRs
Vue-storefront API [PR 299](https://github.com/DivanteLtd/vue-storefront-api/pull/299) 

This is usefull when you want to load the category images with the thumbnail helper.

### DIFF
--- a/core/helpers/index.ts
+++ b/core/helpers/index.ts
@@ -31,9 +31,10 @@ export function slugify (text) {
  * @param relativeUrl
  * @param width
  * @param height
+ * @param type
  * @returns {*}
  */
-export function getThumbnailPath (relativeUrl, width, height) {
+export function getThumbnailPath (relativeUrl, width, height, type = 'product') {
   if (config.images.useExactUrlsNoProxy) {
     return relativeUrl // this is exact url mode
   } else {
@@ -43,11 +44,12 @@ export function getThumbnailPath (relativeUrl, width, height) {
     let baseUrl = processURLAddress(config.images.proxyUrl ? config.images.proxyUrl : config.images.baseUrl)
     if (baseUrl.indexOf('{{') >= 0) {
       baseUrl = baseUrl.replace('{{url}}', relativeUrl)
+      baseUrl = baseUrl.replace('{{type}}', type)
       baseUrl = baseUrl.replace('{{width}}', width)
       baseUrl = baseUrl.replace('{{height}}', height)
       resultUrl = baseUrl
     } else {
-      resultUrl = `${baseUrl}${parseInt(width)}/${parseInt(height)}/resize${relativeUrl}`
+      resultUrl = `${baseUrl}${type}/${parseInt(width)}/${parseInt(height)}/resize${relativeUrl}`
     }
     return relativeUrl && relativeUrl.indexOf('no_selection') < 0 ? resultUrl : config.images.productPlaceholder || ''
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Default storeview settings are now overridden by specific storeview settings - @lukeromanowicz (#3057)
 - Apache2 proxy header support for store based on host - @resubaka (#3143)
 - Items count badges for Compare products and wishlist icons at header - @vishal-7037 (#3047)
+- Changed the thumbnail helper so you can load category images with it - @resubaka (#3224)
 
 ### Fixed
 - Fixed product link in wishlist and microcart - @michasik (#2987)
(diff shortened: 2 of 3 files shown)


pkarw:
Hi @ResuBaka this is pretty cool feature but we need to make it backward compatible

core/helpers/index.ts
```diff
@@ -43,11 +44,12 @@ export function getThumbnailPath (relativeUrl, width, height) {
     let baseUrl = processURLAddress(config.images.proxyUrl ? config.images.proxyUrl : config.images.baseUrl)
     if (baseUrl.indexOf('{{') >= 0) {
       baseUrl = baseUrl.replace('{{url}}', relativeUrl)
+      baseUrl = baseUrl.replace('{{type}}', type)
       baseUrl = baseUrl.replace('{{width}}', width)
       baseUrl = baseUrl.replace('{{height}}', height)
       resultUrl = baseUrl
     } else {
-      resultUrl = `${baseUrl}${parseInt(width)}/${parseInt(height)}/resize${relativeUrl}`
+      resultUrl = `${baseUrl}${type}/${parseInt(width)}/${parseInt(height)}/resize${relativeUrl}`
```
pkarw: This change assumes that you've got changed the `vs-api` config:  `imgUrl` (by removing `product` from it). it's not backward compatible change. We must probably have a `type=null` by default to support the old paths - the guys who have installed VS before this changes and have the path with `product` inside the `local.json`

ResuBaka: But here we could just check if the URL in the API has /product at the end when we load the config and remove it so we have the right URL.

Would that fix it?

pkarw: yes, probably it's a fix but it's very well specific one :) I mean - if somebody has been using differnt url strategy won't probably have `product` in there - so maybe chaning the deafult value is more general solution? No strong prefernce however.

ResuBaka: To fix it easily i would say i add a new config value to the config.images with which we can use the new logic. Which would fix the problem. 

So it would be the old way when the config is false and the new way when it is true. So the user needs to check if it is going to break stuff for him or not.

This would be a good way here and in the API.

ResuBaka:
@pkarw  So as I have wrote in the PR for the API should I change the logic a bit more so we can load all images under /media?

And then mark the old way as deprecated and remove it with vue-storefront 2?

core/helpers/index.ts
pkarw: Yes, that makes sense for me

pkarw:
By the way this seems like a duplicate / different approach to https://github.com/DivanteLtd/vue-storefront/pull/3209/files

In 3209 basically there is the same approach applied (with Boolean flag for backward compatibility).

cewald:
@ResuBaka I had the same issue in the Ticket above #3209. In my solution the backwards compatibility is considered an can be just switched in 2.0. And you dont have to change the API code. I think the api shouldnt care from which folder the image is of to make the VSF more flexible in its configuratio . Have a look at my PR #3209 and its related issue.

ResuBaka:
@cewald Your PR looks like it fixes the problem we had and fixes it. But you still need to change the default config in the API to reflect your changes and that breaks the backwards compatibility. 

When you have the url example.org/api/img/300/300/resize/catalog/product/p/l/niceproductimage.jpg

The API with default config would call http://demo-magento2.vuestorefront.io/media/catalog/product/catalog/product/p/l/niceproductimage.jpg

So it is not a 100% backwards compatibility out of the box.

So here we need to know which way we go. Mine would be backwards compatibility when i check for the type of the image and have like a new Config value like the same as the PR from @cewald in the API where we check it. 

But when i look at the way @cewald  has implemented it looks easier to use with manuell build image URLs from like CMS. So i would prefer with the code from @cewald with some miner changes, but we still need to make one change at least in the API which is changing the config value for the base image URL.

ResuBaka:
Or on a second thought shouldn't we move the changes from the config from @cewald  PR in to the API and use the type parameter because than we could load images from different image domains?

Which would add even more functionality to this feature.

cewald:
It is backward compatible as if you don't enable it via `useSpecificImagePaths` by default. So you just have to update the API url if you like to use this feature. This could later might be forced in version 2.0 when you enable it by default and change the default API-Url without `/catalog/product`.

> Or on a second thought shouldn't we move the changes from the config from @cewald PR in to the API and use the type parameter because than we could load images from different image domains?
Yes indeed, it'd be a nice feature with multi domain support. But this would complicate the yet so nice and simple API code – if you use a different domain for a frontend you would probably also use a different theme and VSF instance.

@ResuBaka Thanks, I will have a look at your proposed changes.

patzick:
@ResuBaka should we close it in favour of #3209 ?

ResuBaka:
@patzick yea im am closing mine.

PR #3257, opened 2019-07-19
Fix typo in product store doc

tp:
Also switched from `a` to `the` to be more consistent with the previous examples.
(Also seem more appropriate here as the document is referring to specific instances.)


tp:
According to the contributing guidelines this should go to `develop` though. Does that also apply for the docs fix? Can retarget then.

PR #3271, opened 2019-07-23
Fix auto-login after registering user, fix refreshing token

### LINKED ISSUE #3095: Refactor the `user` vuex module

In order to prepare VS for 2.0 release we're refactoring some vuex modules in order to keep the good practices:
- remove state write operations from the actions,
- replace state read operations with vuex getters,
- preferably migrate from events-based logic to vuex actions,
- remove notifications from vuex (`spawnNotification`),
- remove `rootStore` references,
- extract network calls to `data-resolveres` (as introduced in the #2890)

- [ ] `resetPassword`, `login`, `register`, `update`, `me`, `getOrdersHistory` network calls refactoed to helpers + then to `data-resolvers`
- [ ]  remove `callback_event` event's based logic - use just the `then(helperFnc)` instead
- [ ]  try to remove `spawnNotification` move the notifications one level higher (to the UI)
- [ ] replace all `rootStore` calls with the `context` instead (for example `context.rootState`)
- [ ] remove the calls to `history` and `me` from `login` (it should be extremly simple) - the `sessionAfterAuthorized` should be called instead - by some  observer (plugin?) on the user token
- [ ] replace all `fetch` calls with Task api calls

- [ ] YES
- [ ] NO

Pick one option.

andrzejewsky:
### Related issues

closes #3095

It fixes an issue with auto-login after registering user and refreshing the token.

### DIFF
--- a/core/data-resolver/UserService.ts
+++ b/core/data-resolver/UserService.ts
@@ -84,16 +84,14 @@ const changePassword = async (passwordData: DataResolver.PasswordData): Promise<
     }
   })
 
-const invalidateToken = async (refreshToken: string): Promise<Task> =>
-  TaskQueue.execute({
-    url: processLocalizedURLAddress(config.users.refresh_endpoint),
-    payload: {
-      method: 'POST',
-      mode: 'cors',
-      headers,
-      body: JSON.stringify({ refreshToken })
-    }
-  })
+const refreshToken = async (refreshToken: string): Promise<string> =>
+  fetch(processLocalizedURLAddress(config.users.refresh_endpoint), {
+    method: 'POST',
+    mode: 'cors',
+    headers,
+    body: JSON.stringify({ refreshToken })
+  }).then(resp => resp.json())
+    .then(resp => resp.result)
 
 export const UserService: DataResolver.UserService = {
   resetPassword,
@@ -103,5 +101,5 @@ export const UserService: DataResolver.UserService = {
   getProfile,
   getOrdersHistory,
   changePassword,
-  invalidateToken
+  refreshToken
 }
--- a/core/modules/user/store/actions.ts
+++ b/core/modules/user/store/actions.ts
@@ -74,16 +74,13 @@ const actions: ActionTree<UserState, RootState> = {
   async refresh ({ commit }) {
     const usersCollection = StorageManager.get('user')
     const refreshToken = await usersCollection.getItem('current-refresh-token')
-    const resp = await UserService.invalidateToken(refreshToken)
+    const newToken = await UserService.refreshToken(refreshToken)
 
-    if (resp.code === 200) {
-      commit(types.USER_TOKEN_CHANGED, {
-        newToken: resp.result,
-        meta: resp.meta ? resp.meta : null
-      }) // TODO: handle the "Refresh-token" header
+    if (newToken) {
+      commit(types.USER_TOKEN_CHANGED, { newToken })
     }
 
-    return resp
+    return newToken
   },
   /**
    * Update user groupToken and groupId in state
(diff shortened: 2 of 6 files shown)


andrzejewsky:
@pkarw Unfortunately I had to use `fetch` instead of `TaskQueue` because it causes and infinity loop with refreshing the token. I've tried to add some `if` to cover the case where there is a `refresh` task but it took too much time with no result. I think for now `fetch` is enough, but one day, would be nice to refactor these piece of code responsible for `Task/TaskQueue`.

core/data-resolver/UserService.ts
```diff
@@ -84,16 +84,13 @@ const changePassword = async (passwordData: DataResolver.PasswordData): Promise<
...
-  TaskQueue.execute({
-    url: processLocalizedURLAddress(config.users.refresh_endpoint),
-    payload: {
-      method: 'POST',
-      mode: 'cors',
-      headers,
-      body: JSON.stringify({ refreshToken })
-    }
-  })
+const invalidateToken = async (refreshToken: string): Promise<any> =>
```
patzick: it should not be any
it returns object like that:
```json
{"code":200,"result":"mwgdkaspawubhrmyc64oq4qj72ng9yyr","meta":{"refreshToken":"6474ccd3931747fd7bbf08d27a4f15bcc8dec730acee04882b6a046d7a768a165e55e2949d739adb6b103c8b594c543a977342b13768a2f74d2a8baa896d3a0d795862c02f7d35c3bfaa852ae8718330ce9c0c71af859c11cf354a408a19b7fd37ce0a434b61dc7e20c4dbf895a72721a7b883ed126cd3826c92adef00313d2a5680d5b19b5e4ae28f5b21f2f75484c761fed56c741af1f3976efb3f6330462d96c472c4812ac494"}}
```
so we should return just string with resp.result. Also this endpoint is not invalidatingToken - it refreshes it, so creating new current-token for user session, so it would be nice to change this method this way :)

andrzejewsky: Yeah, sure, but there is also `refreshToken` in that response which is used in `USER_TOKEN_CHANGED` mutation. So I can just create a type/interface for that.

patzick: Actually no, refresh token is not changing, passed only by meta data which we can ignore here. We're giving it as a post params to get ner current-token. I really think that just returning string with new current-token and renaming method is a proper way to go here :)

andrzejewsky: @patzick I've just tested it - doesn't work, I can't even login without that. Basically this is the only place where `refreshToken` is set

andrzejewsky: @patzick Updated. You are right, I've noticed that `refreshToken` during the login & it works fine. Thanks!

core/modules/user/store/actions.ts
```diff
@@ -74,16 +74,13 @@ const actions: ActionTree<UserState, RootState> = {
...
     const refreshToken = await usersCollection.getItem('current-refresh-token')
-    const resp = await UserService.invalidateToken(refreshToken)
+    const newToken = await UserService.refreshToken(refreshToken)
 
-    if (resp.code === 200) {
-      commit(types.USER_TOKEN_CHANGED, {
-        newToken: resp.result,
-        meta: resp.meta ? resp.meta : null
-      }) // TODO: handle the "Refresh-token" header
+    if (newToken) {
```
patzick: how much better is THAT! <3

youanden: @andrzejewsky @patzick So there's a bug here. If a user's password changes on another browser, device, w/e., the refresh token value will fail to login. In such a case, Magento returns:
```
{
  "code": 401,
  "result": "The account sign-in was incorrect or your account is disabled temporarily. Please wait and try again later."
}
```
As per line 94 of core/data-resolver/UserService.ts, the response.result is returned into the newToken variable and commits the value "The account sign-in was incorrect or your account is disabled temporarily. Please wait and try again later." into the user's current-token variable. 

The correct behavior would be to check that magento's returned code 200 on line 94 and return the result, otherwise null. Of course maybe for the user it would be nice to show an error like "Authorization expired. Please login again."

PR #3295, opened 2019-07-26
#3286 + #3332 `catalog-next` module improvements

### LINKED ISSUE #3286: New getter `category-next/getCategoryFrom` and `category-next/getCurrentCategory` always returns the root category
## Current behavior

The new getter `category-next/getCategoryFrom` (and `category-next/getCurrentCategory`) on the new category page always returns the root category because of a wrong filter in its getter.

This happens if there is a category with a path like `/` or a empty one in the `category-next/categoriesMap` state.

E.g. if you are in `/boys.html` and the it crawls to the states it tests like:
```javascript
return getters.getCategories.find(category => path.includes(category.url_path)) || {}
```
If there is a category with path `/` (or empty) in the state, it will always return this one. This filter is quiet flacky. I would prefer something more concrete like this:
```javascript
return getters.getCategories.find(category => path.replace(/^(\/)/gm, '') === category.url_path) || {}
```

Also what happens if you use a different url path, then it would never find the right current category, e.g. I added an extra mapper fallback to let the `/new.html` category availaible ever `/new` this won't over work with the new getters based on route path. You will need at least another getter for this at least.

I think this new functionality leads a bit in the wrong direction and needs way more flexibility.

## Expected behavior

Get the right category from state.

See above.

## Repository

develop

- [x] YES
- [ ] NO

Pick one option.

- [x] This is a bug report for test version on https://test.storefrontcloud.io - In this case Developer should create branch from `develop` branch and create Pull Request `2. Feature / Improvement` back to `develop`.

## Environment details

- Browser: All
- OS: All
- Node: All
- Code Version: develop:latest

cewald:
### Related issues

closes #3286 and #3332 

I added a new state item, containing the current category id, to be able to traverse the current category even if the current route path is not the same as in `url_path` nor `slug` (e.g. if you use rewrites). I decided to use just the id in `currentId` state property to prevent duplicate content in the state, because the category is already in `categoriesMap`.

Also I changed how the filter for the category is added to the action in the category page. This way you don't need to ask for `config.products.useMagentoUrlKeys ` and use just the given parameters of the current route. This is more specific and also more flexible than just get it by the path.

And I fixed the bug mentioned in #3286 where it sometimes returns the wrong category because of the flacky `include()`-search in the `getCategoryFrom()` getter.


lukeromanowicz:
Looks good, awesome job :+1:

patzick:
Thanks @cewald, i think it's better to improve category match pattern instead of extending state to store categoryId. Removing it was a purpose to have single source of truth without need to set it maunually.
Let's hold this one to rethink it :)

cewald:
@patzick I totally like this approach of having a "true" single source of truth, thats why I just added the ID instead of a own category object, but yes indeed it would be better to get that without a extra state entry. But how is it then possible to see if this url-key is probably connected to a custom route if I don't use the url-key in category-state?

For example I added a route like via a custom url `mappingFallback()` rule like:
```
{
  "name": "category",
  "params": {
    "slug": "new"
  }
}
```

A use-case for that is to have custom url's for a specific standard route. For example: our marketing want's an url that isn't conform with our default url-category pattern but then again sounds better in communication.

cewald:
@patzick I updated my pull request as you said. I found a way to traverse the current url in a better way without the need of an extra current category state. I was able to traverse the current category in a more specific way than url, using the `route.params` (like in URL mapping). This way it's possible to have any url-key you want and still find the current category by it's original url-key from `route.params`.

I put it into a new method so we still have both ways – there are use-cases when it's enough to look just after the url-key.

pkarw:
@cewald is this one ready to be merged in?

pkarw:
awesome job by the way @cewald  :-)

core/modules/catalog-next/helpers/categoryHelpers.ts
```diff
@@ -14,3 +14,14 @@ export const _prepareCategoryPathIds = (category: Category): string[] => {
   if (!category || !category.path) return []
   return category.path.split('/')
 }
+
+export const getSearchOptionsFromRouteParams = (params: { [key: string]: string } = {}): Object => {
+  const filterableKeys = ['url-key', 'slug', 'id']
```
pkarw: I'd move this one to config - to have the filterable keys editable in the config

cewald: Yes indeed, that'd be more clean.
I've added them to the configs.

pkarw:
Thasnk

core/modules/catalog-next/store/category/mutations.ts
```diff
@@ -12,18 +13,14 @@ const mutations: MutationTree<CategoryState> = {
...
     }
   },
   [types.CATEGORY_ADD_CATEGORIES] (state, categories: Category[] = []) {
     if (categories.length) {
-      let newCategoriesEntry = {}
       categories.forEach(category => {
-        newCategoriesEntry[category.id] = category
+        Vue.set(state.categoriesMap, category.id, category)
       })
-      state.categoriesMap = Object.assign(state.categoriesMap, newCategoriesEntry)
```
patzick: i'll only change that back for one reason - with every category vue will mutate state and will invoke recalculate all computed properties and getters which are based on this. Fix for reactivity here is actually change
`state.categoriesMap = Object.assign(state.categoriesMap, newCategoriesEntry)`
to
`state.categoriesMap = Object.assign({}, state.categoriesMap, newCategoriesEntry)`

cewald: Ah nice, didn't figured that out :)

PR #3313, opened 2019-07-31
[ WIP ] Category

filrak:
Category Page work in progress


src/themes/capybara/components/category/FiltersSidebar.vue
```diff
@@ -0,0 +1,123 @@
+<template>
+  <SfSidebar
+    :visible="visible"
+    @close="onClose"
+    class="filters"
+  >
+    <h3 class="filters__title">
+      Color
+    </h3>
+    <SfFilter
```
pkarw: When user adds additional fields to `config.products.defaultFilters` they should be also rendered using kind of general filter (`GenericSelector` in the current theme) -> https://github.com/DivanteLtd/vue-storefront/blob/e486c368b51fbbc20f29e3cdda68e4c24ef5fe98/src/themes/default/components/core/blocks/Category/Sidebar.vue#L63

src/themes/capybara/components/category/SubCategoriesSidebar.vue
```diff
@@ -0,0 +1,64 @@
...
+    currentCategory: {
+      type: Object,
+      default: () => {}
+    }
+  },
+  methods: {
+    filterCategories (level, parentId) {
+      return this.categories.filter(category => category.level === level && category.parent_id === parentId)
+    },
+    filterCategoryChildren (parentId) {
```
pkarw: Will this work with dynamic category pre-fetching? We don't have full categories tree loaded at once. The sub categories should be loaded once user clicked the parent category to save the `__INITIAL_STATE__` size: it has been introduced with https://github.com/DivanteLtd/vue-storefront/pull/2100

On sites like `wonect.com` the category tree itslelf is 2MB we can't render it into HTML @patzick was dealing with this with the category refactor as well

pkarw: We should load just the root categories at first - display the first level of menu

pkarw:
In general it looks pretty cool!

src/themes/capybara/components/category/FiltersSidebar.vue
filrak: imho this should be part of vuex then to filter it @patzick . The question is do we really need this or want to leave it up to developers for better flexibility

src/themes/capybara/components/category/SubCategoriesSidebar.vue
filrak: @pkarw here (according to designs) categories are only 2 levels deep (based on which category we are because those are subcategories of the current one)

filrak:
Ok one update. Core pages are now in proper modules. In theme we expose them as components with scoped slots. While a was against mixin-based pages this approach is much better because:
- users know exactly what the API is
- no naming cinflicts
- **we can still update those components** and theme logic is purely ui-driven 
- this components can be configured via props so we can still make them customizable while encapsulating business logic inside

patzick:
@filrak is it still WIP? What do we have here? :)

PR #3326, opened 2019-08-02
Feature/3281 state optimization

### LINKED ISSUE #3281: Remove attribute.list from initial state for product page

Currently, the initial state can be really huge mostly because of the __INITIAL_STATE__. Removing the server pre-fetched attributes list seems like a safe bet to improve it (as the attributes will be loaded client's side then anyway)

To comment out this line for tests: https://github.com/DivanteLtd/vue-storefront/blob/d6ace4928f3334739c099f7a481ad80e683175a4/core/modules/catalog/store/product/actions.ts#L627

TODO:
- [ ] figure out the Category initial state if there are any other things which can be safely removed from the initial state (and loaded just client's side)

pkarw:
### Related issues

closes #3281 

I've removed the `attribute.list_by_id` and `attribute.list_by_code` from the `window.__INITIAL_STATE__` which could be even up to 50% of the product page size. It was possible because of [`lazy-hydrate`](https://github.com/maoberlehner/vue-lazy-hydration). The attributes are loaded immediately on the product page (CSR) but then hydrated only when visible

I've fixed also some other minor things like `claims` and `elasticCacheCollection` namings + we started using `localStorage` for attributes which has been populated but ... never read before :)


core/modules/catalog/store/attribute/actions.ts
```diff
@@ -6,20 +6,24 @@ import RootState from '@vue-storefront/core/types/RootState'
...
    * @param {Array} attrCodes attribute codes to load
    */
-  list (context, { filterValues = null, filterField = 'attribute_code', only_user_defined = false, only_visible = false, size = 150, start = 0, includeFields = config.entities.optimize ? config.entities.attribute.includeFields : null }) {
+  async list (context, { filterValues = null, filterField = 'attribute_code', only_user_defined = false, only_visible = false, size = 150, start = 0, includeFields = config.entities.optimize ? config.entities.attribute.includeFields : null }) {
     const commit = context.commit
-
     let searchQuery = new SearchQuery()
     const orgFilterValues = filterValues ? [...filterValues] : []
     if (filterValues) {
+      const cachedAttributes = await prefetchCachedAttributes(config, filterField, filterValues)
```
patzick:
```suggestion
      const cachedAttributes = await prefetchCachedAttributes(filterField, filterValues)
```

PR #3340, opened 2019-08-07
User hooks

filrak:
### Related issues

https://github.com/DivanteLtd/vue-storefront/issues/3205

Added hooks:
- `beforeUserAuthorized(user: { username, password })` (mutator) - Tbh I'm not sure if this should be a mutator because of security reasons. I can't find any valid use case. @pkarw you suggested it as a mutator, is it ok to change it to listener?
- `afterUserAuthorized(reponse)`(listener) 
- `afterUserUnauthorized`(listener)


pkarw:
the idea of authorization hook should be that developer can use the given login and password to authorize against 3rd party service so in the end: to override the token. Maybe it should be listener but having an option to override the token or having a token to override? Developer should be able to use the login password and return the token instead of using our default authorization service @filrak

pkarw:
So, maybe the afterUserAuthorize should be mutator extended by login and password input params(?) or other way: if beforeUserAuthorize callback is returning/mutating the token then the login logic is bypassed by this token and not being executed at all (the desired state) @filrak

filrak:
As we discussed on priv 3rd parties should probably be handled differently, maybe @andrzejewsky can suggest something as he worked with this module

filrak:
@andrzejewsky ;)

core/modules/user/store/actions.ts
```diff
@@ -49,7 +50,9 @@ const actions: ActionTree<UserState, RootState> = {
    * Login user and return user profile and current token
    */
   async login ({ commit, dispatch }, { username, password }) {
-    const resp = await UserService.login(username, password)
+    const userAfterHook = userHooksExecutors.beforeUserAuthorize({ username, password })
+    const resp = await UserService.login(userAfterHook.username, userAfterHook.password)
```
andrzejewsky: hmm... with this you are not able to connect with the external login provider because you still use `UserService.login` which calls our API. So maybe this hook should return response and if it's null then you can use `UserService.login` instead.

Eg. you can create dedicated action for that:
```javascript
async attemptLogin (context, { login, password }) {
 const resp = await userHooksExecutors.beforeUserAuthorize({ username, password })

 if (resp) {
  return resp
 }

 return UserService.login(username, password)
}
```
and in the login action:
```javascript
const resp = await dispatch('attemptLogin', { username, password })
```

andrzejewsky:
I've just posted a comment but when it comes to 3rd party providers I think it's more advanced process. This is very similar to payments integration - each provider is different and sometimes that integration consume much effort. So, the extending of login by 3rd party services I think it's a topic for a new task/feature as it should be so flexible to cover as many cases as it could be possible.

So I my opinion @filrak in this PR just modify that hook (read my comment) it should be enough for now as you will be able to override login API call to something else via this hook, but some advanced 3rd integration - it might be a new github issue?

filrak:
After discussions with @andrzejewsky and @patzick I decided to remove beforeUserAuthorize hook - it's clearly redundant and such bahavior should be handled by replacing `UserService.login` That was the exact purpose of introducing those data resolvers

filrak:
Can I merge this one?

pkarw:
It's fine for me!

PR #3344, opened 2019-08-08
Apply lazy-hydrate on category page

### LINKED ISSUE #3327: Optimize category page size with lazy-hydrate
## What is the motivation for adding/enhancing this feature?

Related: https://github.com/DivanteLtd/vue-storefront/pull/3326

https://github.com/maoberlehner/vue-lazy-hydration

I've used it on the Product page within this branch: https://github.com/DivanteLtd/vue-storefront/tree/feature/3281_state_optimization to avoid attributes being included into __INITIAL_STATE__ (#3281)

If we managed to do the same for `category-next.products` we can save 70% of category page size and decrease the TTI significantly.

**The challenge**

We can hydrate [`<product-listing`](https://github.com/DivanteLtd/vue-storefront/blob/f56113fc10180ffb106fe7fb3499fa2e51bc9286/src/themes/default/pages/Category.vue#L70) only after the CSR request for products has been executed (and it's always executed [in here](https://github.com/DivanteLtd/vue-storefront/blob/f56113fc10180ffb106fe7fb3499fa2e51bc9286/src/themes/default/pages/Category.vue#L142))

The problem is that `lazy-hydarate` has only the following triggers: `ssr-only`, `on-interaction`, `when-visible`.

How to add the hydration only after products have been loaded? 

Otherwise, we're getting CSR/SSR mismatch - when the hydration is done before the data is provided.

Idea: maybe the `product-listing` component should take data not from the getter but from kind of Promise (in the end it could be a getter but returning promise; not sure how it will work and if it's even possible?)

**How to remove `category-next.products` from __INITIAL_STATE__

It's very easy. Modify your `default.json`:

FROM:

```
    "initialStateFilter": ["__DEMO_MODE__", "version", "storeView", "attribute", "checkout"],
```

TO:

```
    "initialStateFilter": ["__DEMO_MODE__", "version", "storeView", "attribute", "checkout", "category-next.products"],
```

- [ ] `window.__INITIAL_STATE__` doesn't contains `category-next.products`

- [ ] YES
- [ ] NO

Pick one option.

andrzejewsky:
### Related issues
closes #3327 

This is a PoC of introducing lazy-hydrate for category pages. I had to face many issues with that. First of all the goal was to remove `category-next.products` from `__INITIAL_STATE__` but it causes lack of reactivity in the vuex so the `products` field will not be refreshed. Second one - SSR rendering - the `asyncData` is calling before components were rendered and it stores the products in the vuex state that you have deleted beforehand.... and even more...

let me explain what I did and why:
1. In order to keep reactivity we can't just remove fields from the `state` - I've introduced `filterInitialState` which uses new option in the config `initialStateFilterDefaults ` that includes _fields that will be set to their default values_ I used new option because we might need a possibility to remove totally some field, so `initialStateFilter` still works as before.
2. Because we have deleted (set to default) products by the filter (`initialStateFilterDefaults`), so there is no data in the state but we actually have fetched products so... I stored them in the just ordinary variable `ssrProducts` and if it's a SSR, it uses this variable, otherwise - read the data from getter (computed property `categoryProducts`)

The rest of code should be understandable. I know it's a bit _hacky_ not it does work. I've not noticed a big difference in the lighthouse, but I tested this only on the dev environment. I have to admit that it reduces page size by ~40-44% though. 

Let me know guys what you think.

### DIFF
--- a/src/themes/default/pages/Category.vue
+++ b/src/themes/default/pages/Category.vue
@@ -67,7 +67,10 @@
             </h4>
             <p>{{ $t('Please change Your search criteria and try again. If still not finding anything relevant, please visit the Home page and try out some of our bestsellers!') }}</p>
           </div>
-          <product-listing :columns="defaultColumn" :products="getCategoryProducts" />
+          <lazy-hydrate :trigger-hydration="!loading" v-if="isLazyHydrateEnabled">
+            <product-listing :columns="defaultColumn" :products="getCategoryProducts" />
+          </lazy-hydrate>
+          <product-listing v-else :columns="defaultColumn" :products="getCategoryProducts" />
         </div>
       </div>
     </div>
@@ -75,6 +78,7 @@
 </template>
 
 <script>
+import LazyHydrate from 'vue-lazy-hydration'
 import Sidebar from '../components/core/blocks/Category/Sidebar.vue'
 import ProductListing from '../components/core/ProductListing.vue'
 import Breadcrumbs from '../components/core/Breadcrumbs.vue'
@@ -86,6 +90,7 @@ import ButtonFull from 'theme/components/theme/ButtonFull.vue'
 import { mapGetters } from 'vuex'
 import uniq from 'lodash-es/uniq'
 import onBottomScroll from '@vue-storefront/core/mixins/onBottomScroll'
+import rootStore from '@vue-storefront/core/store';
 
 const composeInitialPageState = async (store, route) => {
   try {
@@ -106,6 +111,7 @@ const composeInitialPageState = async (store, route) => {
 
 export default {
   components: {
+    LazyHydrate,
     ButtonFull,
     ProductListing,
     Breadcrumbs,
@@ -118,7 +124,8 @@ export default {
     return {
       mobileFilters: false,
       defaultColumn: 3,
-      loadingProducts: false
+      loadingProducts: false,
+      loading: true
     }
   },
   computed: {
@@ -129,6 +136,9 @@ export default {
       getCategoryProductsTotal: 'category-next/getCategoryProductsTotal',
       getAvailableFilters: 'category-next/getAvailableFilters'
     }),
+    isLazyHydrateEnabled () {
+      return config.products.lazyLoadingCategoryProducts
+    },
     isCategoryEmpty () {
       return this.getCategoryProductsTotal === 0
     },
@@ -143,12 +153,16 @@ export default {
     if (isServer) next() // SSR no need to invoke SW caching here
     else if (from.name) { // SSR but client side invocation, we need to cache products
       next(async vm => {
+        vm.loading = true
         await vm.$store.dispatch('category-next/cacheProducts', { route: to })
+        vm.loading = false
       })
     } else { // Pure CSR, with no initial category state
       next(async vm => {
+        vm.loading = true
         await composeInitialPageState(vm.$store, to)
         await vm.$store.dispatch('category-next/cacheProducts', { route: to })
+        vm.loading = false
       })
     }
   },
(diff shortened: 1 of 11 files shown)


pkarw:
@patzick and @filrak take a look at this one. It’s probably our way to beat Lighthouse 👍

src/themes/default/pages/Category.vue
```diff
@@ -134,21 +143,33 @@ export default {
...
+      if (isServer) {
+        return ssrProducts
+      }
+
+      return this.getCategoryProducts
     }
   },
   async asyncData ({ store, route }) { // this is for SSR purposes to prefetch data - and it's always executed before parent component methods
     await composeInitialPageState(store, route)
+    ssrProducts = store.getters['category-next/getCategoryProducts']
```
pkarw: I don't like this `ssrProducts` local variable we should probably hide it inside the `category-next/getCategoryProducts` getter; it would be much cleaner and easier to understand if user will be to use only this getter and to have the logick distinguishing if it's CSR (so the state should be returned) or SSR (so the local kept reference should be returned instead)

src/themes/default/pages/Category.vue
```diff
@@ -118,7 +126,8 @@ export default {
     return {
       mobileFilters: false,
       defaultColumn: 3,
-      loadingProducts: false
+      loadingProducts: false,
```
pkarw: is this `loadingProducts` used anywhere?

andrzejewsky:
@filrak 
> If I understand correctly
> - initialStateFilter removes certain STORES from Vuex completely
> - initialStateFilterDefaults sets certain FIELDS to their initial values, right?
Yeah this is how it works.

You are right, we can set always initial values instead of removing them. I did this in that way just because I wanted to avoid potential errors / BC-breaks, but I'll check that later.

src/themes/default/pages/Category.vue
andrzejewsky: hmm @pkarw right, hiding this inside of getter sound much better but I'm wondering if I'm able to get access to this data... but I'll check, for sure worth to investigate 👍

src/themes/default/pages/Category.vue
andrzejewsky: yeah in `onBottomScroll` so I kept that

andrzejewsky:
Ok, I've spent some time on it and I even removed this _local-variable-way_ hack @pkarw . Actually everything works via vuex (getters). It was possible because I've changed the way of the filtering `__INITIAL_STATE__`. Now, the state filter is calling just before the output render (sending to the client), so the SSR application is able to use fully vuex - beforehand it was a problem because sometimes we were removing the state fields when the application wanted to read something from the state (this is what I've noticed). Now, the `initialStateFilter` sets indicated fields to their default values (as you mentioned @filrak )

pkarw:
OK, for me it's fine! we need some @alinadivante on QA + @patzick final approval

pkarw:
@andrzejewsky just one more thing :-) can we enable this lazy loading feature only if config variable like: config.products..useLazyLoading?

alinadivante:
I started testing it today, but I will continue on Friday. Just write here if any others changes will be applied and what I should take into account during testing :)

andrzejewsky:
@pkarw I added option `config.products.lazyLoadingCategoryProducts`. It enables `lazy-hydrate` component and puts `category-next.products` to the filter list (because this option doesn't make sense without that filter)

pkarw:
Cool, thanks! Waiting for @alinadivante feedback before we merge it in

alinadivante:
for me it works good :) 
I didn't notice anything wrong

pkarw:
Ok! I’m merging it in

pkarw:
So @andrzejewsky the next challenge will be to optimize the product page ;) but first @patzick must finish his refactoring effort on it

pkarw:
Great job! Thanks!

Issue #3337, opened 2019-08-06
Refactor the minor-vuex-stores to the proper conventions


pkarw:
## What is the motivation for adding / enhancing this feature?

We're just a step from having all the vuex stores refactored to keep up with our coding standards :) Well, in a away (product vuex store still needs some love but the minimal refactoring has been already done)

## What are the acceptance criteria 

The following stores got refactored (the list is keeping the priorities - starting with the highest):
- [x] attribute,
- [x] order,
- [x] tax,
- [x] user,
- [x] stock,
- [x] checkout,
- [x] payment,
- [x] compare,
- [x] newsletter,
- [x] reviews,
- [x] wishlist,
- [x] cms.

Only the `cms` is kind of bigger one. It could be done at the very end.

Each should be covered with the following changes:
- [ ] getters refactored
- [ ] data resolvers extracted
- [ ] async/await instead of promises
- [ ] helpers extracted
- [ ] no direct state changes (only mutations)
- [ ] no modfify UI in the vuex action directly

Multiple Pull Requests are welcomed. Some vuex stores refactored even to 1.11 are welcomed.

pkarw:
Reference - `core/modules/cart`, `core/modules/catalog-next`

alinadivante:
I have to reject this issue, because, as I mentioned in PR #3378 I have problem with displaying prices in offline mode.

pkarw:
@alinadivante are you using the newest API (test.storefrontcloud.io/api or local copy)?

alinadivante:
@pkarw You are right. Today I changed endpoints in my config and I used demo.storefrontcloud.io
Using test.storefrontcloud.io/api endpoints then problem does not occur.

PR #3317, opened 2019-07-31
Refactoring vuex cart module

andrzejewsky:
### Related issues
closes #3249

### Short description and why it's useful
Refactoring the cart module.

Checklist:
- [x] move network calls to the data resolvers
- [x] action: `syncPaymentMethods` refactored
- [x] action: `syncShippingMethods ` refactored
- [x] action: `sync ` refactored
- [x] action: `load ` refactored
- [x] action: `addItems ` refactored
- [x] action: `removeItem ` refactored
- [x] action: `updateQuantity ` refactored
- [x] action: `configureItem ` refactored
- [x] action: `syncTotals ` refactored
- [x] action: `removeCoupon ` refactored
- [x] action: `applyCoupon ` refactored
- [x] action: `authorize ` refactored
- [x] action: `connect ` refactored
- [x] action: `merge ` refactored
- [x] difflog / notifications
- [x] add typescript typings as much as possible
- [x] unit tests for helpers

### Which environment this relates to
- [x] Test version (https://test.storefrontcloud.io) - this is a new feature or improvement for Vue Storefront. I've created branch from `develop` branch and want to merge it back to `develop`
- [x] RC version (https://next.storefrontcloud.io) - this is a stabilisation fix for Release Candidate of Vue Storefront. I've created branch from `release` branch and want to merge it back to `release`
- [ ] Stable version (https://demo.storefrontcloud.io) - this is an important fix for current stable version. I've created branch from `hotfix` or `master` branch and want to merge it back to `hotfix`

### Upgrade Notes and Changelog
- [x] No upgrade steps required (100% backward compatibility and no breaking changes)
- [ ] I've updated the [Upgrade notes](https://github.com/DivanteLtd/vue-storefront/blob/develop/docs/guide/upgrade-notes/README.md) and [Changelog](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md) on how to port existing VS sites with this new feature

### Contribution and currently important rules acceptance
- [x] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)
- [x] I read the [TypeScript Action Plan](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/TypeScript%20Action%20Plan.md) and adjusted my PR according to it
- [x] I read about [Vue Storefront Modules](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/api-modules/about-modules.md) and I am aware that every new feature should be a module

PR #3395, opened 2019-08-20
Refactoring vuex stock module

andrzejewsky:
### Related issues
closes #3337

Refactoring vuex stock module

### DIFF
--- a/core/modules/cart/store/actions/syncrhonizeActions.ts
+++ b/core/modules/cart/store/actions/syncrhonizeActions.ts
@@ -75,8 +75,9 @@ const synchronizeActions = {
     cartHooksExecutors.afterSync(result)
     return createDiffLog()
   },
-  stockSync: debounce(async ({ dispatch, commit }, stockTask) => {
+  async stockSync ({ dispatch, commit }, stockTask) {
     const product = { sku: stockTask.product_sku }
+
     const cartItem = await dispatch('getItem', { product })
 
     if (!cartItem || stockTask.result.code === 'ENOTFOUND') return
@@ -99,7 +100,7 @@ const synchronizeActions = {
       product: { info: { stock: i18n.t('In stock!') }, sku: stockTask.product_sku, is_in_stock: true }
     })
     EventBus.$emit('cart-after-itemchanged', { item: cartItem })
-  }, 1000)
+  }
 }
 
 export default synchronizeActions
--- a/core/modules/cart/store/actions/syncrhonizeActions.ts
+++ b/core/modules/cart/store/actions/syncrhonizeActions.ts
@@ -1,10 +1,14 @@
 import * as types from '@vue-storefront/core/modules/cart/store/mutation-types'
 import { Logger } from '@vue-storefront/core/lib/logger'
 import { isServer } from '@vue-storefront/core/helpers'
+import debounce from 'lodash-es/debounce'
+import overArgs from 'lodash-es/overArgs'
 import config from 'config'
 import { StorageManager } from '@vue-storefront/core/lib/storage-manager'
 import { CartService } from '@vue-storefront/core/data-resolver'
 import { createDiffLog } from '@vue-storefront/core/modules/cart/helpers'
+import i18n from '@vue-storefront/i18n'
+import EventBus from '@vue-storefront/core/compatibility/plugins/event-bus'
 import { cartHooksExecutors } from './../../hooks'
 
 const synchronizeActions = {
@@ -42,10 +46,10 @@ const synchronizeActions = {
     Logger.warn('The "cart/serverPull" action is deprecated and will not be supported with the Vue Storefront 1.11', 'cart')()
     return dispatch('sync', { forceClientState, dryRun })
   },
-  async sync ({ getters, rootGetters, commit, dispatch }, { forceClientState = false, dryRun = false }) {
+  async sync ({ getters, rootGetters, commit, dispatch, state }, { forceClientState = false, dryRun = false }) {
     const shouldUpdateClientState = rootGetters['checkout/isUserInCheckout'] || forceClientState
     const { getCartItems, canUpdateMethods, isSyncRequired, bypassCounter } = getters
-    if (!canUpdateMethods || !isSyncRequired) return
+    if (!canUpdateMethods || !isSyncRequired) return createDiffLog()
     commit(types.CART_SET_SYNC)
     const { result, resultCode } = await CartService.getItems()
     const { serverItems, clientItems } = cartHooksExecutors.beforeSync({ clientItems: getCartItems, serverItems: result })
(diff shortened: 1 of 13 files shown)


core/modules/cart/store/actions/syncrhonizeActions.ts
```diff
@@ -70,7 +73,31 @@ const synchronizeActions = {
     Logger.error(result, 'cart')
     cartHooksExecutors.afterSync(result)
     return createDiffLog()
-  }
+  },
+  stockSync: debounce(async ({ dispatch, commit }, stockTask) => {
```
pkarw: You added denounce for a case that cart sync is executed in short intervals and stock endpoint responding slowly? Or what’s the case for this?

core/modules/catalog/store/stock/actions.ts
```diff
@@ -1,196 +1,73 @@
...
   clearCache ({ commit }) {
     commit(stockMutationTypes.SET_STOCK_CACHE, {})
-  },
-  async stockAfterCheck ({ dispatch, commit }, event) {
-    setTimeout(async () => {
-      // TODO: Move to cart module
-      const cartItem: any = await dispatch('cart/getItem', event.product_sku, {
-        root: true
-      })
-      if (cartItem && event.result.code !== 'ENOTFOUND') {
```
pkarw: This check - ENOTFOUND was for a offline mode case; I couldn’t find it anywhere in the new code?

pkarw:
@alinadivante the result is that /stock/check api requests is being made after each add to cart; in case the product is not available it should be removed form the cart with error meassage; otherwise “In stock!” Label should appear on the cart item

When entering checkout checkout is waiting for the results of all stock checks

core/modules/cart/store/actions/syncrhonizeActions.ts
andrzejewsky: This is old action `stockAfterCheck` which was moved to the cart module. I used `debounce` instead of `setTimeout`.

core/modules/catalog/store/stock/actions.ts
andrzejewsky: Basically I wanted to ask about that, because I used instead `result.code !== 404`:  (https://github.com/DivanteLtd/vue-storefront/pull/3395/files#diff-729df756e1596492966011c1249d3b8bR80) - I'll change it

alinadivante:
ok, from my side:
1. `TypeError: Failed to fetch` when I tried add to cart available product in offline mode
2. After the internet connection is back - `Add to cart` button remains inactive

3. Unfortunately I don't see the label `"In stock!"` in the products added to the cart..

4. and one more question/doubt: if I add for example this product http://localhost:3000/women/tops-women/tees-women/tees-25/elisa-evercool-and-trade-tee-1465.html?childSku=WS06  (`qty = 76`) to the cart I see request `stock/check` with response: ``` {"code":200,"result":{"item_id":1452,"product_id":1452,"stock_id":1,"qty":76,"is_in_stock":true,"is_qty_decimal":false,"show_default_notification_message":false,"use_config_min_qty":true,"min_qty":0,"use_config_min_sale_qty":1,"min_sale_qty":1,"use_config_max_sale_qty":true,"max_sale_qty":10000,"use_config_backorders":true,"backorders":0,"use_config_notify_stock_qty":true,"notify_stock_qty":1,"use_config_qty_increments":true,"qty_increments":0,"use_config_enable_qty_inc":true,"enable_qty_increments":false,"use_config_manage_stock":true,"manage_stock":true,"low_stock_date":null,"is_decimal_divided":false,"stock_status_changed_auto":0}}``` 
where there is still `qty ": 76,` no matter how many times I add this product to the cart..  Shouldn't it be changed? Shouldn't this amount decrease accordingly?

5. When I choose a product, for example this above one, it has `Quantity (76 available)`, and I change the quantity (on Product Details Page) to 5, and after adding this product to the cart I change my mind and I'll think that I need all available items of this product, so again I change Quantity (but on the cart) to 76 - I will receive message `The requested qty is not available`

Probably sums the amount entered on the PDP with the amount entered in the cart. In my opinion, this is wrong behavior.

6. And the last thing, when I try to add to the cart bundle product http://localhost:3000/gear/gear-3/sprite-yoga-companion-kit-45.html - I receive  `TypeError: Cannot read property 'clientNotifications' of undefined` message (there are no requests on Network tab)

pkarw:
ad5) @alinadivante do you remember? it was a bug in Magento API - for updates it's adding current value + new value instead of using just new value for checking the stock; it's not a VS issu

pkarw:
The rest is up to be checked by @andrzejewsky; you'll become a `cart` module master afterwards ;-]

core/modules/cart/store/actions/syncrhonizeActions.ts
andrzejewsky: I've been testing that - doesn't work without that timeout and I don't really know why, hard to explain, it seems like there is no `result` field (even thought i can see it in the console.log) but it appears afters some time... that's why probably someone used a delay here

andrzejewsky:
@pkarw @alinadivante ready. I'm looking forward to `the card module award` 🏆  😄

pkarw:
OK, waiting for @alinadivante feedback in order to merge this one in

alinadivante:
@andrzejewsky what about my doubts (point 4) ?

1. Please remove edit mode if user opens the cart and there is no internet
2. Small fix for this message `The requested qty is not available This product is out of stock. ` - Please add dote at the end of the first sentence :)
3. There is no  label "In stock!" for **bundle** and **group** products

4.Could you look at this screenshot?
image
I'm not sure if this case is the same as I wrote in point 5, but I change the quantity by entering the available number, and still have the message that there is no such quantity..
5. And this one:
→ I can open two side menus at the same time
This is probably due to this PR #3317, but maybe now you can fix it :)

andrzejewsky:
@alinadivante 
1. Done
2. That message comes from backend, can't change it
3. Fixed for grouped ones
4. This is API issue as that message comes from there. It occurs only on `test` API (`next` is fine)
5. Fixed

andrzejewsky:
another note:
@pkarw Unfortunately I couldn't remove that `callback_event` for `TaskQueue.queue` because without that checking stock doesn't work. The reason of that is the `queue` puts the request to the queue, so the returned task has not had response yet - it must wait for executing. This probably the reason why someone used a timeout for that... and that's why I have to keep that callback event as there is no other way to wait for full response.

pkarw:
Yeah, the callback_event is here to stay with us for some time. We shouldn’t have this denounce/setTimer though - is it random side effect - we should refactor it to proper callback_event callback then or change from queue to execute (which returns promise); however it will change the behavior that stock checks are queued in ifffline mode; so callback_event is preferred; can you change it please?

andrzejewsky:
@pkarw I've already changed, it's ready

PR #3397, opened 2019-08-21
fix issue productlist image not shown when child product image empty …

ngongoll:
Magento1 with configuable products. If child/variant product don't have a image the product doesn't have a image in category view when category is selected from menu. Add fallback condition for value === 'no_selection'


core/modules/catalog/helpers/index.ts
```diff
@@ -584,3 +584,15 @@ export function configurableChildrenImages (product) {
...
 }
+
+/**
+ * check if one of the configuableChildren has an image
+ */
+
+export function childHasImage (children) {
+  let hasImage = false;
+  children.forEach((child) => {
+    if (child.image.length == 0 || child.image === 'no_selection') hasImage = true
```
patzick: I'm not sure I get this right - could you explain to me why a child has an image if its images have no length or if the image is `no_selection`? Shouldn't that be the opposite conditions?
also please concider to use `includes` instead of `forEach` and provide a default value for children, because with the wrong argument we could have an error here. If you'd add unit tests for this helper method I'd be really grateful :)

core/modules/catalog/helpers/index.ts
```diff
@@ -567,11 +567,11 @@ export function attributeImages (product) {
 
 export function configurableChildrenImages (product) {
   let configurableChildrenImages = []
-  if (product.configurable_children && product.configurable_children.length > 0) {
+  if (product.configurable_children && product.configurable_children.length > 0 && this.childHasImage(product.configurable_children)) {
     let configurableAttributes = product.configurable_options.map(option => option.attribute_code)
     configurableChildrenImages = product.configurable_children.map(child =>
       ({
-        'src': getThumbnailPath(child.image, config.products.gallery.width, config.products.gallery.height),
+        'src': getThumbnailPath(((child.image === '' || child.image === 'no_selection') ? product.image : child.image), config.products.gallery.width, config.products.gallery.height),
```
patzick: we could use childHasImage here passing child as array

core/modules/catalog/helpers/index.ts
```diff
@@ -567,11 +567,11 @@ export function attributeImages (product) {
 
 export function configurableChildrenImages (product) {
   let configurableChildrenImages = []
-  if (product.configurable_children && product.configurable_children.length > 0) {
+  if (product.configurable_children && product.configurable_children.length > 0 && this.childHasImage(product.configurable_children)) {
```
patzick: if childHasImage method would have default value then here it could only look like this
```suggestion
  if  (childHasImage(product.configurable_children)) {
```

pkarw:
@ngongoll can you please introduce the requested changes in order to merge this PR?

pkarw:
Looks fine!

PR #3403, opened 2019-08-22
Bugfix/category hydration issues

patzick:

Fixes hydration errors caused randomly by not fetching scripts before client hydration started.
This should increase TTI and get rid of this problem.

During tests, this error hasn't shown up even once. Haven't noticed any unpleasant side-effects.

QA - please test it thoroughly by refreshing category page multiple times.
As change is in the root of the application, please also check Other pages to be sure.

### Edit: Extended explanation
The error appears mostly on Linux-based systems distributions (with docker, Linux, mac). When a client gets server image it starts hydrating too fast, even if didn't get yet route component or it's dependencies.
To avoid that kind of situations route chunks (and all dependencies) should be smaller. In some points, client browser can skip waiting for all dependencies and start hydrating too soon.

To switch off this new set of lazy hydrating route component we have the property:
`themeConfig.lazyRouteComponentHydration`
in a config file, which by default is set to `true`

### DIFF
--- a/src/themes/default/layouts/Default.vue
+++ b/src/themes/default/layouts/Default.vue
@@ -25,7 +25,11 @@
         :is-open="isWishlistOpen"
         @close="$store.commit('ui/setWishlist')"
       />
-      <slot />
+      <lazy-hydrate when-idle :trigger-hydration="!isRouteComponentLazyHydrated">
+        <div>
+          <slot />
+        </div>
+      </lazy-hydrate>
       <main-footer />
       <notification />
       <sign-up />
@@ -51,6 +55,7 @@ import OfflineBadge from 'theme/components/core/OfflineBadge.vue'
 import { isServer } from '@vue-storefront/core/helpers'
 import Head from 'theme/head'
 import config from 'config'
+import LazyHydrate from 'vue-lazy-hydration'
 
 const SidebarMenu = () => import(/* webpackPreload: true */ /* webpackChunkName: "vsf-sidebar-menu" */ 'theme/components/core/blocks/SidebarMenu/SidebarMenu.vue')
 const Microcart = () => import(/* webpackPreload: true */ /* webpackChunkName: "vsf-microcart" */ 'theme/components/core/blocks/Microcart/Microcart.vue')
@@ -76,7 +81,10 @@ export default {
       isSidebarOpen: state => state.ui.sidebar,
       isMicrocartOpen: state => state.ui.microcart,
       isWishlistOpen: state => state.ui.wishlist
-    })
+    }),
+    isRouteComponentLazyHydrated () {
+      return config.themeConfig.lazyRouteComponentHydration
+    }
   },
   methods: {
     onOrderConfirmation (payload) {
@@ -118,7 +126,8 @@ export default {
     CookieNotification,
     OfflineBadge,
     OrderConfirmation,
-    AsyncSidebar
+    AsyncSidebar,
+    LazyHydrate
   }
 }
 </script>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Invalid Discount code error handled by theme - @grimasod (#3385)
+- Hydration problems on category page caused by to early hydration - @patzick (#3403)
 
 ## [1.10.0] - 2019.08.10
 
--- a/config/default.json
+++ b/config/default.json
@@ -421,6 +421,9 @@
       "sendConfirmation": true
     },
     "theme": "@vue-storefront/theme-default",
+    "themeConfig": {
+      "lazyRouteComponentHydration": true
+    },
     "analytics": {
       "id": false
     },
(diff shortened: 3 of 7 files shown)


pkarw:
Try if switching off lazy layout loading in App.vue is solving this issue - it would be probably a btter (simpler) solution:https://github.com/DivanteLtd/vue-storefront/blob/463036f87461e9142caeb54fd092be4ca8294228/src/themes/default/App.vue#L11

Switching to normal eager imports, commenting out not-used layouts

pkarw:
Cool; let's merge it in

patzick:
cc @dawerx - pleas confirm that works for you guys :)

adamjedlicka:
I think this change requires more testing. It could cause some problems on low end mobile devices or on slower network connections. Because you are lazy-hydrating the whole page, if user interacts with it before hydration (which has higher probability of happening when the page is loading for longer period of time - low end phones and slow network), Vue discards the page and re-renders it from scratch. This causes the page to disappear for a brief moment and resets the whole state (so for eaxmple selected filter won't be selected anymore).

It's possible that most of these issues will be automatically solved by caching from service worker. Not on initial load, though.

That's why I would suggest testing it on production mode, on real devices and real networks.

Darwex:
Hi.

I agree with Adam.

The fix seem to be working, I was not able to reproduce the SSR issue on vanila VSF, however we should dig deeper to identify the core problem.

The page seem to be slower though, menu content appears after 1 second.
I will test this with Dr. Max VSF instance as well.

pkarw:
Please note that the impact of this issue (and the solution) is pretty limited as it can occur only when hydration takes place which is always just the 1st page load (all the subsequent requests are processed 100% CSR)

patzick:
@adamjedlicka , @Darwex guys, there is also another way to help you with your problem quickly. Not too pretty but should work if you would turn off lazy loading on the category page. This one is causing your problems so it would be the quickest way for fix.
Then you'll just change `lazyRouteComponentHydration` config to false. And in router:
```js
const Category = () => import(/* webpackChunkName: "vsf-category" */ 'theme/pages/Category.vue')
// change to
import Category from 'theme/pages/Category.vue'
```

We tested our develop which has category module rewritten. Problem is not appearing there, could you check it by yourselves on develop branch if you're able to reproduce it there?
Then with the new version, you could return to lazy load category page and everything should be good.

Please let us know what do you think!

adamjedlicka:
@patzick I was still able to reproduce it on the latest develop of VSF. But it took a long time (about 10 minutes). The reason why it's much more rare on the develop is because of this commit c2c6e8f9f69d76ed8dd487fd48aa4a1aa56cf73d . This one is on the right track of solving the issue properly but there is still some race condition left.

pkarw:
@adamjedlicka this PR hasn't been yet merged to develop. please make syoure you've tested rhe right version - this PR not 1.11.rc1 (develop)

```
git checkout -b patzick-bugfix/category-hydration-issues hotfix/v1.10.1
git pull https://github.com/patzick/vue-storefront.git bugfix/category-hydration-issues
```

adamjedlicka:
@pkarw Are you sure it's not merged? I can see it here: https://github.com/DivanteLtd/vue-storefront/blob/develop/core/client-entry.ts#L28 and here: https://github.com/DivanteLtd/vue-storefront/blob/develop/core/modules/url/store/actions.ts#L32

pkarw:
@adamjedlicka I meant that this PR (#34030) hasn't been yet merged :-)

patzick:
@adamjedlicka small misunderstanding here between you guys; we'll be checking to develop also to provide a better fix for this issue.
But for this version, there is nothing more we can do I think. Have you guys considered turning off lazy load category on the router or using lazy hydration from this PR? Have you checked the impact on your site for both solutions?

adamjedlicka:
@patzick I don't think turning off lazy loading of page components will help. Vue should abstract the difference away, if there is bug in Vue in lazy loading of components I'm sure someone else in different project would already noticed.

Also it does not happen only in category. It happens anywhere where dynamic URL dispatch is used (category, product, ...). That's why I think it's caused by the router. And even if turning off lazy loading of category would help, we can't afford to turn off lazy loading for all of these pages.

src/themes/default/layouts/Default.vue
```diff
@@ -25,7 +25,11 @@
         :is-open="isWishlistOpen"
         @close="$store.commit('ui/setWishlist')"
       />
-      <slot />
+      <lazy-hydrate when-idle :trigger-hydration="!isRouteComponentLazyHydrated">
+        <div>
```
andrzejewsky: Do you need that extra `div` it might work with `<template></template>`?

pkarw:
OK, but to just sync; @adamjedlicka does this PR (3403) with a lazy-hydration for the whole page content solves the issue you?

adamjedlicka:
@Darwex is testing it right now. I'm fairly certain that this will fix the SSR issue because it is basically the same as my fix, so we will probably just merge this change and be done with it. But let's wait for David's final opinion.

pkarw:
OK, cool thanks for confirmation - @patzick is double-checking one more another hypothesis regarding the client-entry.ts `.mount()` call - if there is a race condition between this call (which causes hydration) and between `urlDispatcher`. If it's a case we'd be able to propose another, simpler solution to this issue. However, it's basically the same solution as lazy-hydrate (to postpone the hydration)  just maybe more elegant ;)

patzick:
@adamjedlicka @Darwex we've found a solution to this problem which is not related with lazy hydration. Please give ma ~1h to clean up and final tests of a solution and I will provide another PR.

Btw. that's really great discussion here, and I'm grateful guys that you provided your concerns and point of view.
Expect another PR really soon 🚀

patzick:
@adamjedlicka , @Darwex let's move to #3412, I'm closing this one without merge :)

PR #3412, opened 2019-08-23
Bugfix/too early hydration with urldispatcher

patzick:
### Related issues

 
Improvement for #3403 

Fixes hydration issue caused by UrlDispatcher with a lock to prevent multiple dynamic route calculations.
Continue thread from #3403 

@adamjedlicka, @Darwex - guys please check if this solution solves the problem. Are you able to reproduce this now and have you seen performance gain? My impression is that the page loads faster with route lock.


adamjedlicka:
This fix doesn't seem to work for us. We are still getting the SSR error and now it's even worse. Before the only visible problem was duplicated footer. Now the whole client-side rendering stops working. Whole page is non-interactive and lazy-loaded components won't load at all.

I'm adding screenshot of the error: [https://i.imgur.com/ALmOSSR.png](https://i.imgur.com/ALmOSSR.png)

Darwex:
BTW the previous fix wit LazyLoad was working guys, it made the website slower, but it worked :-)

patzick:
@adamjedlicka, @Darwex that's really bad. Let's start with the clean storefront - do you experience the same issue when checkout just this branch?

adamjedlicka:
@patzick No. It looks like it fixes the error for clean VSF (or greatly minimizes it's occurrence). I was unable to get the SSR error to happen even after about fifteen minutes of refreshing the category page.

I think It's not working on Dr. Max because of invalid HTML markup (like <p> inside of another <p>) or SSR renderer having problems with comments inside Vue HTML templates.

patzick:
Okay, that's actually good. I think we could help you somehow with finding the problem in your template. Most cases it is related to cms blocks which are not <no-ssr> wrapped or as you said - some invalid HTML. It shouldn't be caused by the comments though.
Let us know if you need assistance on that, I'll merge this fix for 1.10.1

Darwex:
Hi Guys.

Why was this PR merged? I Understand it may be fixes it on pure VSF but it does not fix it on Dr. Max project. I asked whether we could merge the previous one which fixes it on both, Dr. Max and clean VSF.

@patzick

patzick:
@Darwex this solves an issue without slowing down the page (even probably with making faster TTI). We also discussed that matter also internally and I merged that in after @adamjedlicka confirmation. We didn't have any other reports from implementations. I think first of all we should take a look at yours and try to find the cause.

Darwex:
Hi @patzick 

I understand that. But the first solution worked for both clean VSF and our project. So from our side, I do not consider it as fixed. What is it what we can do to fix it on our project then?

Thanks.

PR #3354, opened 2019-08-11
Fixed rendering after jest watch fix destroyed it.

ResuBaka:
Fixed rendering which doesn't work any more because of PR #3351 but with this fix now jest watch and SSR are working.

PR #3443, opened 2019-08-29
Reduce addRoutes function calls

### LINKED ISSUE #3442: Router beforeEach hooks run many times on page reload
## Current behavior

On a page reload, all the registered beforeEach hooks run many times.

This is because the router calls the hooks when any routes are added to it. Localized multi-store routes are added one store at a time and dynamic routes are also added one at a time, causing the hooks to run each time. With several stores and many pre-loaded dynamic routes (eg many Categories), this can result in hundreds of runs.

On a regular page this might not be noticeable, but it can be a problem on an Account (requiresAuth) page. The User module beforeEach hook dispatches a `user/startSession` event, which leads to `sessionAfterAuthorized` and therefore `user/me` and `user/getOrdersHistory` events. Hundreds of these events run simultaneously, flooding the API, and preventing the page from loading.

It's not noticeable on the demo store, with only a few pre-loaded categories, only 3 stores, and a test user with only a few orders. But it can still be demonstrated.

## Expected behavior

Router beforeEach hooks should ideally only run once, or at least be minimized.

1. Open the file `/core/modules/user/router/beforeEach.ts`
2. At the top of the forEach function, add something like `console.log('user module - beforeEach')`
3. Open Developer Tools in a browser and reload the home page
4. Notice how many times the message appears, around 15 times
5. Browse to some other catalog pages, reload them and see the message many many times

- [x] YES (Partially)
- [ ] NO

Pick one option.

- [x] This is a bug report for current Stable version on https://demo.storefrontcloud.io and should be placed in next stable version hotfix - In this case Developer should create branch from `hotfix` or `master` branch and create Pull Request `4. Hotfix` back to `hotfix`.

## Environment details

- Browser: 
- OS: 
- Node: 
- Code Version: v1.10.0

grimasod:
### Related issues

closes #3442

Greatly reduces the number of times the router beforeEach hooks are run

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `order.order_id` was not assigned in the `orders.directBackendSync` mode - @pkarw (#3398)
 - Hydration problems with UrlDispatcher :rocket: - @patzick (#3412)
 - if condition of quoteId from the `_serverDeleteItem` method on core/modules/cart/store/action.ts - @AshishSuhane (#3415)
+- Router beforeEach hooks running many times - @grimasod (#3443)
 
 ## [1.10.0] - 2019.08.10
 
@@ -119,8 +120,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Product video - retrieve video id from 'video_id' field (if set) instead of 'id' - @afirlejczyk
 - Webpack config improvement - @yogeshsuhagiya (#2689)
 - BaseSelect input event - @ResuBaka (#2683)
-- Fixed static file handler to immediately return 404 status for missing files - @grimason (#2685)
-- Fixed maxAge Response Header for static files and Content-Type for Service Worker - @grimason (#2686)
+- Fixed static file handler to immediately return 404 status for missing files - @grimasod (#2685)
+- Fixed maxAge Response Header for static files and Content-Type for Service Worker - @grimasod (#2686)
 - Default log verbosity is changed to show only errors - @lromanowicz (#2717)
 - Remembering last search query - @webdiver, @patzick (#2787)
 - Extracted ProductImage component to support faster images loading - @przemyslawspaczek (#2925)
--- a/core/lib/module/index.ts
+++ b/core/lib/module/index.ts
@@ -4,7 +4,7 @@ import Vue from 'vue'
 import merge from 'lodash-es/merge'
 import rootStore from '@vue-storefront/core/store'
 import { Logger } from '@vue-storefront/core/lib/logger'
-import { setupMultistoreRoutes } from '..//multistore'
+import { setupMultistoreRoutes } from '../multistore'
 import { router } from '@vue-storefront/core/app'
 import { isServer } from '@vue-storefront/core/helpers'
 import { VSF, VueStorefrontModuleConfig } from './types'
@@ -47,7 +47,6 @@ class VueStorefrontModule {
   private static _extendRouter (routerInstance, routes?: RouteConfig[], beforeEach?: NavigationGuard, afterEach?: NavigationGuard): void {
     if (routes) {
       setupMultistoreRoutes(config, routerInstance, routes)
-      RouterManager.addRoutes(routes, routerInstance)
     }
     if (beforeEach) routerInstance.beforeEach(beforeEach)
     if (afterEach) routerInstance.afterEach(afterEach)
--- a/core/lib/multistore.ts
+++ b/core/lib/multistore.ts
@@ -181,16 +181,16 @@ export function localizedRoute (routeObj: LocalizedRoute | string | RouteConfig 
 }
 
 export function setupMultistoreRoutes (config, router: VueRouter, routes: RouteConfig[]): void {
+  const allStoreRoutes = [...routes]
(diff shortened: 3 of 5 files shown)


pkarw:
I’m not sure if it is reducing the router.addRoutes to just single one (which can obviously solve the multiple routes call once and for all) but maybe we can do it by postponing the registration till very end just before app.mount call(?) the other option is to add a paragraph to the docs on how to use the locking mechanism in the RouteManager level and a note to not use the router hooks directly

Or even better - add docs + add a mechanism to register hooks from roterManager level that are locking the calls themselves

Then, we need to find all the instances or adding the hooks listeners to router directly and replace them with the wrapper calls from the router manager

I think this makes sense as the routermanager is the only class which can be safely used for adding the routes because of the url dispatcher module

@filrak ive mentioned you in the other PR regarding passing routermanager instead of router to the module constructors
@patzick wdyt?

grimasod:
Hi @pkarw, yeah I saw #3412 and tested after it was merged. It didn't affect this issue.

As you suggest, postponing all additions to the end could ensure there was exactly only one addRoutes. I thought about something like that when I started working on this, but it seemed quite complex! And I wonder if it could affect some custom modules or themes?

So, I thought the solution in this PR might be more suitable for a hotfix, as it shouldn't have any knock-on effects. It's not 100% perfect, but it's a huge improvement. In our custom theme, it reduced User beforeEach calls from about 750 to just 1 or 2, depending on the page.

There's another point to consider (not dealt with in this fix) ...

**Is it necessary to create routes for all stores?**

Considering it's not possible to navigate directly between stores. Instead, the app gets reloaded, and all the routes recreated. So only the current store routes need to be added. Right? The exception is SSR for SEO, but that can be handled separately. This also seems like a big change though and could possibly affect some customizations.

In the end, I thought this simple fix would be best for 1.10.1 and maybe a bigger update can be planned for 1.11. What do you think?

grimasod:
@ResuBaka Good point. I haven't done any unit testing work on VS, do you have any suggestions or pointers?

ResuBaka:
@grimasod One point would be this [**test file**](https://github.com/DivanteLtd/vue-storefront/blob/master/core/lib/test/unit/multistore.spec.ts) which we have four the multistore.ts.

Then all unit test are under **test/unit/<function/file you want to test>.spec.ts** inside the folder the to testing file is.

Example of the structure:
**core/lib/multistore.ts** 
**core/lib/test/unit/multistore.spec.ts** or **core/lib/test/unit/setupMultistoreRoutes.spec.ts**

Then the unit test are written in **jest** ([docs](https://jestjs.io/docs/en/getting-started)) when you need some help you can write me in the **slack a dm**. 

Test can be startet with **`yarn test:unit `** the **watch mode** of **jest** does currently not work in 1.10 I have only fixed it in 1.11. (When you want to use the watch feature for easier test writing you could fix it locally for you when you look at these PRs  #3354, #3351)



@pkarw @patzick should we maybe backport my fix so more people can use the watch feature of jest in the older version?

grimasod:
@ResuBaka Thanks. I'll look into it!

grimasod:
> Is it necessary to create routes for all stores?

Thinking about this more... isn't it actually incorrect to generate routes for all stores based on the routes of one store? For example:

- Each store can have a different theme, which could contain different pages
- Each store uses a different Elasticsearch index, which can contain different categories and products - this is quite common when selling overseas. Some products can't be shipped to certain countries.

ResuBaka:
@grimasod 

When I think about that we should do, for a quick improvement to your code check if the stores have enabled the **appendStoreCode** config value because when they don't they most likely have different/need to domains for each store so only mapping stores where the **appendStoreCode** is set to true would improve this too.

grimasod:
@ResuBaka Yeah, you're right.

Static routes are added using `localizedRoute` which accounts for that config value, but the Dynamic routes function has its own construct. It should probably use `localizedRoute`, if possible.

patzick:
This looks really good, I just need a while to test this properly first to be sure :)

pkarw:
Thanks for the discussion, it was really cool! Great insights 👍

@grimasod how about trying to implement these upgrades?

grimasod:
Thanks for the feedback :)

@pkarw I can work on it, but it won't be soon. I'm releasing our v1.10 site this week and don't have much time for anything else. 

1. This sounds good. I just need to look into a bit more, to understand where to call flushRouteQueue

2. This seems like would need quite a bit of testing and maybe it's too risky for a hotfix. WDYT about doing this part in a later PR for v1.11?

pkarw:
Sure, 1.11 sounds reasonable @patzick

pkarw:
Cool; i created a #3454 just for that

mryellow:
Just stopping by to thank you guys for saving me many more hours of debugging on an unrelated project. Also to add a few observations and keywords for those coming after.

Interestingly this only happens with async calls to `next()` in `beforeEach` navigation guards when `addRoute` is used after `beforeEach`.

```
router.beforeEach((to, from, next) => {
  // Single call
  //next();

  // Many calls
  //setTimeout(() => next(), 5000);

  // Many calls
  Promise.resolve()
    .then(() => next());
});

routes.forEach(route => router.addRoute(route));
```

Also note that this does not happen in a very simply Vue application such as this fiddle below. It's somehow related to further complexities around how we're bootstrapping up Vue.

https://jsfiddle.net/MrYellow/zps18dg0/

PR #3472, opened 2019-09-03
Hotfix/v1.9.3

patzick:
Prepare release with backporting #3412 from 1.10.1


benjick:
> TS2339: Property 'finally' does not exist on type 'Promise<void>'.

I had to to change `compilerOptions.lib` in tsconfig.js to this: `["es2017", "dom", "es2018.promise"]`

PR #3510, opened 2019-09-06
3508 fixed comparison list preservation

### LINKED ISSUE #3508: Comparison feature is not keeping a state on page reload.

A small unconvenience - products on comparison list are not preserved between page reloads. That means we can not compare one product with another opened in a new tab.

We should keep products in comparison list in a session. This will prevent resets on page reload. We will be also able to open new links in new tabs with preserved items.

Page reload should not reset comparison list items.

vue-kacper:
closes #3508

Comparison list is now preserved between page refresh action

- [x ] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)
- [x ] I read the [TypeScript Action Plan](https://github.com/DivanteLtd/vue-storefront/blob/master/doc/TypeScript%20Action%20Plan.md) and adjusted my PR according to it

### DIFF
--- a/core/modules/compare/store/plugin.ts
+++ b/core/modules/compare/store/plugin.ts
@@ -6,8 +6,7 @@ const watchedMutations = [types.COMPARE_ADD_ITEM, types.COMPARE_DEL_ITEM, types.
 const cacheStorage = StorageManager.get('compare')
 
 const cachePersistPlugin = (mutation, state) => {
-  const mutations = watchedMutations
-  if (mutations.includes(mutation.type)) {
+  if (watchedMutations.includes(mutation.type)) {
     cacheStorage.setItem('current-compare', state.compare.items).catch((reason) => {
       Logger.error(reason, 'compare')
     })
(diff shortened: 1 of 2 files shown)


core/modules/compare/store/plugin.ts
```diff
@@ -6,8 +6,7 @@ const watchedMutations = [types.COMPARE_ADD_ITEM, types.COMPARE_DEL_ITEM, types.
 const cacheStorage = StorageManager.get('compare')
 
 const cachePersistPlugin = (mutation, state) => {
-  const mutations = watchedMutations.map(m => `compare/${m}`)
-
+  const mutations = watchedMutations
   if (mutations.includes(mutation.type)) {
```
pkarw: wouldn't it be better to remove `const mutations` and have just `watchedMutations.includes` instead?

vue-kacper: Yeah, totally right, we are not using this variable anyway anymore

PR #3518, opened 2019-09-09
[WIP] Server modules

### LINKED ISSUE #2494: Add a way to extend server middlewares + access server context from modules

Currently, there is no way to access server context from modules - either add Express middlewares. Therefore sometimes we need to use [ugly workarounds](https://github.com/pkarw/vue-storefront/blob/bbc1a8e89a87c9aebf549def694e634dd97de21f/core/modules/url/router/beforeEach.ts#L58) to do simple things like server-redirect.

The `Vue.prototype.$ssrRequestContext` is a pretty powerful tool:

```js
    const context = {
      url: req.url,
      output: {
        prepend: (context) => { return '' }, // these functions can be replaced in the Vue components to append or prepend some content AFTER all other things are rendered. So in this function You may call: output.prepend() { return context.renderStyles() } to attach styles
        append: (context) => { return '' },
        appendHead: (context) => { return '' },
        template: 'default',
        cacheTags: null
      },
      server: {
        app: app,
        response: res,
        request: req
      },
      meta: null,
      vs: {
        config: config,
        storeCode: req.header('x-vs-store-code') ? req.header('x-vs-store-code') : process.env.STORE_CODE
      }
    }
```

The `server` section is all ExpressJS. So you can easily add middlewares to `context.server.app` or add response headers to `context.server.response`.

Would be great to have some access to this object from inside modules. Of course accessing it makes only sense when `isServer` is true.

filrak:
### Related issues

closes #2494

todo: ability to pass config to the module in `local.json`

How to use:

1. create `server.js` in the root of your module and export function that will extend the express context:
```js
module.exports = function (app) {
  app.get('/robots.txt', (req, res) => {
    res.end('User-agent: *\nDisallow: ')
  })
}
```
or with config
```js
module.exports = function (app, config) {
  app.get('/robots.txt', (req, res) => {
    res.end(config.content)
  })
}

```
2. Register server side module in `config/local.json`
```
  "serverSideModules": [
    "src/modules/example"
  ],
```
or with config
```
  "serverSideModules": [
    ["src/modules/example", { "content" : "User-agent: *\nDisallow: " }
  ],
```

voila!

**NOTE**: `serverSideModules` property is filtered on client side because it can contain fragile data such as API keys, also it's redundant anyway because it's server side ^^


filrak:
Imho it would be nice to also filter out serverSideModules field from the config json on client side. It will probably contain fragile info such as API keys

filrak:
@kkdg can you add this to modules docs?

core/scripts/server.js
```diff
@@ -120,10 +121,18 @@ app.use('/service-worker.js', serve('dist/service-worker.js', false, {
...
 }))
-
+/** @deprecated in 2.0  in favour of module-based mechanism */
 const serverExtensions = require(resolve('src/server'))
 serverExtensions.registerUserServerRoutes(app)
 
+config.serverSideModules.forEach(serverModule => {
+  if (Array.isArray(serverModule)) {
+    require(resolve(serverModule[0] + '/server.js'))(app, serverModule[1])
+  } else {
```
pkarw: Cool, but few questions:
- Does it support ES6/TS in server modules? Otherwise no possibility to reuse VS code and libs
- do we really need this config variable? It’s one time operation so we can probably traverse core/modules and src/modules(?)
- we need an example module in the module template

ResuBaka: @pkarw 

I can say for the first point that it does not support typescript, but that would be a really nice feature.

patzick: yeah, would be great to have these modules in separate dir or register in some index.js file like modules, this index.js would always be consumed on the server side. I'm afraid that config option could be overcomplicated.

And example +1 :)

filrak: @pkarw @patzick  imho we need to if you want to include modules from node_modules as well. We can't really assume that their always named `vsf-something`. 
Separate index.js is an optiion as well but for sure we can't just traverse the file tree. Please pick either config or file.

About es6 tbh I don't know, will check it. It works exactly the same as server.js extensions

patzick: so I'm voting for registering in a file instead of config :)

pkarw:
@patzick @filrak I get the idea with config - so I vote for config.
But we should have the TypeScript/ES6 support and test it with our Vuex. I mean user should be ably to import store and query elasticsearch (eg. `product/list`) in server module; otherwise it will be pretty much not usefull feature

filrak:
Ok i applied suggested . hanges:
- you can now use ES6 is server side modules (i also got rid of redundant `entry.ts` and just changed zserver.js` to typescript file)
- instead of using config there is `modules/server.ts` file, also renamed `index.ts` to `client.ts` for constancy
- I removed `src/server` folder to keep common approach for server side extensions via modules

Please lmk if that's fine, then I will add upgrade notes

kkdg:
Sure thing!

kkdg:
@filrak This is one of the most anticipated parts of my interest. I will start working on it after 1.10 -> 1.11 upgrade tutorial is done as @pkarw told me to do. thanks!

filrak:
sure!

PR #3540, opened 2019-09-13
Router manager queue

### LINKED ISSUE #3454: Optimize the `RouterManager.addRoutes`

Related to: #3443 

The goal: to reduce the number of times the router beforeEach hooks are run

grimasod:
### Related issues

closes #3454

Efficiently add routes to Vue Router, with optional priority for theme routes


grimasod:
Still in progress

pkarw:
Cool, looks like a step into right direction!

grimasod:
@pkarw What do you think about moving dynamic route generation out of the beforeEach hook? It seems to cause problems there - as mentioned before, registering new routes causes the hook to be called again, so it's kind of a circular event chain.

Have you previously tried generating the routes when the categories and products are loaded? That could be a nice solution, although I haven't tried it yet :)

pkarw:
Hi! It won’t cover all the custom urls user might have - I think we can’t aboid it. It would be pretty much optimal to have a single call for registering all the routes other than those and then additional call like forcing to add the routes exactly in here - in the hook

?

grimasod:
Ok, I see. So, we need to keep the hook for those cases.

I'm aiming to have a single call for all known urls. Only the dynamic urls are a  remaining challenge.

The catalog urls are not know on the initial load, except for the categories that are prefetched. So, I'm thinking about how to handle them best. I think avoiding the beforeEach hook might be better for these urls.

pkarw:
Hi @grimasod by fixing some issue regarding multistore url dispatcher in 1.10.3 I also optimized the code a little bit: https://github.com/DivanteLtd/vue-storefront/pull/3568/files; now we're registering just the routes for the current store as it makes no sense to register routes for other stores (as the context switcher is being executed before router processes the URL - in `createApp`; so we now always have the valid `currentStoreView()`)

grimasod:
Ok, thanks. Sounds great. I'll look at it tomorrow and merge into this PR. I had got it working, but for the client only. The only other remaining issue I have is a bug on page-not-found. I should get that resolved tomorrow too.

grimasod:
@pkarw This is ready for testing. I ported that 1.10.3 PR and combined the changes. That was a fun challenge :) 
Just one question - the routes probably don't need to have the store code prefixed to the route name anymore (eg name: 'it-category'). Provisionally I've removed them. If there's a case I've not found, I can easily put them back.
Other than that, I just need to update the documentation, when these changes are accepted

pkarw:
Cool, thanks @grimasod; We'll review it shortly

core/lib/multistore.ts
```diff
@@ -234,9 +241,10 @@ export function localizedRoute (routeObj: LocalizedRoute | string | RouteConfig
     if (typeof routeObj !== 'object') {
       return '/' + storeCode + routeObj
     }
-    if (routeObj.name) {
-      routeObj.name = storeCode + '-' + routeObj.name
-    }
+    // TODO - need route name prefix?
+    // if (routeObj.name) {
+    //   routeObj.name = storeCode + '-' + routeObj.name
```
pkarw: I think it's still required for `setupMultistoreRoutes` which is adding the static routes - I belive we should optimize `setupMultistoreRoutes` the same way we did with `prepareDynamicRoutes` to add the routes just for the current storeView; shouldn't we?

core/lib/multistore.ts
```diff
@@ -246,17 +254,14 @@ export function localizedRoute (routeObj: LocalizedRoute | string | RouteConfig
...
+  const allRoutes = []
+  const isMultistore = config.storeViews.multistore === true && config.storeViews.mapStoreUrlsFor.length > 0
+  const { storeCode } = currentStoreView()
+  if (isMultistore && storeCode) {
+    allRoutes.push(...routes.map(route => localizedRoute(Object.assign({}, route), storeCode)))
+  } else {
+    allRoutes.push(...routes)
   }
-  RouterManager.addRoutes(allStoreRoutes, router)
+  RouterManager.addRoutes(allRoutes, router, true, priority)
```
pkarw: Oh, I've seen hyou already optimized it :)

core/modules/url/helpers/index.ts
```diff
@@ -1,64 +1,55 @@
...
   const parametrizedRoute = Object.assign({}, routeData)
   parametrizedRoute.params = Object.assign({}, parametrizedRoute.params || {}, query)
-  if (storeCodeInPath && !parametrizedRoute.name.startsWith(storeCodeInPath + '-')) {
-    parametrizedRoute.name = storeCodeInPath + '-' + parametrizedRoute.name
-  }
+  // TODO - need route name prefix? (if not storeCodeInPath not needed)
+  // if (storeCodeInPath && !parametrizedRoute.name.startsWith(storeCodeInPath + '-')) {
+  //   parametrizedRoute.name = storeCodeInPath + '-' + parametrizedRoute.name
+  // }
   return parametrizedRoute
```
pkarw: I think we don't need it however it should be tested if there won't be duplicates

core/server-entry.ts
```diff
@@ -56,28 +58,26 @@ function getHostFromHeader (headers: string[]): string {
...
+  let storeCode = context.vs.storeCode
+  if (config.storeViews.multistore === true) {
+    if (!storeCode) { // this is from url
+      const currentRoute = Object.assign({ path: queryString.parseUrl(context.url).url/* this gets just the url path part */, host: context.server.request.headers.host })
+      storeCode = storeCodeFromRoute(currentRoute)
+    }
+  }
+  const { app, router, store, initialState } = await createApp(context, context.vs && context.vs.config ? context.vs.config : buildTimeConfig, storeCode)
+
+  RouterManager.flushRouteQueue(router)
```
pkarw: we should move it to `createApp` to avoid the risk that this line will be missed across the upgrades/user stores

core/server-entry.ts
```diff
@@ -56,28 +58,26 @@ function getHostFromHeader (headers: string[]): string {
...
-          storeCode = storeCodeFromRoute(currentRoute)
-        }
-        if (storeCode !== '' && storeCode !== null) {
-          prepareStoreView(storeCode)
-        }
-      }
       const matchedComponents = router.getMatchedComponents()
-      if (!matchedComponents.length) {
-        return reject(new HttpError('No components matched', 404))
+      if (!matchedComponents.length || !matchedComponents[0]) {
```
pkarw: I don't get the ` !matchedComponents[0]` part of this condition?

grimasod: During development there were cases when getMatchedComponents returned `[undefined]` - an array with a single undefined item, instead of an empty array

core/lib/multistore.ts
grimasod: Yeah, I removed it from every case :) 
It's easier and simpler without the name prefix and it works correctly in all the cases I tested

core/server-entry.ts
grimasod: That's what I thought at first, but the client has to `registerDynamicRoutes` between createApp and flushRouteQueue

pkarw:
@andrzejewsky pleease take a look at this one by fixing the issue with multistore routes I've raised yesterday on Slack

grimasod:
> i'm not sure about this idea - ... - so the question is what for is this change?

I found inconsistencies causing routes to be registered twice. Specifically in Category and Product Vuex actions, they register mappings with `localizedDispatcherRoute`, which adds a leading slash. It was easier to add the one exception for mappingFallback. And routes need the '/ when adde to the router. But if this causes other problems, or you think it's a bad idea, I can change the catalog actions instead.

grimasod:
We have to do one or the other. Without any change there's a mismatch when the page is reloaded on a category or product page and nothing is loaded

patzick:
@grimasod could you merge the master branch into this branch also? I started to merge master-> develop and I have multiple conflicts around route manager. And you provided more changes here so I think best way would be to merge master here, test everything with QA and merge into develop as a final solution. What do you say? I'll do CR right after merge if you would let me know that it's done :)

grimasod:
All tested...

> * checking if `url/mappingFallback` is not being called (debugger or logger) - to make sure `registerMappings` still works just fine

Tested - mappingFallback only runs when the initial page load is a category or product page

> * checking swithcing the URL between multistore instances (`/de` -> `/it` based urls etc)

Tested - all multistore urls load correctly

> * checking in the console for hydration errors

Tested - there are the same issues on Category and Products pages as the Develop branch

> * checking the `route not found` errors in the console

Tested - no errors reported

> * page not found,

Tested - default, IT, DE, all work correctly

> * if it;s possible to check if the routes are not duplicated (debugger or log)

Tested - both client and server appear correct, with no duplication

> * check the number of `addRoutes` call to make sure this optimization gives us proper value

Tested - addRoutes runs once on server and once on client during initial page load. Then once when each new dynamic route is opened for the first time

grimasod:
@patzick I can try, after we resolve the last issue.

I have merged Develop into this branch every day, with all the other changes that have been made there. I also combined the 1.10.3 router changes into this branch, so hopefully this has all the updates already

patzick:
@grimasod great, that's why I'd like to do this thru your branch as it contains the most accurate final version :)

pkarw:
Hey @grimasod thanks for the tests you executed; looking cool; when it will be ready for merge?

grimasod:
@patzick I've merged Master. I hope I got it right! :)
@pkarw It's ready now. 

I realized the multistore route names do need the store prefix, like "it-home", so they can be customized in the theme route definitions when not using appendStoreCode

pkarw:
Cool! Thanks! @patzick and @andrzejewsky will take care of CR and tests along with @alinadivante - it’s pretty crucial change

alinadivante:
Ok, I've already started testing this branch :) There is the same problem which I reported here #3579 
@andrzejewsky will handle these errors

alinadivante:
There is a small problem with /page-not-found
I see the difference between Server-side and Client-side.. 
go to http://localhost:3000/it and click on `About us (Magento CMS)` link from footer - - the page was not found.
Now refresh http://localhost:3000/it/i/about-us and you will see that pictures from `Visualizza i prodotti più venduti` disappear
image

image

alinadivante:
Second problem regarding differences between server side and client side:
1.  Open http://localhost:3000
2. Log in
3. Go to my profile
4. Log out - I'm redirected to Home page and logged out correctly
now:
1.  Open http://localhost:3000
2. Log in
3. Go to my profile
4. Refresh the page
5. Log out  - I'm redirected to Home page, but I'm not logged out

Issue #3515, opened 2019-09-09
Hash doesn't work in dynamic URLs


lukeromanowicz:
## Current behavior
When entering a dynamic url product/category with hash, it doesn't get resolved correctly in CSR. SSR seems to be working correctly

## Expected behavior
Hash should not impact resolving dynamic URLs

## Steps to reproduce the issue
1. go to any category or product page
2. append hash to url e.g. https://demo.storefrontcloud.io/men/men-11#test

## Can you handle fixing this bug by yourself?

- [x] YES
- [ ] NO

## Which [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html) state this refers to? Info for developer.
Pick one option.

- [ ] This is a bug report for test version on https://test.storefrontcloud.io - In this case Developer should create branch from `develop` branch and create Pull Request `2. Feature / Improvement` back to `develop`.
- [ ] This is a bug report for current Release Candidate version on https://next.storefrontcloud.io - In this case Developer should create branch from `release` branch and create Pull Request `3. Stabilisation fix` back to `release`.
- [x] This is a bug report for current Stable version on https://demo.storefrontcloud.io and should be placed in next stable version hotfix - In this case Developer should create branch from `hotfix` or `master` branch and create Pull Request `4. Hotfix` back to `hotfix`.

PR #3541, opened 2019-09-13
fix invalid route in ButtonOutline

lukeromanowicz:
### Related issues

followup to #3515

### DIFF
--- a/core/lib/multistore.ts
+++ b/core/lib/multistore.ts
@@ -234,7 +234,7 @@ export function localizedRoute (routeObj: LocalizedRoute | string | RouteConfig 
     }
   }
 
-  if (storeCode && config.defaultStoreCode !== storeCode && config.storeViews[storeCode].appendStoreCode) {
+  if (storeCode && config.defaultStoreCode !== storeCode && config.storeViews[storeCode] && config.storeViews[storeCode].appendStoreCode) {
     if (typeof routeObj !== 'object') {
       return '/' + storeCode + routeObj
     }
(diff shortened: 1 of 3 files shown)


core/lib/multistore.ts
```diff
@@ -228,18 +233,18 @@ export function localizedRoute (routeObj: LocalizedRoute | string | RouteConfig
...
-    if (typeof routeObj === 'object') {
-      if (routeObj.name) {
-        routeObj.name = storeCode + '-' + routeObj.name
-      }
 
-      if (routeObj.path) {
-        routeObj.path = '/' + storeCode + '/' + (routeObj.path.startsWith('/') ? routeObj.path.slice(1) : routeObj.path)
-      }
-    } else {
+  if (storeCode && config.defaultStoreCode !== storeCode && config.storeViews[storeCode].appendStoreCode) {
```
patzick: what if somehow `config.storeViews[storeCode]` would not exist?

src/themes/default/components/theme/ButtonOutline.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <component
     :is="link ? 'router-link' : 'button'"
-    :to="localizedRoute(link)"
+    :to="link ? localizedRoute(link) : null"
```
patzick: i believe we shouldn't direct on users such logic; we're returning here the same object, it's not necessary and overcomplication for newcomers.

PR #3568, opened 2019-09-17
Bugfix/multistore routes

pkarw:
### Related issues

It fixes `UrlDispatcher` issues with multistore routes. Few independent issues were introduced from 1.10.1 with the fixes provided for SSR hydration  + `router.addRoutes` optimization. 

There was also a change regarding `prepareStoreView` call which now has been moved before `router.onReady` and therefore we could removed the second call in url dispatcher

Reported issues that this fix is solving (were not raised by Github issues):

**Issue 2 - Default storeview bug**
This is only reproducible on 1.10.2.
When trying to force 404:s I can see the following pattern:https://dearsam.com/se/nonExistingPage => https://dearsam.com/se/page-not-foundhttps://dearsam.com/no/nonExistingPage => https://dearsam.com/page-not-foundhttps://dearsam.com/de/nonExistingPage => https://dearsam.com/page-not-foundhttps://dearsam.com/fr/nonExistingPage => https://dearsam.com/page-not-foundhttps://dearsam.com/en/nonExistingPage => https://dearsam.com/page-not-found

As you can see only the /se site redirects to a localized /se/page-not-found. That means the other sites redirect to /page-no-found and that means the header and footer are not loaded so the visitor can’t navigate away from the 404 page.

I thought it had something to do with that the default storeview in base.json is set to se. That is why it “works” for se. But it seems I was mistaken because this is what is in base.json: "defaultStoreCode": "". Alhough the se storeview is the first one in the storeViews.mapStoreUrlsFor array in base.json.

**Issue 3**
Reproducible on 1.10.2
Context: We have a category in Magento called Posters which has a url_key posters (default storeview). In the country specific storeviews this category has different url_key as described by the table below.Working VSF urls are:https://dearsam.com/se/postershttps://dearsam.com/no/posters

Non working urls are:https://dearsam.com/de/posterhttps://dearsam.com/fr/afficheshttps://dearsam.com/en/printshttps://dearsam.com/us/prints
Possible pattern: In a multistore setup; each localized category url that matches the url_path of the default storeview works.

Magento Storeview      url_key
default posters              se
posters                          no
posters                          de
affiches                         fr
prints                            en
prints                            us

**Issue 4**

Any returning client (with cache from prev. deploy?) has various problems even with the <router-link> links.If I (and others) go to dearsam.com/se on mobile and navigate to ‘menu->posters->alla posters' I do come to the /se/posters page but the resultset is empty. And the Newsletter component in footer is French (Fredrik got German). I’m thinking VSF somehow mixes up which storefront is used.    I cannot reproduce this on desktop/computer. Only on mobile. this tells med this issue has to do with client cache (since mobile cache/app data is way harder to clear than on desktop)

### DIFF
--- a/core/modules/catalog/store/category/actions.ts
+++ b/core/modules/catalog/store/category/actions.ts
@@ -71,7 +71,7 @@ const actions: ActionTree<CategoryState, RootState> = {
         for (let category of resp.items) {
           if (category.url_path && updateState) {
             rootStore.dispatch('url/registerMapping', {
-              url: localizedDispatcherRoute('/' + category.url_path, currentStoreView().storeCode),
+              url: localizedDispatcherRoute(category.url_path, currentStoreView().storeCode),
               routeData: {
                 params: {
                   'slug': category.slug
--- a/core/modules/catalog/store/category/actions.ts
+++ b/core/modules/catalog/store/category/actions.ts
@@ -12,10 +12,11 @@ import { optionLabel } from '../../helpers/optionLabel'
 import RootState from '@vue-storefront/core/types/RootState'
 import CategoryState from '../../types/CategoryState'
 import SearchQuery from '@vue-storefront/core/lib/search/searchQuery'
-import { currentStoreView } from '@vue-storefront/core/lib/multistore'
+import { currentStoreView, localizedDispatcherRoute } from '@vue-storefront/core/lib/multistore'
 import { Logger } from '@vue-storefront/core/lib/logger'
 import { isServer } from '@vue-storefront/core/helpers'
 import config from 'config'
+import { formatCategoryLink } from 'core/modules/url/helpers'
 
 const actions: ActionTree<CategoryState, RootState> = {
   /**
@@ -70,7 +71,7 @@ const actions: ActionTree<CategoryState, RootState> = {
         for (let category of resp.items) {
           if (category.url_path && updateState) {
             rootStore.dispatch('url/registerMapping', {
-              url: category.url_path,
+              url: localizedDispatcherRoute('/' + category.url_path, currentStoreView().storeCode),
               routeData: {
                 params: {
                   'slug': category.slug
--- a/core/modules/url/helpers/index.ts
+++ b/core/modules/url/helpers/index.ts
@@ -19,7 +19,7 @@ function prepareDynamicRoutes (routeData: LocalizedRoute, fullPath: string): Rou
     const currentStoreCode = currentStoreView().storeCode
     const dynamicRouteName = (config.defaultStoreCode !== currentStoreCode) ? `urldispatcher-${fullPath}-${currentStoreView().storeCode}` : `urldispatcher-${fullPath}`
     const dynamicRoute = Object.assign({}, userRoute, routeData, { path: '/' + ((config.defaultStoreCode !== currentStoreCode) ? (currentStoreCode + '/') : '') + fullPath, name: dynamicRouteName })
-    return [dynamicRoute]    
+    return [dynamicRoute]
   } else {
     return null
   }
--- a/core/modules/url/helpers/index.ts
+++ b/core/modules/url/helpers/index.ts
@@ -13,34 +13,20 @@ export function parametrizeRouteData (routeData: LocalizedRoute, query: { [id: s
   return parametrizedRoute
 }
 
-function prepareDynamicRoute (routeData: LocalizedRoute, fullPath: string, addToRoutes: boolean = true): RouteConfig[] {
+function prepareDynamicRoutes (routeData: LocalizedRoute, fullPath: string): RouteConfig[] {
   const userRoute = RouterManager.findByName(routeData.name)
   if (userRoute) {
-    if (addToRoutes) {
-      const routes = []
-      const rootDynamicRoute = Object.assign({}, userRoute, routeData, { path: '/' + fullPath, name: `urldispatcher-${fullPath}` })
-      routes.push(rootDynamicRoute)
-      if (config.storeViews.mapStoreUrlsFor.length > 0 && config.storeViews.multistore === true) {
-        for (let storeCode of config.storeViews.mapStoreUrlsFor) {
-          if (storeCode) {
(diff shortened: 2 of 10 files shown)


core/modules/url/router/beforeEach.ts
```diff
@@ -44,16 +34,16 @@ export async function beforeEach (to: Route, from: Route, next) {
           next(dynamicRoutes[0])
         } else {
           Logger.error('Route not found ' + routeData['name'], 'dispatcher')()
-          next('/page-not-found')
+          next(localizedRoute('/page-not-found', currentStoreView().storeCode))
```
patzick: couldn't we just have a default value for `storeCode` param in `localizedRoute`? There links would look much prettier with just:
```suggestion
          next(localizedRoute('/page-not-found'))
```

core/modules/url/store/actions.ts
```diff
@@ -7,23 +7,24 @@ import queryString from 'query-string'
...
   /**
    * Register dynamic vue-router routes
    */
   async registerDynamicRoutes ({ state, dispatch }) {
     if (state.dispatcherMap) {
-      processMultipleDynamicRoutes(state.dispatcherMap)
+      processMultipleDynamicRoutes(state.dispatcherMap) // check if we're to add routes to vue router
       for (const [url, routeData] of Object.entries(state.dispatcherMap)) {
-        dispatch('registerMapping', { url, routeData })
+        await dispatch('registerMapping', { url, routeData })
```
patzick: maybe `Promise.all` here? Could be faster with larger dispatcherMap object

core/server-entry.ts
```diff
@@ -59,23 +60,21 @@ function _ssrHydrateSubcomponents (components, store, router, resolve, reject, a
 }
 
 export default async context => {
-  const { app, router, store } = await createApp(context, context.vs && context.vs.config ? context.vs.config : buildTimeConfig)
+  let storeCode = context.vs.storeCode
+  if (config.storeViews.multistore === true) {
+    if (storeCode === undefined) { // this is from url
```
patzick:
```suggestion
    if (!storeCode) { // this is from url
```

core/server-entry.ts
```diff
@@ -59,23 +60,21 @@ function _ssrHydrateSubcomponents (components, store, router, resolve, reject, a
...
+      const currentRoute = Object.assign({ path: queryString.parseUrl(context.url).url/* this gets just the url path part */, host: context.server.request.headers.host })
```
patzick: there's no need for Object.assign when you're creating a new object
```suggestion
      const currentRoute = {
        path: queryString.parseUrl(context.url).url, // this gets just the url path part
        host: context.server.request.headers.host
      }
```

core/modules/url/helpers/index.ts
```diff
@@ -13,33 +13,20 @@ export function parametrizeRouteData (routeData: LocalizedRoute, query: { [id: s
...
-    } else {
-      const dynamicRoute = Object.assign({}, userRoute, routeData, { path: '/' + fullPath, name: `urldispatcher-${fullPath}` })
-      return [dynamicRoute]
-    }
+    const currentStoreCode = currentStoreView().storeCode
+    const dynamicRouteName = (config.defaultStoreCode !== currentStoreCode) ? `urldispatcher-${fullPath}-${currentStoreCode}` : `urldispatcher-${fullPath}`
+    const dynamicRoute = Object.assign({}, userRoute, routeData, { path: '/' + fullPath, name: dynamicRouteName })
+    return [dynamicRoute]
   } else {
     return null
```
patzick: please consider returning here an empty array

core/modules/catalog/store/category/actions.ts
```diff
@@ -70,7 +71,7 @@ const actions: ActionTree<CategoryState, RootState> = {
         for (let category of resp.items) {
           if (category.url_path && updateState) {
             rootStore.dispatch('url/registerMapping', {
-              url: category.url_path,
+              url: localizedDispatcherRoute('/' + category.url_path, currentStoreView().storeCode),
```
patzick: maybe localizedDispatcherRoute could detect slash at the beginning by itself? It reduces complexity.

And also storeCode as default

core/modules/url/store/actions.ts
pkarw: OK

core/modules/url/helpers/index.ts
pkarw: Done

core/modules/catalog/store/category/actions.ts
pkarw: OK

benjick:
Works great for me!

alinadivante:
- What about "Trailing slash" issue?
http://localhost:3000/de/men/herren-11 works
http://localhost:3000/de/men/herren-11/ doesn't work
http://localhost:3000/de/c/herren-11 works
http://localhost:3000/de/c/herren-11/ works

- **ad. issue 2:**  What about 404 HTTP code for 404 pages? 
This was fixed by https://github.com/DivanteLtd/vue-storefront/pull/3296 but now I have code 200 OK for page-not-found. 

- **ad. issue 3:** I didn't have problems with category URLs, but I noticed that "Training" category is missing from the `/de` and `/it` menu, although I can get to it via the link http://localhost:3000/de/training/training-video/video-download-10

pkarw:
Thanks @alinadivante!
1. Will backport the trailing slash fix to 1.10
2. Yeah, it was fixed just in 1.11 we can’t backport this feature; too heavy
3. I’ll check it

pkarw:
1: fixed

alinadivante:
ad.1 thanks, it works 👍 
ad. 2 ok, I understood 
ad 3. are u sure? Your screenshot looks like the default store (US) and there are all categories in the menu, but in /de and /it stores there is no "Traning" category with videos http://localhost:3000/it/training/training-video/video-download-10

pkarw:
This fix was ported to 1.11 in the: https://github.com/DivanteLtd/vue-storefront/pull/3540 (which may cause conflicts)

PR #3586, opened 2019-09-20
filterMinimumShouldMatch(1) to query in order to support ES7

### LINKED ISSUE #1692: Support for Elastic 7.x
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

* **What is the current behavior?**

Currently we’re storing multiple document types (product, attribute, categories, taxrules) in single index (vue_storefront_catalog by default). ES 6.x is no longer supporting this setup and requires us to have:
- single index per entity type (A)
- custom type field (B)

https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html

I believe that we should follow the A (looking forward to Your feedback!) using the  following naming convention for indexes:
- vue_storefront_catalog_product
- vue_storefront_catalog_category
- ...

We should avoid as many changes required on the users as we could.

1. Changes in the config
The index name should be changed just before reaching the ES  - ** without requiring the config changes** (eg. we have `vue_storefront_catalog` in the config so we can just add the `entityType` to `indexName` in here: https://github.com/DivanteLtd/vue-storefront-api/blob/e0aa58460f6a986e2527f364946fc751620d7428/src/api/catalog.js#L55) 

2. Automatic migration
We can base on `apiVersion` ( https://github.com/DivanteLtd/vue-storefront-api/blob/e0aa58460f6a986e2527f364946fc751620d7428/config/default.json#L33) and check - if there is the `6.x` selected we should run the automatic migration in `vue-storefront-api` + placing a lock file like `.es.migrated` - to let the script know no further migration is required.

It can be a separate migration script or just another migration (then no lock file is required) -> https://github.com/DivanteLtd/vue-storefront-api/tree/master/migrations

3. Front-end 
This change should be TRANSPARENT to the frontend VS - with no changes required (both config, and code)

To do so we must refactor:
- [ ] vue-storefront-api/src/scripts/db.js
- [ ] vue-storefront-api/migrations
- [ ] mage2vuestorefront
- [ ] vue-storefront/core/store/lib/search.js
- [ ] vue-storefront-api/package.js to modify the restore and migrate scripts
- [ ] docker files
- [ ] add migration scripts to reindex existing indexes to the new format

Later on:
- [ ] https://github.com/DivanteLtd/magento1-vsbridge-indexer
- [ ] https://github.com/DivanteLtd/magento2-vsbridge-indexer

* **What is the expected behavior?**

We should have the ES 6.x supported by default

* **What is the motivation / use case for changing the behavior?**

The main motivation is that we re bound to 5.x

pkarw:
### Related issues

closes #1692 

Related to: https://github.com/DivanteLtd/vue-storefront-api/pull/342

### DIFF
--- a/core/lib/search/adapter/api/elasticsearchQuery.js
+++ b/core/lib/search/adapter/api/elasticsearchQuery.js
@@ -68,7 +68,7 @@ export async function prepareElasticsearchQueryBody (searchQuery) {
     }
 
     if (hasCatalogFilters) {
-      query = query.filterMinimumShouldMatch(1).orFilter('bool', (b) => attrFilterBuilder(b))
+      query = query.filterMinimumShouldMatch(1).orFilter('bool', attrFilterBuilder)
         .orFilter('bool', (b) => attrFilterBuilder(b, optionsPrefix).filter('match', 'type_id', 'configurable')) // the queries can vary based on the product type
     }
   }
--- a/core/lib/search/adapter/api/elasticsearchQuery.js
+++ b/core/lib/search/adapter/api/elasticsearchQuery.js
@@ -68,7 +68,7 @@ export async function prepareElasticsearchQueryBody (searchQuery) {
     }
 
     if (hasCatalogFilters) {
-      query = query.orFilter('bool', (b) => attrFilterBuilder(b))
+      query = query.filterMinimumShouldMatch(1).orFilter('bool', (b) => attrFilterBuilder(b))
         .orFilter('bool', (b) => attrFilterBuilder(b, optionsPrefix).filter('match', 'type_id', 'configurable')) // the queries can vary based on the product type
     }
   }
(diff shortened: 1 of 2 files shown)


core/lib/search/adapter/api/elasticsearchQuery.js
```diff
@@ -68,7 +68,7 @@ export async function prepareElasticsearchQueryBody (searchQuery) {
     }
 
     if (hasCatalogFilters) {
-      query = query.orFilter('bool', (b) => attrFilterBuilder(b))
+      query = query.filterMinimumShouldMatch(1).orFilter('bool', (b) => attrFilterBuilder(b))
```
andrzejewsky: `filterMinimumShouldMatch(1).orFilter('bool', attrFilterBuilder)`

pkarw: thx; applied

PR #3599, opened 2019-09-22
Fix loading bestsellers

### LINKED ISSUE #3454: Optimize the `RouterManager.addRoutes`

andrzejewsky:
### Related issues
closes #3454

It solves the problem with disappearing bestsellers mentioned in https://github.com/DivanteLtd/vue-storefront/pull/3540


core/pages/PageNotFound.js
```diff
@@ -1,32 +0,0 @@
-import { prepareQuery } from '@vue-storefront/core/modules/catalog/queries/common'
```
patzick: we shouldn't remove this file now, leave it for backward compatibility like the rest of core pages, but just remove dependency from PageNotFound.vue

src/themes/default/store/homepage.ts
```diff
@@ -19,16 +20,31 @@ export const homepageStore = {
         { products: newProductsResult.items
         }, { root: true })
       commit('SET_NEW_COLLECTION', configuredProducts)
+    },
+    async fetchBestsellers ({ commit, dispatch }) {
```
patzick: as we established some time ago, when we fetch and change state, then methods should be called `load...` when just fetching and not mutating state, then `fetch...`
```suggestion
    async loadBestsellers ({ commit, dispatch }) {
```

Issue #3597, opened 2019-09-21
Problem with custom options


vue-kacper:
## Current behavior
I'm unable to add bundle type product to cart.

## Expected behavior
I should be able to see bundle type product price and add it to cart. 

## Steps to reproduce the issue

Visit: https://test.storefrontcloud.io/gear/gear-3/sprite-yoga-companion-kit-45.html
Try to add the product to cart.

## Which [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html) state this refers to? Info for developer.
Pick one option.

- [x] This is a bug report for test version on https://test.storefrontcloud.io - In this case Developer should create branch from `develop` branch and create Pull Request `2. Feature / Improvement` back to `develop`.
- [ ] This is a bug report for current Release Candidate version on https://next.storefrontcloud.io - In this case Developer should create branch from `release` branch and create Pull Request `3. Stabilisation fix` back to `release`.
- [ ] This is a bug report for current Stable version on https://demo.storefrontcloud.io and should be placed in next stable version hotfix - In this case Developer should create branch from `hotfix` or `master` branch and create Pull Request `4. Hotfix` back to `hotfix`.

## Additional information

There are minor differences in API response for this product between "test." and "demo.":

https://test.storefrontcloud.io/api/catalog/vue_storefront_catalog/product/_search?q=sku:24-WG080

https://demo.storefrontcloud.io/api/catalog/vue_storefront_catalog/product/_search?q=sku:24-WG080

vue-kacper:
Related issue:

When I'm adding a bundle-type product to wishlist from the catalog page, in wishlist sidebar I can see the price of $0

![image](https://user-images.githubusercontent.com/42430550/65375308-355bf400-dc94-11e9-9c80-6f80eeb8ee11.png)

But from product details page - removing this bundle-type product from wishlist and adding it again will output different price in wishlist sidebar:

![image](https://user-images.githubusercontent.com/42430550/65375338-8cfa5f80-dc94-11e9-8473-b3de2c455faf.png)

pkarw:
Related to: #3093 (last comment). @andrzejewsky or @patzick could you please fix it?

pkarw:
I can confim the issue with product options seems to be fixed - it now works for me both for wislists + cart


BUT there is another problem with custom options - I can't add the product to cart

Product with custom options:
http://localhost:3000/gear/gear-3/wayfarer-messenger-bag-custom-options-4.html

Error:

[...]

Issue #3387, opened 2019-08-20
Add product related hooks


pkarw:
## What is the motivation for adding/enhancing this feature?

It would be great to add an elegant set of hooks related to the products to make the 3rd party integrations (like hotjar, GTM, analytics) easier.

Take a look at how guys are [doing GTM integration right now](https://github.com/DivanteLtd/vue-storefront/blob/master/src/modules/google-tag-manager/hooks/afterRegistration.ts). It requires you to dig really deeply into src code to make sure which mutations are doing what etc.

Can we add the following hooks to the core:
- product page before/after displayed (that allows the user to modify the product),
- category page before/after displayed - the category should be passed as a parameter + potentially the list of products,
- the order has been placed,
- the order is about to be placed (mutator letting the user mutate the order),

pkarw:
Related: #3205

pkarw:
Related: #3217

pkarw:
@filrak if you won't find a time to implement this feature then @andrzejewsky please take care of it - we've got a deadline set for 19 th of September (the release date of 1.11-rc1)

filrak:
The order-related ones are already implemented

filrak:
Also what would be a use case to modify the product? Can't we do this just by modifying the searchQuery?

PR #3600, opened 2019-09-22
Bugfix/qa fixes

### LINKED ISSUE #3093: Product page refactoring for theme 2.0

patzick:
### Related issues

closes #3093, #3387

No changelog is required, all features are already added, these are QA improvements

- save scroll position on same page (ex change filters on category page)
- productPageVisited hook
- fix labels on product page
- fixed blocked addToCart button in offline mode
- fixed blocked addToCart button for bundle product
- changed duplicated mutation name


src/themes/default/pages/Product.vue
```diff
@@ -350,17 +352,24 @@ export default {
     getInputName () {
       if (this.isSimpleOrConfigurable && !this.isProductLoading) { return this.$i18n.t('Quantity available', { qty: this.quantity }) }
       return this.$i18n.t('Quantity')
+    },
+    isAddToCartDisabled () {
+      return (!this.$v.getCurrentProduct.qty.minValue || !this.$v.getCurrentProduct.qty.maxValue || !this.$v.getCurrentProduct.qty.numeric) || (!this.quantity && this.isSimpleOrConfigurable && !this.isProductLoading)
```
andrzejewsky: maybe use here few regular `if` statements because this one inline if is a bit hard to read.

patzick: sure, I just moved it here from Html, but I think it could also be something like
```suggestion
      return this.$v.$invalid || (!this.quantity && this.isSimpleOrConfigurable && !this.isProductLoading)
```
I'll short it

patzick: done :)

src/themes/default/pages/Product.vue
```diff
@@ -350,17 +352,24 @@ export default {
...
     }
   },
   created () {
     this.getQuantity()
   },
-  mounted () {
-    this.$store.dispatch('recently-viewed/addItem', this.getCurrentProduct)
+  async mounted () {
+    await this.$store.dispatch('recently-viewed/addItem', this.getCurrentProduct)
+    await this.$store.dispatch('attribute/list', { // load filter attributes for this specific category
```
pkarw: Wasn’t it called in product/fetch? I bet it was :) wondering why there’s a need to call it in here? More over mounted() is not being called while clicking between recommended products (not switching the routes)

patzick:
fixes #3597 also

PR #3630, opened 2019-09-26
perPage load with value from local.json

ngongoll:
i put the perPage value in the config file to be more flexible. i have only 3 columns in my categories and loading 50 perPage return uneven rows.


andrzejewsky:
@pkarw basically it's a theme thing, I wouldn't put options like this to the config file. When you create a theme you calling/dispatching this action and you can decide how many items do you want to display by arguments passed to the action. What do you think @pkarw @filrak ?

filrak:
Having theme.config.js in a theme root could be very useful indeed ;)

PR #3651, opened 2019-09-28
feat: moved trace logic to a server.onLoad hook

ResuBaka:
### Related issues

closes #


ResuBaka:
hey @filrak you mean something like that? 

When yes I would polish it a bit more and maybe move the whole serverModules logic to a hook to so you can just have different exports on like modules two for the server then. One that is loaded pretty early like now the tracing stuff needs to be and the second one for added routes or middlewares to the server.

ResuBaka:
@pkarw 

I have moved the whole extend logic to a hook which takes the `app`, `config.server` and `isProd` as an object as a parameter. 
Then I have changed the first hook to a be named trace hook so you know that is should be used to add your tracing library.

pkarw:
@ResuBaka would be awesome to add some additional hooks by this chance:
- afterOutputRendered - so the users can adjust req/res/context doing some filtering etc - by having this we could have a Munich plugin not in the core as we currently have but asa separate module 
- beforeOutputRendered - passing the request context that can be modified / prepared + res / req so the users can adjust the headers or the context  - we can even do something like if the hook will return any continent / set any content or pass some flag or do req.end we can skip running ssr - I mean the way to use this hook to implement custom caching algorithm
- beforeCacheInvalidated

ResuBaka:
I have added the new hooks but that is only a basic implementation as I am not 100% sure how to add each of them in the best way. 

So it would be great when I get there some feedback on what they should be able to do.

filrak:
@pkarw at this point before adding those hooks n89mI want you to decide if we want to be server-comaptible with nuxt in terms of hooks. The ones you proposed cannot be translated to other frameworks therefore will break **potentially** all server side modules. 

Are we okay with that? I think if you see urgent need for them now we can be ok with that because we will drop support for those modules but have nuxt ones in return.

If you don't see urgent need for this hooks I would advice to wait with adding  (any) new APIs until they needed.

filrak:
@ResuBaka I'll wait with a. review till @pkarw will answer because it takes into account function signatures as well

core/scripts/server.ts
```diff
@@ -71,6 +72,9 @@ function invalidateCache (req, res) {
         tags = req.query.tag.split(',')
       }
       const subPromises = []
+
+      serverHooksExecutors.beforeCacheInvalidated()
```
pkarw: let's pass the all parameters in here nad it's fine

core/scripts/server.ts
```diff
@@ -186,6 +202,15 @@ app.get('*', (req, res, next) => {
           tagsArray
         ).catch(errorHandler)
       }
+
+      const hookResponse2 = serverHooksExecutors.afterOutputRendered({
```
pkarw: Rename to `afterOutputRenderedResponse`

core/scripts/server.ts
```diff
@@ -178,6 +185,15 @@ app.get('*', (req, res, next) => {
         res.setHeader('X-VS-Cache-Tags', cacheTags)
         console.log(`cache tags for the request: ${cacheTags}`)
       }
+
+      const hookResponse = serverHooksExecutors.beforeOutputRendered({
```
pkarw: Rename to `beforeOutputRenderedResponse`

core/scripts/server.ts
```diff
@@ -186,6 +202,15 @@ app.get('*', (req, res, next) => {
...
+
+      const hookResponse2 = serverHooksExecutors.afterOutputRendered({
+        req,
+        res,
+        context,
+        output,
+        isProd
+      })
+
       res.end(output)
```
pkarw: if `afterOutputRenderedResponse` !== null it should be then used instead of `output`

ResuBaka: I have used this logic here: 

```typescript
      if (typeof afterOutputRenderedResponse.output === 'string') {
        res.end(afterOutputRenderedResponse.output)
      } else if (typeof afterOutputRenderedResponse === 'string') {
        res.end(afterOutputRenderedResponse)
      } else {
        res.end(output)
      }
```

As this should be more what we want I think.

ResuBaka:
This PR should not be merged to fast. 

As we should exactly know what each hook should do. As the MutatorHook I think needs a bit more thinking if there are right or not

andrzejewsky:
@ResuBaka is it ready to be merged in? Are you still working on this?

ResuBaka:
@andrzejewsky it should work. I have not tested it, I only have written it after the logic of how each hook should work like from there internal logic.  

So I would say that before we merge it there should be done a bit of QA of each hook so it we are 100% sure that it all works like every buddy would expect this works.

PR #3657, opened 2019-09-30
Fixed bundle products with no options

ResuBaka:
### Related issues

closes #

When bundle products have no options they can have the same hash as a normal
product so I just hash the whole product which fixes the issue. 

Why the has can be the same? 
The hash can be the same as a normal product that is because the product_option are the same from a bundle product to a simple product.

### DIFF
--- a/core/modules/cart/helpers/productChecksum.ts
+++ b/core/modules/cart/helpers/productChecksum.ts
@@ -1,22 +1,7 @@
 import CartItem from '@vue-storefront/core/modules/cart/types/CartItem'
 import { sha3_224 } from 'js-sha3'
 
-const getDataToHash = (product: CartItem): any => {
-  if (!product.product_option) {
-    return null
-  }
-
-  const { extension_attributes } = product.product_option
-
-  if (extension_attributes.bundle_options) {
-    const { bundle_options } = extension_attributes
-    return Array.isArray(bundle_options) ? bundle_options : Object.values(bundle_options)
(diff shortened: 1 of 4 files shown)


core/modules/cart/helpers/productChecksum.ts
```diff
@@ -1,22 +1,7 @@
...
-    const { bundle_options } = extension_attributes
-    return Array.isArray(bundle_options) ? bundle_options : Object.values(bundle_options)
-  }
-
-  return product.product_option
-}
-
 const productChecksum = (product: CartItem): string =>
-  sha3_224(JSON.stringify(getDataToHash(product)))
+  sha3_224(JSON.stringify(product))
```
pkarw: I'm not sure if it this won't mess with other types of products (configurable etc) or situations we're setting additional properties on products. Vue Storefront is sometimes setting `product.error` another property; two same products with just differences error properties will be considered different products. What's the key difference between these products you mentioned? Probably we should just add `type_id` (`simple` vs. `bundle`) to the hash shouldn't we? Maybe also `product_id` and `sku`

ResuBaka: Changed it. 

Now I use the sku and the product_option for the hash which fixes he issue too.

ResuBaka:
@patzick that was when I fixed the test. I have now renamed the test to its old name so it should be clear now what it does.

PR #3672, opened 2019-10-01
Get rid of annoying i18n warn

oskar1233:
### Related issues
None I'm aware of.

### Short description
Get rid of annoying warning about falling back from per-component i18n message to global message.

> To suppress these warnings (while keeping those which warn of the total absence of translation for the given key) set silentFallbackWarn: true when initializing the VueI18n instance.
https://kazupon.github.io/vue-i18n/guide/fallback.html

I'd go for stable as it's tiny PR.

No upgrade steps required.


andrzejewsky:
lack of activity, closing. @oskar1233 please comment if you want to still work on it

PR #3691, opened 2019-10-07
Add catalog filters, Fix breadcrumb issues

grimasod:

1. Fixes breadcrumbs for products that are in multiple categories, in different branches of the category tree

2. Uses breadcrumbs module store, so breadcrumb categories are only calculated once

3. Adds an optional filter of breadcrumb categories to local.json

4. Also adds pre-defined category search query filters in local.json that are applied to every query. This removes the need to define more and more parameters for category/list calls, but still allows parameters to override the pre-defined filters. It also means the currently hardcoded platform-specific parameters, like "is_active" and "product_count", could be deprecated.


src/themes/default/components/core/Breadcrumbs.vue
```diff
@@ -12,41 +12,9 @@
 </template>
 
 <script>
-import { localizedRoute, currentStoreView } from '@vue-storefront/core/lib/multistore'
-import i18n from '@vue-storefront/i18n'
+import { Breadcrumbs } from '@vue-storefront/core/modules/breadcrumbs/components/Breadcrumbs.ts'
 
 export default {
-  computed: {
```
andrzejewsky: Why did you move this logic to the mixin? breadcrumb it's a theme thing, this is the best place for it, we even try to eliminate core-components as much as it's possible, is there any reason?

grimasod: From a theme developer point of view, it's ideal to have logic and basic functionality in the core component mixins. Then the theme components are primarily for the presentation layer. Custom themes can easily incorporate the functionality and can still override anything.

For me, the list of routes used by the breadcrumbs is functional. How they are displayed is presentational and therefore in the theme.  This seemed to be the standard approach, at least until v1.10

> we even try to eliminate core-components as much as it's possible

Is this a new policy from v1.11?

Honestly, it will require so much more work for every custom theme to be updated with new features and bug fixes. It's very complicated to merge / copy changes from one component (default theme) into another (custom theme). It also results in a lot more repeated code, copied into every custom theme.

But if it's actually the new policy, of course I can easily return computed data and props back to the theme component. No problem.

andrzejewsky: @grimasod Ok I see your point. Basically we try to move mixins/core components to the theme (if it makes sense, for sure we don't create new ones) but it doesn't mean that we delete them at all - always we mark them as deprecated. 

The reason of that is mixins causes problems such as name collision or they are hard do debug. Moreover, you have almost everything in the vuex store, so you can use actions/getters to create eg. your own breadcrumb or whatever you like - this is what you should use in your themes, it's much easier to maintain. 

Ok I think for now we can keep it as it is, you don't have to move it anywhere.

Honestly, based on your comment I'm considering to keep using mixins but only really small ones including really basic/default things such as that breadcrumb (it will be even same in most of themes). If that make theme developers lives easier - why not, but for sure we'll be continuing deprecating these complex ones that for sure won't be used by potential developer. (@patzick what do you think?)

ArturDivante:
@andrzejewsky @grimasod Test instructions / suggestion needed here :)

andrzejewsky:
@ArturDivante so, first of all:
1. Check breadcrumbs across entire app, especially on product page that has assigned to more that one category
2. Check if all of categories are loaded correctly
4. Test new options in the config file `entities.category.filterFields`  `entities.category.breadcrumbFilterFields` that are filtering these elements by given field

You can also, text me on slack if you want more details

grimasod:
Here's one way to test the new config options.

#### To prepare the catalog:
- Create a deep-level category, for example `Default category / Sales / October / Week 2 / Friday / Women / Tops / Tees`. (This is because the current breadcrumbs gives preference to deeper paths and we want to compare to that)
- In this category, have `Include in Menu` turned off
- Add a Product to the new Category, say "Radiant Tee" SKU: WS12-XS-BLUE, which is already in an existing Category `Women / Tops / Tees` where `Include in Menu` is turned on
- Update / Reindex Elastic Search

#### To test `entities.category.filterFields`
- Navigate to the new category, to confirm it's appearing and contains the product added above
- set `"filterFields": { "include_in_menu": true }` in local.json and rebuild
- Reload and check that the category will not be accessible now. (Please note, if the side menu has still not been updated to use `category-next`, the category will still appear in menu. This is a separate issue)
- You can also check in Vuex that the category is not included
- remove `"filterFields": { "include_in_menu": true }` in local.json, to be ready for the next test

#### To test `entities.category.breadcrumbFilterFields `
- navigate to the Radiant Tee product page, using the side menu: Women -> Tops -> Tees -> Radiant Tee
- Note the breadcrumbs - they should show the deeper path created above, or maybe none
- Reload the product page and observe the breadcrumbs again
- Now set `"breadcrumbFilterFields": { "include_in_menu": true }` in local.json and rebuild
- Reload the product page and check that the breadcrumbs show the standard path `Women / Tops / Tees`
- Navigate away and back to this product using the menu, the breadcrumbs should be consistent

 I've tested using our real catalog data, but not with the test data

alinadivante:
@ArturDivante  Please take a look at this and test it in your free time:)

grimasod:
@andrzejewsky I see this PR didn't make it into 1.11.0-rc2. Will there be an rc3? Thanks

andrzejewsky:
@grimasod We want to merge it in `1.11`, sorry for that long time.

andrzejewsky:
Hey @grimasod seems like everything works fine, but one thing that must be changed here... Could you add a new options that you've introduced to the `default.json` and set them default value? probably `filterFields` in both cases will be just empty (`{}`) (it also must be covered in the code)

grimasod:
Hi @andrzejewsky no problem, added to default.json :) No code changes.

core/modules/catalog-next/store/category/actions.ts
```diff
@@ -127,20 +128,27 @@ const actions: ActionTree<CategoryState, RootState> = {
...
+    if (searchingByIds) { // removing from search query already loaded categories, they are added to returned results
+      for (const [categoryId, category] of Object.entries(getters.getCategoriesMap)) {
+        if (searchedIds.includes(categoryId)) {
+          loadedCategories.push(category as Category)
+        }
+      }
       categorySearchOptions.filters.id = searchedIds.filter(categoryId => !getters.getCategoriesMap[categoryId] && !getters.getNotFoundCategoryIds.includes(categoryId))
     }
     if (!searchingByIds || categorySearchOptions.filters.id.length) {
+      categorySearchOptions.filters = Object.assign({}, config.entities.category.filterFields ? config.entities.category.filterFields : {}, categorySearchOptions.filters)
```
andrzejewsky:
```suggestion
      categorySearchOptions.filters = Object.assign(config.entities.category.filterFields, categorySearchOptions.filters)
```
because `config.entities.category.filterFields` has `{}` by default

core/modules/catalog/store/product/actions.ts
```diff
@@ -680,12 +680,15 @@ const actions: ActionTree<ProductState, RootState> = {
...
-  async loadProductBreadcrumbs ({ dispatch }, { product } = {}) {
+  async loadProductBreadcrumbs ({ dispatch, rootGetters }, { product } = {}) {
     if (product && product.category_ids) {
-      const categoryFilters = { 'id': product.category_ids }
-      const categories = await dispatch('category-next/loadCategories', {filters: categoryFilters}, { root: true })
-      const deepestCategory = categories.sort((a, b) => (a.level > b.level) ? -1 : 1)[0] // sort starting by deepest level
-      await dispatch('category-next/loadCategoryBreadcrumbs', deepestCategory, { root: true })
+      let currentCategory = rootGetters['category-next/getCurrentCategory'] // use current category, if set
+      if (!currentCategory || !currentCategory.id || !product.category_ids.includes(currentCategory.id.toString())) {
+        const categoryFilters = Object.assign({ 'id': product.category_ids }, config.entities.category.breadcrumbFilterFields ? config.entities.category.breadcrumbFilterFields : {})
```
andrzejewsky:
```suggestion
        const categoryFilters = Object.assign({ 'id': product.category_ids }, config.entities.category.breadcrumbFilterFields)
```

because of `config.entities.category.breadcrumbFilterFields` default value is `{}`

PR #3692, opened 2019-10-08
Fixed a three problems that could happen when you have bundle products

ResuBaka:
### Related issues

closes #

1. Is that the optionChanged event does not work because camel case does not work 100% of the time.
2. Bundles with only one option for the bundle option were not working as the function find does not exist on an object.
3. This can only happen in an bundle or config product and that is that there is no server_item_id or item_id in the products that are going to be compared and then they where equal which is wrong.

All three cases above are fix with this PR.

### DIFF
--- a/core/modules/cart/helpers/productsEquals.ts
+++ b/core/modules/cart/helpers/productsEquals.ts
@@ -15,8 +15,12 @@ const getProductType = (product: CartItem): string =>
 const getServerItemId = (product: CartItem): string | number =>
   product.server_item_id || product.item_id
 
-const isServerIdsEquals = (product1: CartItem, product2: CartItem): boolean =>
-  getServerItemId(product1) === getServerItemId(product2)
+const isServerIdsEquals = (product1: CartItem, product2: CartItem): boolean => {
+  if (getServerItemId(product1) === undefined || getServerItemId(product2) === undefined) {
+    return false
+  }
+  return getServerItemId(product1) === getServerItemId(product2)
+}
 
 const isChecksumEquals = (product1: CartItem, product2: CartItem): boolean =>
   getChecksum(product1) === getChecksum(product2)
--- a/core/modules/cart/helpers/productsEquals.ts
+++ b/core/modules/cart/helpers/productsEquals.ts
@@ -16,10 +16,11 @@ const getServerItemId = (product: CartItem): string | number =>
   product.server_item_id || product.item_id
 
 const isServerIdsEquals = (product1: CartItem, product2: CartItem): boolean => {
-  if (getServerItemId(product1) === undefined || getServerItemId(product2) === undefined) {
-    return false
-  }
-  return getServerItemId(product1) === getServerItemId(product2)
+  const product1ItemId = getServerItemId(product1)
+  const product2ItemId = getServerItemId(product2)
+  const areItemIdsDefined = product1ItemId !== undefined && product2ItemId !== undefined
+
+  return areItemIdsDefined && product1ItemId === product2ItemId
 }
 
 const isChecksumEquals = (product1: CartItem, product2: CartItem): boolean =>
--- a/core/modules/catalog/components/ProductBundleOption.ts
+++ b/core/modules/catalog/components/ProductBundleOption.ts
@@ -29,7 +29,10 @@ export const ProductBundleOption = {
       return `bundleOptionQty_${this.option.option_id}`
     },
     value () {
-      return this.option.product_links.find(product => product.id === this.productOptionId)
+      if (Array.isArray(this.option.product_links)) {
+        return this.option.product_links.find(product => product.id === this.productOptionId)
+      }
+      return this.option.product_links
     },
     errorMessage () {
       return this.errorMessages ? this.errorMessages[this.quantityName] : ''
@@ -57,7 +60,12 @@ export const ProductBundleOption = {
   methods: {
     setDefaultValues () {
       if (this.option.product_links) {
-        const defaultOption = this.option.product_links.find(pl => { return pl.is_default })
+        let defaultOption
+        if (Array.isArray(this.option.product_links)) {
+          defaultOption = this.option.product_links.find(pl => { return pl.is_default })
+        } else {
+          defaultOption = this.option.product_links
+        }
         this.productOptionId = defaultOption ? defaultOption.id : this.option.product_links[0].id
         this.quantity = defaultOption ? defaultOption.qty : 1
       }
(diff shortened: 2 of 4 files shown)


src/themes/default/components/core/ProductBundleOptions.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <form class="custom-options">
     <div v-for="option in product.bundle_options" :key="('bundleOption_' + option.option_id)">
-      <product-bundle-option :option="option" @optionChanged="optionChanged" :error-messages="errorMessages" />
+      <product-bundle-option :option="option" @option-changed="optionChanged" :error-messages="errorMessages" />
     </div>
```
pkarw: Shouldn’t it be “bundleOptionChanged” as for the function name?

core/modules/cart/helpers/productsEquals.ts
```diff
@@ -15,8 +15,12 @@ const getProductType = (product: CartItem): string =>
 const getServerItemId = (product: CartItem): string | number =>
   product.server_item_id || product.item_id
 
-const isServerIdsEquals = (product1: CartItem, product2: CartItem): boolean =>
-  getServerItemId(product1) === getServerItemId(product2)
+const isServerIdsEquals = (product1: CartItem, product2: CartItem): boolean => {
+  if (getServerItemId(product1) === undefined || getServerItemId(product2) === undefined) {
+    return false
```
pkarw: What if both products have empty server_item_id (say both products added in the offline mode and not yet synced); it’s probably from the caller context that it should be false
In this case correct?

core/modules/cart/helpers/productsEquals.ts
```diff
@@ -15,8 +15,12 @@ const getProductType = (product: CartItem): string =>
 const getServerItemId = (product: CartItem): string | number =>
   product.server_item_id || product.item_id
 
-const isServerIdsEquals = (product1: CartItem, product2: CartItem): boolean =>
-  getServerItemId(product1) === getServerItemId(product2)
+const isServerIdsEquals = (product1: CartItem, product2: CartItem): boolean => {
+  if (getServerItemId(product1) === undefined || getServerItemId(product2) === undefined) {
```
andrzejewsky:
```suggestion
  const product1ItemId = getServerItemId(product1)
  const product2ItemId = getServerItemId(product2)
  const areItemIdsDefined = product1ItemId !== undefined && product2ItemId !== undefined
 
 return areItemIdsDefined && product1ItemId === product2ItemId
```

core/modules/catalog/components/ProductBundleOption.ts
```diff
@@ -29,7 +29,10 @@ export const ProductBundleOption = {
       return `bundleOptionQty_${this.option.option_id}`
     },
     value () {
-      return this.option.product_links.find(product => product.id === this.productOptionId)
+      if (Array.isArray(this.option.product_links)) {
```
andrzejewsky:
```suggestion
      const { product_links } = this.option
      if (Array.isArray(product_links)) {
        return product_links.find(product => product.id === this.productOptionId)
      }
      
      return product_links
```

core/modules/catalog/components/ProductBundleOption.ts
```diff
@@ -57,13 +60,18 @@ export const ProductBundleOption = {
   methods: {
     setDefaultValues () {
       if (this.option.product_links) {
-        const defaultOption = this.option.product_links.find(pl => { return pl.is_default })
+        let defaultOption
+        if (Array.isArray(this.option.product_links)) {
+          defaultOption = this.option.product_links.find(pl => { return pl.is_default })
+        } else {
+          defaultOption = this.option.product_links
+        }
```
andrzejewsky:
```suggestion
      const { product_links } = this.option
      const defaultOption = Array.isArray(product_links) ? product_links.find(pl => pl.is_default) : product_links
```

src/themes/default/components/core/ProductBundleOptions.vue
ResuBaka: I think that this is fine as it is. As it only Happens in files that have bundle in the name so it should be clear what is happening here.

When you think it is not so clear we can change it.

core/modules/cart/helpers/productsEquals.ts
ResuBaka: I should always be false when one of the products has no server_item_id or item_id other wise when both don't have it they are the same product but aren't in most cases.

core/modules/catalog/components/ProductBundleOption.ts
ResuBaka: Normally I would write it like this

```typescript
const defaultOption = Array.isArray(product_links) ? product_links.find(pl => {
          return pl.is_default
        }) : product_links
```

as I try to always be under the 120 length in a line. 

What do you think about that?

core/modules/cart/helpers/productsEquals.ts
ResuBaka: Done

core/modules/catalog/components/ProductBundleOption.ts
ResuBaka: Done

core/modules/cart/helpers/productsEquals.ts
ResuBaka: But for that case we have the check of the checksum.

core/modules/catalog/components/ProductBundleOption.ts
andrzejewsky: yeah it's ok, but you can break line in the `inline if` just after question mark or colon eg:
```
condition ? // wrap line
 true : // wrap line
 false
```

andrzejewsky: also you don't need brackets here:

```
product_links.find(pl => { return pl.is_default }) -> product_links.find(pl =>pl.is_default)
```

ResuBaka: I have changed it I can do it 100% like you want it because of eslint :smile:

andrzejewsky: @ResuBaka ESlint always knows better how to make you happier 😆

andrzejewsky:
@ResuBaka could you please add a short info how to test / reproduce bug you mentioned?  - this is for understanding for QA team.

ResuBaka:
As it is already merged I still write in here how you could have run into some of the problems.


The Event name change is that in our theme it has not worked I didn't under stand why. So I changed it to more like how actual dom event are named as there are no uppercase in them allowed as far as I am aware of it. 

Then the change in the for checking if an bundle option has multible options is when each or one bundle option has only one option in it.

The change for if an product has an server_item_id or item_id is there when it happens that both are not set for what ever reason in both product both are undefined that it would be true which is wrong. This is hard to force I thing as I don't even know what that is happen for us but as it is an edge case it is still good to have it fixed.

PR #3837, opened 2019-11-24
Add Intl.js polyfill support

### LINKED ISSUE #3836: Lack of complete `Intl.NumberFormat()`/`toLocaleString()` support in NodeJs
## Current behavior

This might be connected to #3553

If you use a Node version which not yet contains the `full-icu` package by default like `v10-LTS`, the numbers aren't formatted correctly on server-side which leads into a flickering of prices on initial page-load.

With version `v13.2.0` of node they start to implement `--with-intl=full-icu` as default, which contains all languages for nodes server-side localization. Before that version its only `--with-intl=small-icu`.
See here: https://nodejs.org/api/intl.html#

Maybe it would be a good idea to use some kind of polyfill for that like [this Intl polyfill](https://github.com/andyearnshaw/Intl.js) to handle it just in case.
## Expected behavior

Show the prices also on SSR correctly.

* Use Node 10 LTS
* Setup a german store-view
* Load a product in german store-view
* You should see a flickering before the browser is rendering the number using `price.js` – first it is `10.99 €`, then `10,99 €`
* If you log it, you will see that the format is always in US on server-side – like: `10.99 €`

## Repository

Latest `develop` and `release/v1.11`

- [x] YES
- [ ] NO

Pick one option.

- [x] This is a bug report for test version on https://test.storefrontcloud.io - In this case Developer should create branch from `develop` branch and create Pull Request `2. Feature / Improvement` back to `develop`.
- [x] This is a bug report for current Release Candidate version on https://next.storefrontcloud.io - In this case Developer should create branch from `release` branch and create Pull Request `3. Stabilisation fix` back to `release`.

## Environment details

- Browser: Chrome
- OS: OSX
- Node: 10 LTS
- Code Version: `release/v1.11` and `develop`

cewald:
### Related issues

closes #3836

If you use a Node version which not yet contains the full-icu package by default like v10-LTS, the numbers aren't formatted correctly on server-side which leads into a flickering of prices on initial page-load.


cewald:
I'm actually not sure if it wouldn't be better to integrate it directly via webpack also the Intl.js package seems to be a pretty heavy polyfill lib.

Maybe in the end it would be easier and faster to make this configurable as a `seperator` config value and do it using a simply string function.

pkarw:
Nice one! However I’m not a fan of increasing the bundle size - especially as this issue is won’t affect all the users. To make this import really dynamic (to not increase the bundle size) it should be done rather by “import” to enforce webpack to do a lazy loading only when the conditions have been met.

Wdyt?

cewald:
Yes, indeed. I will implement it using an dynamic import.

cewald:
@pkarw I added the dynamic import and put it in an extra file.
We could also include it as an extra module using a `afterAppInit` hook – to make it optional if needed – but as it is it will only be included if not available – so it shouldn't increase any package size unnecessarily.

core/i18n/intl.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export const checkForIntlPolyfill = async (storeView) => {
+  if (global.Intl) {
+    if (!areIntlLocalesSupported(storeView.i18n.defaultLocale)) {
+      await importIntlPolyfill()
+    }
+  } else {
+    await importIntlPolyfill()
+  }
```
andrzejewsky: It's exacly same logic:
```
if (!global.Intl || !areIntlLocalesSupported(storeView.i18n.defaultLocale)) {
 await importIntlPolyfill()
}
```

cewald: Yes indeed, I changed it.

core/i18n/intl.ts
```diff
@@ -0,0 +1,12 @@
+import areIntlLocalesSupported from 'intl-locales-supported'
+
+export const importIntlPolyfill = async () => {
+  const IntlPolyfill = await import('intl')
+  global.Intl = IntlPolyfill.default
+}
+
+export const checkForIntlPolyfill = async (storeView) => {
+  if (!global.Intl || !areIntlLocalesSupported(storeView.i18n.defaultLocale)) {
```
gibkigonzo: shouldn't we need to check window also :thinking: ? `!(window || global).Intl` and same in line 5

gibkigonzo:
If we have fully supported Intl, then maybe we can replace `dayjs` module with `Intl.DateTimeFormat` :thinking: But then `date` filter will require different format as argument, so it's some kind of breaking change.

core/i18n/intl.ts
cewald: Good point, there are browsers which don't support `Intl` –  like mobile Safari: https://caniuse.com/#search=Intl

andrzejewsky:
@cewald Please fix tests / build - there is an error related to intl (SSR)

cewald:
@andrzejewsky I fixed it – It was a TS type exception bug during the build.

cewald:
@andrzejewsky I'll get a `ReferenceError: window is not defined` on the server.
It's caused by this change: https://github.com/DivanteLtd/vue-storefront/pull/3837#discussion_r361850049 – I'll will improve this line and open another PR.

PR #3843, opened 2019-11-26
Use cloneDeep to make deep copies of filters

grimasod:

Some search filters contain arrays, so they need to be deep copied, to avoid polluting the config object. Previously filters were cloned with Object.assign(), but this results in arrays being copied by reference. Now using lodash's cloneDeep.


core/modules/catalog-next/store/category/actions.ts
```diff
@@ -140,7 +141,7 @@ const actions: ActionTree<CategoryState, RootState> = {
       categorySearchOptions.filters.id = searchedIds.filter(categoryId => !getters.getCategoriesMap[categoryId] && !getters.getNotFoundCategoryIds.includes(categoryId))
     }
     if (!searchingByIds || categorySearchOptions.filters.id.length) {
-      categorySearchOptions.filters = Object.assign(config.entities.category.filterFields, categorySearchOptions.filters)
+      categorySearchOptions.filters = Object.assign(cloneDeep(config.entities.category.filterFields), categorySearchOptions.filters ? cloneDeep(categorySearchOptions.filters) : {})
```
andrzejewsky:
```suggestion
      categorySearchOptions.filters = cloneDeep(Object.assign(config.entities.category.filterFields, categorySearchOptions.filters ? categorySearchOptions.filter) : {}))
```

core/modules/catalog-next/store/category/actions.ts
```diff
@@ -176,7 +177,7 @@ const actions: ActionTree<CategoryState, RootState> = {
     let resultFilters = aggregationFilters
     const filtersKeys = Object.keys(filters)
     if (categoryMappedFilters && filtersKeys.length) {
-      resultFilters = Object.assign({}, categoryMappedFilters, omit(aggregationFilters, filtersKeys))
+      resultFilters = Object.assign(cloneDeep(categoryMappedFilters), cloneDeep(omit(aggregationFilters, filtersKeys)))
```
andrzejewsky:
```suggestion
      resultFilters = cloneDeep(Object.assign(categoryMappedFilters, omit(aggregationFilters, filtersKeys)))
```

core/modules/catalog/store/product/actions.ts
```diff
@@ -684,7 +685,7 @@ const actions: ActionTree<ProductState, RootState> = {
     if (product && product.category_ids) {
       let currentCategory = rootGetters['category-next/getCurrentCategory'] // use current category, if set
       if (!currentCategory || !currentCategory.id || !product.category_ids.includes(currentCategory.id.toString())) {
-        const categoryFilters = Object.assign({ 'id': product.category_ids }, config.entities.category.breadcrumbFilterFields)
+        const categoryFilters = Object.assign({ 'id': [...product.category_ids] }, cloneDeep(config.entities.category.breadcrumbFilterFields))
```
andrzejewsky:
```suggestion
        const categoryFilters = cloneDeep(Object.assign({ 'id': product.category_ids },config.entities.category.breadcrumbFilterFields))
```

grimasod:
Yeah, I understand. The issue comes from passing around the Config object, assigning parts of it to other variables and then passing those vars to other functions. It becomes very difficult to determine where the mutations are happening.

Of course the Config object should be absolutely read-only. When it's shallow copied, using Object.assign() or destructuring, then there's always a chance of causing problems. So, wouldn't it be safer to always clone the relevant section? Especially in cases like this, when we can't know what the contents are.

Anyway, if it's a concern, I can look into each one in more detail.

I'll comment on your suggestions. At least one of them needs multiple clone operations.

core/modules/catalog-next/store/category/actions.ts
grimasod: In this suggestion, the Object.assign() function would mutate the config, by merging categorySearchOptions.filters into config.entities.category.filterFields.
We need to either clone both config.entities.category.filterFields and categorySearchOptions, or we could use an empty object {} as the first parameter of Object.assign()

core/modules/catalog-next/store/category/actions.ts
grimasod: In this case too, we need either multiple clone operations or Object.assign({}, ...)

core/modules/catalog/store/product/actions.ts
grimasod: I'm not sure about this suggestion. If we know that product.category_ids is a 1-dimension array, then it's sufficient to use destructuring. This way we're cloning a smaller object.

core/modules/catalog-next/store/category/actions.ts
andrzejewsky: yeah you are right, ignore it 👍

andrzejewsky:
> Of course the Config object should be absolutely read-only. When it's shallow copied, using Object.assign() or destructuring, then there's always a chance of causing problems. So, wouldn't it be safer to always clone the relevant section?

Not really, we can use config directly but we should always keep it mind that we can't do anything that can mutate it. For example when you picking just one section from the config, please don't operate on it later, create a new object instead (eg. with using spread operator) etc.

```
 const newObj = { ...config.section, yourModidication: ... }
```
instead of:

```
 const newObj = config.section
 newObj.yourModidication = ....
```

Of course mentioned `config.section` should be a shallow object.

I think you totally know what I mean.

@grimasod please update the changelog as well.

PR #3941, opened 2019-12-23
docs: syntax error in example

pixelass:
### Related Issues

documentation issue

The example code didn't make sense. (function call instead of closing bracket)


andrzejewsky:
Not really, `log` returns a function, you have to call it beforehand - so the only thing that is missing there is curly bracket at the end, so `YourCorePage.methods.foo = function () { Logger.log('Overrided method foo')() }`

pixelass:
Since there is no reference to `Logger.log` I assumed that it is, as the name suggests a log and therefore behaves similar to `console.log` which returns `undefined` or a winston Logger, which returns the logger instance.

I can change it back and add the closing brace if that makes more sense.

pixelass:
changed!

PR #3968, opened 2020-01-02
bugfix - revert init filters

gibkigonzo:

On init load https://test.storefrontcloud.io/women/women-20 filters are missing on SSR so we get price without format.

### DIFF
--- a/core/lib/multistore.ts
+++ b/core/lib/multistore.ts
@@ -11,6 +11,9 @@ import { coreHooksExecutors } from '@vue-storefront/core/hooks'
 import { StorageManager } from '@vue-storefront/core/lib/storage-manager'
 import { LocalizedRoute, StoreView } from './types'
 import storeCodeFromRoute from './storeCodeFromRoute'
+import cloneDeep from 'lodash-es/cloneDeep'
+import get from 'lodash-es/get'
+import { isServer } from '@vue-storefront/core/helpers'
 
 function getExtendedStoreviewConfig (storeView: StoreView): StoreView {
   if (storeView.extend) {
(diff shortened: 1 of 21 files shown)


gibkigonzo:
I just reverted this from 1.10.5. It's wierd problem because if we call those initialization in `once` then filters exist on Vue prototype but there are not triggered in template. And it happens only in ssr (It's ok in csr). And same is with mixin. They are not triggered in ssr. I will try to investigate it further

gibkigonzo:
@pkarw Problem is that we add filters once and then in `price` filter we import `currentStoreView` which returns storeView from cached module (from first request). So I've added `storeView` to global in SSR, we will get always newest `storeView` in ssr. Also removed this helper https://github.com/DivanteLtd/vue-storefront/pull/3968/files#diff-c2effc5bdc3d0e591b453db20306e6d9L82 because it's not needed anymore.

gibkigonzo:
Or we could pass storeView to price filter (and date filter). Filters would be more pure, but we would have to pass this every time in template `product.price | price(currentStoreView())`

core/lib/multistore.ts
```diff
@@ -30,21 +33,28 @@ function getExtendedStoreviewConfig (storeView: StoreView): StoreView {
...
+    storeCode: null,
+    storeId: config.defaultStoreCode && config.defaultStoreCode !== '' ? config.storeViews[config.defaultStoreCode].storeId : 1,
+    seo: config.seo
+  })
+}
+
 export function currentStoreView (): StoreView {
-  // TODO: Change to getter all along our code
-  return rootStore.state.storeView
+  const serverStoreView = get(global, 'process.storeView', undefined)
```
andrzejewsky: what about client-side? there is no global?

andrzejewsky:
@gibkigonzo one comment left. Is it ready?

core/lib/multistore.ts
gibkigonzo: there is no global on client-side, I could use window, but I think we can stay with same behavior as it was for client

gibkigonzo: it is in next line   `const clientStoreView = get(rootStore, 'state.storeView', undefined)`

gibkigonzo:
@pkarw @andrzejewsky I've applied suggestion

PR #3994, opened 2020-01-17
add lazy create cart token 

gibkigonzo:

This allows to create cart token when user add product to cart. There is edge case: when user load page, go to offline, add product, go to checkout => can't make order. That's because he hasn't got payment methods which are loaded after token creation. Also I've added refresh payment methods in checkout. Right now in demo if somehow user get to checkout without loaded payment-methods then he can't make order, even if he get back online

### QA
A.
1. clear cache
2. add product to cart
3. go to checkout
4. make order
5. Expected: user made order

B.
1. clear cache
2. go offline
3. add product to cart
4. check cart - there should be product
5. go to checkout
6. you should not make order
7. turn online
8. now you can make order

Same can be checked with user logged in.

### DIFF
--- a/core/modules/cart/store/actions.ts
+++ b/core/modules/cart/store/actions.ts
@@ -226,7 +226,7 @@ const actions: ActionTree<CartState, RootState> = {
     return dispatch('sync', { forceClientState, dryRun })
   },
   /** @description this method is part of "public" cart API */
-  async load ({ getters, commit, rootGetters, dispatch }, { forceClientState = false, createCartToken = false }: {forceClientState?: boolean, createCartToken?: boolean} = {}) {
+  async load ({ getters, commit, rootGetters, dispatch }, { forceClientState = false }: {forceClientState?: boolean} = {}) {
     if (isServer) return
     const cartShippingMethod = getters.getShippingMethod
     if ((!cartShippingMethod || !cartShippingMethod.method_code) && (Array.isArray(rootGetters['shipping/shippingMethods']))) {
@@ -252,10 +252,16 @@ const actions: ActionTree<CartState, RootState> = {
         Logger.info('Cart token received from cache.', 'cache', token)()
         Logger.info('Syncing cart with the server.', 'cart')()
         dispatch('sync', { forceClientState, dryRun: !config.cart.serverMergeByDefault })
-      } else if (createCartToken && storedItems.length) { // create cart token only when it's needed and when there are any products
-        Logger.info('Creating server cart token', 'cart')()
-        await dispatch('connect', { guestCart: false })
       }
+      await dispatch('createCartToken')
+    }
+  },
+  async createCartToken ({dispatch, getters}) {
+    const storedItems = getters['getCartItems'] || []
+    const cartToken = getters['getCartToken']
+    if (storedItems.length && !cartToken) {
+      Logger.info('Creating server cart token', 'cart')()
+      await dispatch('connect', { guestCart: false })
     }
   },
   /** Get one single item from the client's cart */
@@ -350,6 +356,7 @@ const actions: ActionTree<CartState, RootState> = {
       }
       productIndex++
     }
+    await dispatch('createCartToken')
     if (getters.isCartSyncEnabled && getters.isCartConnected && !forceServerSilence) {
       return dispatch('sync', { forceClientState: true })
     } else {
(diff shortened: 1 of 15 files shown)


gibkigonzo:
You are right. I revert `authorize` and add changed `createCartToken` to `create`

ArturDivante:
After latest changes, I tested both scenarios with annonymous and logged-in customer. Behavior is consistent with your description (note:in scenario 2 for logged-in user, cart is created in the moment of logging in, even if customer have no products in it) 

In the meantime, we discovered unrelated problem with cart sync in some specific cases. Will be reported separately in: #3995

core/modules/cart/store/actions.ts
```diff
@@ -252,10 +251,8 @@ const actions: ActionTree<CartState, RootState> = {
         Logger.info('Cart token received from cache.', 'cache', token)()
         Logger.info('Syncing cart with the server.', 'cart')()
         dispatch('sync', { forceClientState, dryRun: !config.cart.serverMergeByDefault })
-      } else {
-        Logger.info('Creating server cart token', 'cart')()
-        await dispatch('connect', { guestCart: false })
       }
```
pkarw: Shouldn’t we have else condition in here OR remove “sync” call above? Isn’t this form causing double sync? It probably is because create calls connect which calls sync for the second time. Isn’t it?

core/modules/cart/store/actions.ts
```diff
@@ -141,7 +141,6 @@ const actions: ActionTree<CartState, RootState> = {
     if (options.recreateAndSyncCart && getters.isCartSyncEnabled) {
       await commit(types.CART_LOAD_CART_SERVER_TOKEN, null)
       await commit(types.CART_SET_ITEMS_HASH, null)
-      await dispatch('connect', { guestCart: !config.orders.directBackendSync }) // guest cart when not using directBackendSync because when the order hasn't been passed to Magento yet it will repopulate your cart
```
pkarw: How do we handle this directBackendSync = false? It’s when orders are not directly passed to the server but rather queued in local storage (offline mode). In that case the cart should be recreated as guest cart bc you can’t be sure that the order has been already passed to the server so the authorized user quote can still contain the products so the effect will be they “return” to the cart after placing an order

core/modules/cart/store/actions.ts
gibkigonzo: we are checking in `create` if token already exist https://github.com/DivanteLtd/vue-storefront/pull/3994/files#diff-2101c92c40c144e3e56fb400ac211e3dR505 So if in above if statement we add token, then in `create` it is not called anymore.

pkarw: Ok 👌

core/modules/cart/store/actions.ts
gibkigonzo: It sounds complicated :sweat_smile: Why not always clear cart after order? No matter if it's send to backend or not. We don't need to send cart token with order because it will be created in api https://github.com/DivanteLtd/vue-storefront-api/blob/master/src/platform/magento2/o2m.js#L65

So maybe let's make it even simpler:
```
  async clear ({ commit, dispatch, getters }, options = { recreateAndSyncCart: true }) {
    await commit(types.CART_LOAD_CART, [])
    await commit(types.CART_LOAD_CART_SERVER_TOKEN, null)
    await commit(types.CART_SET_ITEMS_HASH, null)
  },
```
If we always clear cart then there is no problem with returning products back to cart. :thinking: Do we need to keep cart token after order, even if it's not send?

gibkigonzo: Ok, I make it as wrote before. So now we will remove cart always after an order.

andrzejewsky:
@gibkigonzo can you fix the tests?

gibkigonzo:
@andrzejewsky done :ok_hand:

gibkigonzo:
I've made one more update https://github.com/DivanteLtd/vue-storefront/pull/3994/commits/d3debc4d457e07633f94db9efbac479a1821f05b
We don't need this config anymore as with this pr we don't create token after order.

andrzejewsky:
@gibkigonzo and... again... changelog... 😄

gibkigonzo:
@andrzejewsky done :sweat_smile:

PR #3728, opened 2019-10-18
remove prefech dayjs locales

gibkigonzo:
### Related issues

closes #3498

### Short description and why it's useful

Added function that creates regexp with needed `dayjs` locales

### QA
1. open http://next.storefrontcloud.io, clear cache and refresh page
2. open Elements tab in devtools and expand `head` tag. There should be loaded a lot of `dayjs...js` files
3. checkout on current branch, prepare env (`yarn build` and `yarn start`)
4. make 1 and 2 step on current branch page
5. In `head` should be loaded only needed `dayjs...js` files based on config. 
6. Previous steps could be tested with different config: 
 - without specified defaultLocale in `config.i18n.defaultLocale` (should load only `en` file, because it's fallback language)
 - without `multistore` config (should load `en` and one specified in `config.i18n.defaultLocale`)
 - with `multistore` config (should load `en` and all specified in storeView `i18n.defaultLocale`)

### Which environment this relates to
Check your case. In case of any doubts please read about [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html)

- [ ] Test version (https://test.storefrontcloud.io) - this is a new feature or improvement for Vue Storefront. I've created branch from `develop` branch and want to merge it back to `develop`
- [x] RC version (https://next.storefrontcloud.io) - this is a stabilisation fix for Release Candidate of Vue Storefront. I've created branch from `release` branch and want to merge it back to `release`
- [ ] Stable version (https://demo.storefrontcloud.io) - this is an important fix for current stable version. I've created branch from `hotfix` or `master` branch and want to merge it back to `hotfix`

### Upgrade Notes and Changelog

- [x] No upgrade steps required (100% backward compatibility and no breaking changes)
- [x] I've updated the [Upgrade notes](https://github.com/DivanteLtd/vue-storefront/blob/develop/docs/guide/upgrade-notes/README.md) and [Changelog](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md) on how to port existing VS sites with this new feature

**IMPORTANT NOTICE** - Remember to update `CHANGELOG.md` with description of your change

### Contribution and currently important rules acceptance

- [x] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)

PR #4002, opened 2020-01-21
Proposing two fixes for i18n issues

jpkempf:

This PR proposes to adjust some behaviour around locales and localisation files that was originally introduced in #3728:

- `092493e` fixes a problem where the only working locale in a single-store, multi-language scenario was the default locale.
- `60c1206` fixes a problem where, if `bundleAllStoreviewLanguages` is set to `false`, a `${lang}.json` file would only be generated for the fallback locale.

I work on a team that has built a custom b2b application on top of VSF and without `092493e`, our language switching mechanism would be completely broken. The changes from `60c1206` are not actually relevant to us, but we came across this in our investigation and from what I can tell that line in it's current form does not make any sense, so we're proposing a fix for that as well here.

Perhaps @gibkigonzo can weigh in on whether or not the proposed changes in this PR represent an improvement.

I'm happy to adjust any mistakes I might have made in the steps I took to create this PR, and please feel free to also refer to @gopkumargopi for questions. Thank you in advance for your time!


jpkempf:
Happy to help! Would love to see it merged if there are no concerns 😇

Issue #4004, opened 2020-01-22
Comparison of bundled products in the productsEquals.ts


gk-daniel:
## Current behavior 

In the productsEquals.ts file, we can notice that bundle products are checked for attributes such as product.server_item_id, product.item_id or checksum. In our online store, we found that this condition (line 37) can return false, and in this case, when a bundled product is added to the basket, no more products can be added.

## Expected behavior 

To fix this issue, we need to slightly modify the condition so that it is possible to check the product bundle for sku.

## Can you handle fixing this bug by yourself?

- [x] YES
- [ ] NO 

## Additional information
Below I will publish a link to a PR request, where you could consider my proposal

https://github.com/DivanteLtd/vue-storefront/pull/4005

PR #4005, opened 2020-01-22
Comparison of bundled products in the productsEquals.ts

gk-daniel:
### Related Issues
#4004

closes #

In the absence of these attributes from the product, as a result, returning false can lead to the problem of working with adding the product to the card, as happened in our store. This decision corrects this situation.


andrzejewsky:
@gk-daniel I've tried to rebase your branch again but the second time - I've got permission denied - here is your PR: https://github.com/DivanteLtd/vue-storefront/pull/4013

andrzejewsky:
@gk-daniel and when it comes to your changes - make sense you just allow compare by sku when checksums are not equals

PR #4013, opened 2020-01-22
Bundlefix

### LINKED ISSUE #4004: Comparison of bundled products in the productsEquals.ts

andrzejewsky:
### Related Issues
closes #4004
https://github.com/DivanteLtd/vue-storefront/pull/4005


gk-daniel:
@andrzejewsky thanks for applying the changes!

PR #4015, opened 2020-01-23
Fix: slugify sub-category names

1070rik:
The old function wouldn't show any categories at all if you added the "*.name" field in the includedFields array in your config. The way it works now is also easier to read.

- [ ] I've updated the [Upgrade notes](https://github.com/DivanteLtd/vue-storefront/blob/develop/docs/guide/upgrade-notes/README.md) and [Changelog](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md) on how to port existing Vue Storefront sites with this new feature

### DIFF
--- a/core/modules/catalog/helpers/slugifyCategories.ts
+++ b/core/modules/catalog/helpers/slugifyCategories.ts
@@ -20,11 +20,13 @@ const createSlug = (category: ChildrenData): string => {
 
 const slugifyCategories = (category: Category | ChildrenData): Category | ChildrenData => {
   if (category.children_data) {
-    for (let subcat of category.children_data) {
-      if (subcat.name) {
-        return slugifyCategories({ ...subcat, slug: createSlug(subcat) } as any as ChildrenData)
+    category.children_data.forEach((subCat: ChildrenData): void => {
+      if (subCat.name) {
+        subCat.slug = createSlug(subCat)
       }
-    }
+
+      slugifyCategories(subCat)
+    })
   }
 
   return category
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed Search product fails for category filter when categoryId is string - @adityasharma7 (#3929)
 - Revert init filters in Vue app - @gibkigonzo (#3929)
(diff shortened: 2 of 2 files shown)


core/modules/catalog/helpers/slugifyCategories.ts
```diff
@@ -20,11 +20,13 @@ const createSlug = (category: ChildrenData): string => {
 
 const slugifyCategories = (category: Category | ChildrenData): Category | ChildrenData => {
   if (category.children_data) {
-    for (let subcat of category.children_data) {
-      if (subcat.name) {
-        return slugifyCategories({ ...subcat, slug: createSlug(subcat) } as any as ChildrenData)
+    category.children_data.forEach((subCat: ChildrenData): void => {
```
andrzejewsky: @1070rik What about shomething like this?
```
category.children_data = category.children_data.map(ch => ({
  ...ch,
  ...(ch.name ? { slug: createSlug(subCat) } : {})
}))
```

1070rik: Although I don't really see the added value of doing this instead of a forEach, this would only work for the first 2 levels of categories.
The `slugifyCategories(subCat)` call at the end of the loop would cause a recursive loop which would give the subcategories of that subcategory a slug.

andrzejewsky: @1070rik yes, but I didn't mention about deleting `slugifyCategories(subCat)` my changes are related to only `for`, it's just simpler, and actually entire function could be refactored with using `map` call. However - it's your choice, just let me know if you want to spare time on it or not - it's still good PR (it solves the issue,  thanks!)

1070rik: Turning it into a map and doing the function call doesn't really make the code look better or easier, it only makes it more complex in my opinion. I still prefer the forEach method over the map since it easier to understand.

andrzejewsky: for me, it looks much better, it's just imperative vs functional. In this case, it maybe doesn't change so much, but remember that in the `for` you are mutating values while `map` returns a new array - no side effects (they could be painful sometimes). As I mentioned entire function could be refactored using `map` - instead of modifying category you can return a new one with slugs inside (again, no side effects). 

Of course, this solution is also fine, thanks!

PR #4019, opened 2020-01-23
Server redirects to localizedRoute('/') not '/'

Fifciu:
When we open `/storeCode/checkout` we are being redirected to `/`. It is really bad when in our multistore, there is not defaultStoreCode.

With that it always redirect to `localizedRoute('/')`


Fifciu:
> Please just update the changelog

Ok, done

PR #4056, opened 2020-02-05
Fix misplaced parenthesis for `taxClasses.find`

DylannCordel:
(and also avoid a `ParseInt` call for each item tested)


andrzejewsky:
@DylannCordel just changed base for you - please only update changelog as @pkarw and we can merge it.

DylannCordel:
Thx @andrzejewsky  and @pkarw 
Changelog has been updated. Same fix on the API has been merged into master branch. If `calculateServerSide` is set to True, same bug will still be there until people update the API once it will be released. To avoid that without changing the API source code, a trick is to filter tax rules before the `calculateProductTax` call. It can been done in the `<platform>.taxFor` method. eg:

```typescript
  taxFor (product) {
    let product_tax_class_id = parseInt(product.tax_class_id)
    let productTaxRules = this._taxClasses.filter(
      (taxRule) => taxRule.product_tax_class_ids.indexOf(product_tax_class_id) > -1
    );
    // we DO NOT use `this._taxClasses` because `calculateProductTax` is currently bugged.
    // we MUST pre-filter tax rules to only keep matching rules for the current product.
    return calculateProductTax(product, productTaxRules, this._taxCountry, this._taxRegion, this._sourcePriceInclTax)
  }
```

PR #4084, opened 2020-02-14
Updated paypal-payments.md Documentation

philippsander:
yarn should be used instead of npm


andrzejewsky:
@philippsander it doesn't matter which package manager you want to use, it can be npm as well. Why it must be there npm instead? Consistency with the rest of docs or there is another reason?

PR #4270, opened 2020-04-21
refactor(ayc): api-client revamp

lukeromanowicz:


packages/about-you/api-client/package.json
```diff
@@ -1,19 +1,19 @@
 {
   "name": "@vue-storefront/about-you-api",
-  "version": "0.0.1",
+  "version": "0.1.0",
   "private": true,
   "sideEffects": false,
   "main": "lib/index.cjs.js",
   "module": "lib/index.es.js",
-  "types": "lib/index.d.ts",
+  "types": "lib/about-you/api-client/src/index.d.ts",
```
krskibin, line 8: Is this path correct? Rest of the index files are in lib/index*.

packages/about-you/composables/src/composables/useProduct/index.ts
```diff
@@ -1,9 +1,18 @@
...
+import { getProductsByQuery } from '@vue-storefront/about-you-api';
 
 const useProduct: (cacheId: string) => UseProduct<BapiProduct> = useProductFactory<BapiProduct, any>({
-  productsSearch: mapProductSearch
+  productsSearch: async (params) => {
+    const products = await getProductsByQuery(mapProductSearchByQueryParams(params));
+
+    return {
+      data: products.entities,
+      total: products.pagination?.total || products.entities.length
```
krskibin, line 13:
````
total: products.pagination?.total || products.entities?.length
````

File: packages/about-you/composables/src/helpers/product/mapProductSearchByQueryParams.ts [hunk 3]
```diff
@@ -0,0 +1,22 @@
+import { ProductSearchQuery } from '@aboutyou/backbone/types/ProductSearchQuery';
+import { ProductsSearchEndpointParameters } from '@aboutyou/backbone/endpoints/products/products';
+
+const mapProductSearchByQueryParams = (params): ProductsSearchEndpointParameters => {
+  const searchQuery: ProductSearchQuery = {};
```
krskibin, line 5:
```

const searchQuery = { categoryId: params.catId, term: params.term}
```

packages/about-you/api-client/package.json
lukeromanowicz: Yes, for some reason it is. I double-checked it with @filrak

krskibin: Well. Ok, then

PR #4272, opened 2020-04-22
feat(core): Internal fields

andrzejewsky:


packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -30,6 +29,138 @@ export type UseCartFactoryParams<CART, CART_ITEM, PRODUCT, COUPON> = {
...
+      applyCoupon,
+      removeCoupon,
+      loading: computed(() => loading.value)
+    };
+  };
+
+  return { useCart, setCart };
+};
+
+/*
```
Qrzy, line 163: Existence of this commented out code hides the actual diif - would you kindly remove it?

PR #4274, opened 2020-04-22
feat(ayc-theme): implement category tree

lukeromanowicz:


packages/about-you/theme/helpers/category/getCategoryUrl.js
```diff
@@ -0,0 +1,3 @@
+export function getCategoryUrl(category) {
```
filrak, line 1: isn't tthis `Path` rather than `url`

PR #4281, opened 2020-04-24
fix(core): Fix langs, deployment, packages for CT

andrzejewsky:


packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -39,7 +39,7 @@ export function useCartFactory<CART, CART_ITEM, PRODUCT, COUPON>(
   return function useCart(): UseCart<CART, CART_ITEM, PRODUCT, COUPON> {
     const { initialState, saveToInitialState } = useSSR('vsf-cart');
 
-    factoryParams.cart.value = initialState || null;
+    factoryParams.cart.value = factoryParams.cart.value ? factoryParams.cart.value : initialState || null;
```
Qrzy, line 42:
```suggestion
    factoryParams.cart.value = factoryParams.cart.value || initialState || null;
```

PR #4309, opened 2020-04-28
fix(commercetools): Add missing SSR, fix authenticate bugs

andrzejewsky:


packages/commercetools/theme/plugins/commercetools.js
```diff
@@ -2,14 +2,15 @@ import { setup } from '@vue-storefront/commercetools-api';
 import { config } from './commercetools-config';
 
 const CT_TOKEN_COOKIE_NAME = 'vsf-commercetools-token';
-
+// setup@setip.com
```
Qrzy, line 5: What's this?

PR #4351, opened 2020-05-08
Next: Forgot password support

carlokok:
This implements the forgot password sequence, by adding 3 things:

* Forgot password page in login model
* "we've sent you a mail" page in login model
* Landing page to enter the new password (/new-password/my%40example.com/token)

### Screenshots of Visual Changes before/after (if There Are Any)
afbeelding

afbeelding

afbeelding

* [x] Next version

* [x] No upgrade steps required (100% backward compatibility and no breaking changes)

While this is a non breaking change, nothing happens if the useUser implementation doesn't actually implement the new methods. (There's no changelog/upgrade notes in Next yet)

* [x] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)

┆Issue is synchronized with this [Jira Story](https://vsf.atlassian.net/browse/NEXT-102) by [Unito](https://www.unito.io/learn-more)


filrak:
While `forgotPassword` makes sense isn't `createPassword` just `updateUser` ?

carlokok:
not exactly. Though I agree I don't like the name, I can't think of anything better. Think of this flow:

* User enters email & clicks forgot password
* Server backend sends a mail that ends up with a link like http://localhost:3000/new-password/me%40example.com/tokengeneratedbyserver)
* User enters a new password
* Given the token, the username and the new password, reset the password

So it does update the user, but the user isn't logged in at this point, so can't call update user.

filrak:
ok that makes sense to me, @andrzejewsky wdyt?

packages/core/theme-module/theme/components/LoginModal.vue
```diff
@@ -166,10 +212,11 @@ export default {
   },
   setup() {
     const form = ref({});
-    const isLogin = ref(false);
+    const isLogin = ref(1);
```
filrak, line 215: I think it's better to use strings here (or enums) to make it more readable

packages/core/theme-module/theme/components/LoginModal.vue
```diff
@@ -6,7 +6,7 @@
       class="modal"
       @close="toggleLoginModal">
       <transition name="fade" mode="out-in">
-        <div v-if="isLogin" key="log-in">
+        <div v-if="isLogin == 1" key="log-in">
```
filrak, line 9: Wouldn't it be better to have these 3 options as separate components?

filrak:
Also would you mind implementing the desired functionality you have putted into the interface. I think you need to do this anyway

packages/core/theme-module/theme/components/LoginModal.vue
carlokok: Done (string keys)

carlokok:
not quite sure what you meant with your last reply?

bloodf:

Issue #4300, opened 2020-04-28
[AYC] implement docs


lukeromanowicz:
Implement docs for AYC integration usage.

PR #4365, opened 2020-05-11
feat(ayc): initial ayc docs

lukeromanowicz:
### Related Issues
#4300

### DIFF
--- a/packages/core/docs/aboutyou/api-client.md
+++ b/packages/core/docs/aboutyou/api-client.md
@@ -1,0 +1,87 @@
+---
+platform: AboutYouCloud
+---
+
+
+<IncludeContent content-key="api-client" />
+
+<!-- Code example for setup method -->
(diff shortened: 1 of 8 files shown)


packages/core/docs/aboutyou/api-client.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+:::
+
+::: slot methods
+
+You can find detailed information about all API Client methods [here](./api-client/index.html)
+:::
+
+::: slot override
```
filrak, line 72: Does this work for AYC already?

lukeromanowicz: Haven't tested it but I think it should work, otherwise it needs a bug ticket

lukeromanowicz:
I extracted checkboxes issue in #4391

PR #4446, opened 2020-05-21
Example of more user-friendly error handling

jissereitsma:
When details from Vuex can't be properly loaded (for other reasons beyond the scope of this PR), the errors that you receive are often not user-friendly. A simple PR like the following would improve things. What do you guys think of improving the Dev Experience this way?

### DIFF
--- a/src/themes/default/pages/CmsPage.vue
+++ b/src/themes/default/pages/CmsPage.vue
@@ -19,7 +19,13 @@ import CmsPage from '@vue-storefront/core/pages/CmsPage'
 export default {
   computed: {
     cmsPageContent () {
-      return this.$store.state.cmsPage.current
+      if (this.$store.state.cmsPage.current) {
+        return this.$store.state.cmsPage.current
+      }
+
+      Logger.error('Current CMS page could not be loaded from Vuex', '$store.state.cmsPage.current')()
+
+      return {}
     }
   },
   mixins: [CmsPage]


jissereitsma:
@lukeromanowicz Thanks. But before moving ahead, let's have a chat on this first. Doesn't this have an impact on the bundle size, with an eye on tree shaking? I'm no tree shaking expert, but it might be an issue. What do you think?

gibkigonzo:
@jissereitsma Hey :) Probably there a lot of places where we need more logging, but I'm not sure if this is a good example. Right now here we throw error https://github.com/DivanteLtd/vue-storefront/blob/master/core/modules/cms/store/page/actions.ts#L35 if cms page doesn't exist. And we catch it here https://github.com/DivanteLtd/vue-storefront/blob/master/core/pages/CmsPage.js#L25 . So I think that flow is ok. 

Cms page doesn't exist => we catch it in `asyncData` - which is called before component is resolved so getter will not crash => we get redirection to page not found

So we get info for dev in console, for example:
```
CMS query returned empty result null
Redirect for resource not found : /i/customer-service
```
and user gets info that this page doesn't exist.

In other hand if we will keep user on that page with `cmsPageContent` as `{}` then he will see empty page with no information.

Also getter is not the best place to log info :thinking: It should be dumb. get data => transform it => return data

IMO best place to log info is where we get it. Catch error there and return falsy data or throw error to handle it in higher order function.

PR #4488, opened 2020-05-29
Remove categories with unknown `url_path` from breadcrumbs

jissereitsma:
In already 4 instances of Vue Storefront 1 with Magento 1, I've encountered the following issue: When categories are loaded into the breadcrumbs, they include the Magento Root Category (often with a name like **Default Category**). But because this Root Category serves a special purpose, the URL path is always empty. This fix simply checks whether the URL path is `undefined` and if so, it skips that category. Because the breadcrumbs are supposed to only contain categories with the URL path set.


gibkigonzo:
Hey @jissereitsma wdyt about adding changes to `formatCategoryLink` instead
before:
```
return config.seo.useUrlDispatcher
  ? ('/' + storeCode + category.url_path)
  : ('/' + storeCode + 'c/' + category.slug)
```

after:
```
return config.seo.useUrlDispatcher && category.url_path
  ? ('/' + storeCode + category.url_path)
  : ('/' + storeCode + 'c/' + category.slug)
```
format `/c/:slug` is always valid even when we use url_path. If some category will be missing url_path then it will fallback to slug.

gibkigonzo:
Can you please rebase to branch `hotfix/v1.12.1`?

jissereitsma:
@gibkigonzo This should be done now.

core/modules/breadcrumbs/helpers/index.ts
```diff
@@ -5,6 +5,7 @@ import { formatCategoryLink } from '@vue-storefront/core/modules/url/helpers'
 export function parseCategoryPath (categoryPath) {
   let routesArray = []
   for (let category of categoryPath) {
+    if (category.url_path == undefined) continue;
```
gibkigonzo, line 8: build failed https://travis-ci.org/github/DivanteLtd/vue-storefront/jobs/699239622 Maybe `isUndefined(category.url_path) || isNull(category.url_path)` from lodash?

jissereitsma: Thanks. I choose to not use Lodash but simply fix the code (from `==` into `===`).

gibkigonzo: great :) thanks

PR #4593, opened 2020-07-01
Feat/#4584

### LINKED ISSUE #4584: Make all directories in themes watchable

Currently only few from integration theme files (`{integration}/theme`)  are watched and these directories have to be [explicitly declared](https://github.com/DivanteLtd/vue-storefront/blob/next/packages/core/theme-module/index.js#L42).

This obviously is not the best solution as every integration could have different needs and there is no simple way of watching additional directories.

We could create a separate dir in `theme` folder to keep watched files but that would lead to breaking changes. Much safer way would be to just watch all **directories** inside `theme` folder and copy them to `theme/.theme`.

- [ ] All directories except `.theme` and `nuxt`, `node_modules` `test` are watched by `theme-module` in integration themes 
- [ ] Newly added directories are watched by default

Fifciu:
### Related Issues
Closes #4584

### DIFF
--- a/packages/core/theme-module/index.js
+++ b/packages/core/theme-module/index.js
@@ -18,10 +18,10 @@ const log = {
 module.exports = async function DefaultThemeModule(moduleOptions) {
   log.info(chalk.green('Starting Theme Module'));
 
-  const baseThemeDir = path.join(__dirname, 'theme');
+  const agnosticThemeDir = path.join(__dirname, 'theme');
   const projectLocalThemeDir = this.options.buildDir.replace('.nuxt', '.theme');
 
-  const themeFiles = getAllFilesFromDir(baseThemeDir).filter(file => !file.includes(path.sep + 'static' + path.sep));
+  const agnosticThemeFiles = getAllFilesFromDir(agnosticThemeDir).filter(file => !file.includes(path.sep + 'static' + path.sep));
 
   const compileAgnosticTemplate = (filePath) => {
     return compileTemplate(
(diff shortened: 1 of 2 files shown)


packages/core/theme-module/index.js
```diff
@@ -19,10 +19,18 @@ module.exports = async function DefaultThemeModule(moduleOptions) {
 
   const baseThemeDir = path.join(__dirname, 'theme');
   const projectLocalThemeDir = this.options.buildDir.replace('.nuxt', '.theme');
-  const themeComponentsDir = path.join(this.options.rootDir, 'components');
-  const themePagesDir = path.join(this.options.rootDir, 'pages');
-  const themeHelpersDir = path.join(this.options.rootDir, 'helpers');
-  const languageDir = path.join(this.options.rootDir, 'lang');
+
+  const getDirectoriesList = directoryPath => fs.readdirSync(directoryPath).filter(
```
filrak, line 23: **code organization** We could put it in a separate file like `getAllFilesFromDir` and rename it to something that matches this pattern (like `getAllSubDirs`)

packages/core/theme-module/index.js
```diff
@@ -37,14 +45,19 @@ module.exports = async function DefaultThemeModule(moduleOptions) {
...
+  const themeDirectoriesPaths = [];
+  const copyThemeDirectoriesPromises = [];
+
+  for (const directory of getDirectoriesList(this.options.rootDir)) {
+    if (!omittedDirectories.includes(directory)) {
+      const absolutePath = path.join(this.options.rootDir, directory);
+      themeDirectoriesPaths.push(absolutePath);
+      copyThemeDirectoriesPromises.push(copyThemeFiles(absolutePath));
+    }
+  }
```
filrak, line 57: **code organization** Grouping highlited code into a function that returns array of paths (like themeFiles) will make this piece more readable and keep all the code related to this particular action in one place (easier to move or remove).

Also in that case `filter` seems much more natural than `for...of`. 

Try to find similar patterns in the file and follow them 
```
const themeFiles = getAllFilesFromDir(baseThemeDir).filter(file => !file.includes(path.sep + 'static' + path.sep));
```

packages/core/theme-module/index.js
```diff
@@ -19,10 +19,18 @@ module.exports = async function DefaultThemeModule(moduleOptions) {
...
-  const themeComponentsDir = path.join(this.options.rootDir, 'components');
-  const themePagesDir = path.join(this.options.rootDir, 'pages');
-  const themeHelpersDir = path.join(this.options.rootDir, 'helpers');
-  const languageDir = path.join(this.options.rootDir, 'lang');
+
+  const getDirectoriesList = directoryPath => fs.readdirSync(directoryPath).filter(
+    file => fs.statSync(path.join(directoryPath, file)).isDirectory()
+  );
+
+  const omittedDirectories = [
```
filrak, line 27: **code organization** I would put that in the function below that I suggested

PR #4622, opened 2020-07-15
Next: How to use checkout.com & modules

Fifciu:
Basic readmes

### DIFF
--- a/packages/checkout-com/README.md
+++ b/packages/checkout-com/README.md
@@ -42,7 +42,7 @@ const {
 ```
 
 `error` - contains error message from the response if you do not use 3ds or we have some server related issues. If the user just removed stored token from localStorage it will have `There is no payment token` inside.
-`makePayment` - it creates context, then payment, then removes card token. Returns Promise<Payment> if succeed, or Promise<null> if failed.
+`makePayment` - it proceeds with the payment and removes card token afterward. Returns Promise<Payment> if succeed, or Promise<null> if failed.
 
 2. You should `makePayment` at first (remember to check if everything went ok):
 ```js
--- a/packages/checkout-com/README.md
+++ b/packages/checkout-com/README.md
@@ -1,0 +1,70 @@
+# Checkout.com Nuxt module
+## How to install
+1. Open your `nuxt.config.js`
+2. At the bottom of `modules` add:
+```js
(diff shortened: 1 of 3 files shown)


packages/commercetools/composables/nuxt/README.MD
```diff
@@ -1,4 +1,40 @@
...
+      'manage_my_payments:vsf-ct-dev',
+      'view_products:vsf-ct-dev',
+      'view_published_products:vsf-ct-dev'
+    ]
+  }
+}]
+```
+
+## Generate token middleware 
+Commercetools nuxt module injects middleware that is responsible for createing/refreshing token if needed - when user refreshes page. In some rare case, you might want to disable this middleware. If so, add `disableGenerateTokenMiddleware` to the `@vue-storefront/commercetools/nuxt`'s options module in `buildModules`:
```
filrak, line 28: worth putting a link to this middleware

packages/checkout-com/README.md
```diff
@@ -0,0 +1,70 @@
...
+1. In this step you need:
+```js
+const { makePayment, error: paymentError } = useCkoCard();
+const { cart } = useCart();
+const {
+    placeOrder
+} = useCheckout();
+```
+
+`error` - contains error message from the response if you do not use 3ds or we have some server related issue. If user just removed stored token from localStorage it will have `new Error('There is no payment token')` inside.
```
andrzejewsky, line 44:
```suggestion
`error` - contains error message from the response if you do not use 3ds or we have some server related issues. If the user just removed stored token from localStorage it will have `There is no payment token` inside.
```

packages/commercetools/composables/nuxt/README.MD
Fifciu: Added

packages/checkout-com/README.md
```diff
@@ -0,0 +1,70 @@
...
+`makePayment` - it creates context, then payment, then removes card token. Returns Promise<Payment> if succeed, or Promise<null> if failed.
```
andrzejewsky, line 45:
```suggestion
`makePayment` - it proceeds with the payment and removes card token afterward. Returns Promise<Payment> if succeed, or Promise<null> if failed.
```

PR #4625, opened 2020-07-16
Clean checkout

andrzejewsky:


Fifciu:
We should remove build:checkout-com part
```
"build:ct": "yarn build:core && yarn build:checkout-com && yarn build:ct:tools && yarn build:ct:theme",
```

PR #4638, opened 2020-07-20
Scripts to release

### LINKED ISSUE #4586: Prepare scripts to release integration packages and core (part 1)
ATM there is no simple way to publish NPM packages of a certain integration. What we agreed on is that we don't want to rely on lerna anymore as our repo is more like a set of independent monorepos than one big monorepo. We're keeping lerna only for local symlinks

Each integration currently has:
- API Client
- Composables (depending on API Client)
- Theme (depending on Composables)
- CLI (depending on Theme)

We should prepare a script (`release.js`) inside root `scripts` directory that is responsible for releasing packages related to specific integration.

# Propsed API:
```bash
node ./scripts/release.js {integration}/{package[optional])
```

- Release all packages for commercetools
```bash
node ./scripts/release.js commercetools
```

- Release only commercetools composables
```bash
node ./scripts/release.js commercetools/composables
```
There should be 3 gradations of releases: `major` (x.0.0) `minor` (0.x.0), `path` (0.0.x) all fo them specified as cli argument (required)

For example to release a minor upgrade for all commercetools packages we should write:
```bash
node ./scripts/release.js commercetools - minor
```

Commands: `core/comemrcetools/about-you` (you can automate it and base on dir structure for everything so it will be {integration}/package-name-without-workspace} (without `@vue-stroefront`)
# Implementation details

Under the hood you should just use some npm client to release the packages (there is a bunch of them on the internet, do your investigation ;>)

You should be cautious about dependencies versioning. When we release a package and bump its version we want to be sure that all its dependants have it upgraded as well.

Fifciu:
Close #4586


scripts/release/getPackageType.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const integrationWrapperPackages = [
+  'api-client',
+  'composables',
+  'theme'
+];
+
+const getDirectories = (source: string) =>
+  fs.readdirSync(source, { withFileTypes: true })
+    .filter(dirent => dirent.isDirectory())
```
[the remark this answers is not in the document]
Fifciu: Directory Entity

scripts/release/README.md
```diff
@@ -0,0 +1,10 @@
+```
```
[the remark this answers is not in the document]
Fifciu: Ok, removed whole file

scripts/release/getPackageType.ts
```diff
@@ -0,0 +1,50 @@
...
+    .map(dirent => dirent.name);
+
+const isPackage = (pckg: string) => fs.existsSync(path.resolve(__dirname, `${base}${pckg}/package.json`));
+
+const isIntegrationWrapper = (pckg: string) => {
+  try {
+    const directories = getDirectories(path.resolve(__dirname, `${base}${pckg}`)).filter(dir => integrationWrapperPackages.includes(dir));
+    return integrationWrapperPackages.length === directories.length && directories.every(directory => isPackage(`${pckg}/${directory}`));
```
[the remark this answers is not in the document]
Fifciu: Sure. However, if Additional package A deppends on Additional package B it won't always recognize update version in `dependencies` properly. For that, I need to `build dependency tree based on package.json's` (https://github.com/DivanteLtd/vue-storefront/issues/4642).

scripts/release/const.ts
```diff
@@ -0,0 +1,4 @@
+export const base = '../../packages/';
```
andrzejewsky, line 1: path.resolve(__dirname, './../packages')

andrzejewsky: and uppercase if they are  constant

scripts/release/getPackageType.ts
```diff
@@ -0,0 +1,50 @@
...
+import { base } from './const';
+
+const integrationWrapperPackages = [
+  'api-client',
+  'composables',
+  'theme'
+];
+
+const getDirectories = (source: string) =>
+  fs.readdirSync(source, { withFileTypes: true })
```
andrzejewsky, line 13: seems like you always use`path.resolve(__dirname, `${base}${pckg}`)` so put it here and use just `const directories = getDirectories(pckg);`

scripts/release/const.ts
andrzejewsky: also, you have `types.ts` file that contains similar content actually. maybe mixt these two files together and call them `constants.ts`?

andrzejewsky:
@Fifciu @filrak guys, I think there is a mistake in the description @filrak - we should be able to publish packages **by name** not by directory so eg. `@vue-storefront/commercetools` instead of `commercetooos/composables`

scripts/release/const.ts
Fifciu: Sure

scripts/release/getPackageType.ts
Fifciu: Done

scripts/release/constants.ts
```diff
@@ -0,0 +1,35 @@
...
+enum PACKAGE_SUBTYPE {
+    INDEPENDENT = 0,
+    API,
+    COMPOSABLE,
+    THEME,
+    CLI
+}
+
+enum RELEASE_GRADATIONS {
+    path = 1,
```
andrzejewsky, line 19: please next time try to avoid using same names as the imported packages

scripts/release/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  return base.join('/');
+};
+
+const pckgCustomSubpackageToPath = (pckg: string, subpackage: string): string => {
+  const base = pckg.split('/');
+  base[base.length - 1] = subpackage;
+  return base.join('/');
+};
+
+const buildPackageDependencyList = (pckg: string): PACKAGE_SUBTYPE[] => {
```
andrzejewsky, line 60:
```
const buildPackageDependencyList = (pckg: string): PACKAGE_SUBTYPE[] => {
  const packageSubtype = getPackageSubtype(pckg);
  if (packageSubtype === PACKAGE_SUBTYPE.INDEPENDENT) {
    return [];
  }

  const enumKeys = Object.keys(PACKAGE_SUBTYPE)
    .filter(key => isNaN(Number(key)))

  return Array.from(enumKeys.keys());
};
```

andrzejewsky: but please check

scripts/release/index.ts
```diff
@@ -0,0 +1,150 @@
...
+    list.push(i);
+  }
+
+  return list;
+};
+
+const program = () => {
+  const args = process.argv.slice(2);
+  let pckg = args[0];
+  // args[1] === '-'
```
andrzejewsky, line 78: remove comment

scripts/release/index.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+    }
+
+    console.log('\n');
+    for (const log of statusLog) {
+      console.log(log);
+    }
+
+  } else if (packageType === PACKAGE_TYPES.Wrapper && pckg === 'core') {
+    console.log('Pick specific core\'s package');
```
andrzejewsky, line 144: @filrak shouldn't we use `consola` for such a thing?

scripts/release/packageTypes.ts
```diff
@@ -0,0 +1,52 @@
...
+  fs.readdirSync(`${BASE}/${source}`, { withFileTypes: true })
+    .filter(dirent => dirent.isDirectory())
+    .map(dirent => dirent.name);
+
+const isPackage = (pckg: string) => fs.existsSync(`${BASE}/${pckg}/package.json`);
+
+const isIntegrationWrapper = (pckg: string) => {
+  try {
+    const directories = getDirectories(pckg);
+    return integrationWrapperPackages.length <= directories.length && integrationWrapperPackages.every(directory => isPackage(`${pckg}/${directory}`));
```
andrzejewsky, line 20: please `integrationWrapperPackages.every(directory => isPackage(`${pckg}/${directory}`)` move to the separate function and call it maybe `hasPackage(pckg)` or something because this line is really loong (hard do read it)

Fifciu: Sure

scripts/release/index.ts
Fifciu: I wanted to show developer why I use args[0] and args[2] but not [1]. Isn't it ok?

scripts/release/constants.ts
```diff
@@ -0,0 +1,35 @@
...
+
+interface OperationsData {
+    pathsToRun: Array<string>;
+    freshVersions: Record<string, any>;
+    oldFiles: Record<string, any>;
+}
+
+const BASE = path.resolve('../packages/');
+const REGISTRY = 'https://registry.npmjs.org';
+// const REGISTRY = 'http://localhost:4873';
```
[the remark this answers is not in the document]
Fifciu: It is for manual testing on local registry. I can remove it

scripts/release/constants.ts
Fifciu: I will

scripts/release/index.ts
Fifciu: It won't work, look at this part:
```
for (let i = packageSubtype + 1; i < enumKeys.length; i++) {
```
It does not use whole array. It begins from `packageSubtype + 1`. I want to create array of values from X to Y. Like: [X, X+1, X+2, Y-1. Y] (in example Y = X+4). Your example would create array from 0 to Y but always X > 0.

scripts/release/packageTypes.ts
Fifciu: I just created `hasIntegrationWrapperPackages` function

scripts/release/index.ts
andrzejewsky: > I want to create array of values from X to Y.

```
Array.from([1, 2, 3, 4, 5].keys()).slice(X)
```

scripts/release/index.ts
Fifciu: Ok, I will remove it

scripts/release/index.ts
filrak: we can keep console.log - this script is just for us

Fifciu: We would prefer `consola` normally. However, as this script will be only for Core Team - it does not matter

PR #4648, opened 2020-07-21
Next: Checkout com - Save cards

Fifciu:
Draft of saving cards. It properly saves & removes cards but it cannot place an order with stored card because of server error. We are waiting for cko guys to fix it. After that, this PR should do the job without any changes

### DIFF
--- a/packages/checkout-com/README.md
+++ b/packages/checkout-com/README.md
@@ -46,7 +46,7 @@ interface {
     selectedPaymentMethod: CKO_PAYMENT_TYPE,
     savePaymentInstrument: boolean,
     storedPaymentInstruments: any[],
-    submitDisabled: boolean,
+    submitDisabled: ComputedRef<boolean>,
     loadAvailableMethods: (cartId: string, email?: string): { id, apms },
     initForm: (): void,
     submitCardForm: (): void,
--- a/packages/checkout-com/README.md
+++ b/packages/checkout-com/README.md
@@ -45,7 +45,7 @@ interface {
     error: Error | null,
     selectedPaymentMethod: CKO_PAYMENT_TYPE,
     savePaymentInstrument: boolean,
-    storedPaymentInstruments: any[],
+    storedPaymentInstruments: PaymentInstrument[],
     submitDisabled: ComputedRef<boolean>,
     loadAvailableMethods: (cartId: string, email?: string): { id, apms },
     initForm: (): void,
--- a/packages/checkout-com/README.md
+++ b/packages/checkout-com/README.md
@@ -76,7 +76,7 @@ const {
 ```
 
 3. `setBillingDetails` to save billing address. So you will be able to fetch `availableMethods` which base on your billing address (server-side)
-4. Run `loadStoredPaymentInstruments` for logged in user to load stored payment instruments. They will be loaded to `storedPaymentInstruments` array of `PaymentInstrument`. Caution: This interface is being used for storing credit cards currently. It might has different shape for stored different payment methods. Example of usage:
+4. Run `loadStoredPaymentInstruments` for logged in user to load stored payment instruments. They will be loaded to `storedPaymentInstruments` array of `PaymentInstrument`s. Caution: This interface is being used for storing credit cards currently. It might have different shape for stored different payment methods in the future. Interface:
 ```ts
 interface PaymentInstrument {
     id: string;
@@ -99,7 +99,9 @@ interface PaymentInstrument {
     fast_funds: string;
     payment_instrument_id: string;
 }
-
+```
+Example of usage:
+```ts
 if (isAuthenticated.value && cart.value && cart.value.customerId) {
     await loadStoredPaymentInstruments(cart.value.customerId);
 }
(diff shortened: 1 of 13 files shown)


Fifciu:
With that we are able to:
- Decide whether to save card or not
- Remove saved card
- Pay with saved card (it still requires 3ds, should it be like that?)

packages/checkout-com/nuxt/proxyMiddleware.js
```diff
@@ -0,0 +1,79 @@
...
+const apiRequestHeaders = (ckoSecretKey) => ({
+  headers: {
+    authorization: ckoSecretKey,
+    'Content-Type': 'application/json'
+  }
+});
+const getStoredMethods = async ({ ckoPublicKey, ckoSecretKey, customerId }) => {
+  try {
+    const { data } = await axios.get(
+      `https://play-commercetools.cko-playground.ckotech.co/merchants/${ckoPublicKey}/customers/${customerId}`,
```
andrzejewsky, line 26: that should configurable it can't be hardcoded so move It (`https://play-commercetools.cko-playground.ckotech.co`) to the config file

packages/checkout-com/nuxt/proxyMiddleware.js
```diff
@@ -0,0 +1,79 @@
...
+    return data;
+  } catch (err) {
+    console.log(err);
+    return null;
+  }
+};
+const removeStoredMethod = async ({ ckoPublicKey, ckoSecretKey, customerId, paymentInstrumentId }) => {
+  try {
+    return await axios.delete(
+      `https://play-commercetools.cko-playground.ckotech.co/merchants/${ckoPublicKey}/customers/${customerId}/payment-instruments/${paymentInstrumentId}`,
```
andrzejewsky, line 38: same  here

packages/checkout-com/nuxt/proxyMiddleware.js
```diff
@@ -0,0 +1,79 @@
+const axios = require('axios');
+const express = require('express');
+const app = express();
+
+let ckoPublicKey = null;
+let ckoSecretKey = null;
```
andrzejewsky, line 6: rename because of conflict names, or wrap it inside the object:

```
const config {
 ckoPublicKey: null,
 ckoSecretKey
}
```

and use it always  as config.XXX

packages/checkout-com/src/helpers.ts
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable */
+interface PaymentPropetiesWithoutToken {
+    context_id: string,
+    save_payment_instrument?: boolean,
+    secure3d?: boolean,
+    success_url?: string,
+    failure_url?: string
```
andrzejewsky, line 7: just pass here `token?` as optional field and call this interface `PaymentPropeties` simpler.

packages/checkout-com/src/helpers.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface PaymentPropetiesWithToken extends PaymentPropetiesWithoutToken {
+    token: string,
+}
+
+interface PaymentPropetiesWithOptionalToken extends PaymentPropetiesWithoutToken {
+    token?: string,
+}
+
+interface PaymentMethodPayload extends PaymentPropetiesWithoutToken {
```
andrzejewsky, line 18: same here, just `PaymentMethodPayload` with optional token

packages/checkout-com/src/useCkoCard.ts
```diff
@@ -1,47 +1,69 @@
...
 
 declare const Frames: any;
 
-const submitDisabled = ref(false);
+const savePaymentInstrument = ref(false);
+const isCardValid = ref(false);
 const error = ref(null);
+const paymentMethod = ref(0);
+const storedPaymentInstruments = ref([]);
+const submitDisabled = computed(() => {
```
andrzejewsky, line 15: `paymentMethod.value !== CKO_PAYMENT_TYPE.SAVED_CARD || isCardValid.value`

packages/commercetools/theme/pages/Checkout/Payment.vue
```diff
@@ -131,7 +131,7 @@
           :selected="chosenPaymentMethod.value"
```
andrzejewsky, line 131: if you are committing something to the `next` branch please don't modify theme (just don't commit these changes, keep them locally)

packages/checkout-com/nuxt/proxyMiddleware.js
Fifciu: const config = {
  publicKey: null,
  secretKey: null,
  ctApiUrl: null
}

packages/checkout-com/src/useCko.ts
```diff
@@ -1,12 +1,14 @@
...
 import { createContext } from './payment';
-import { Configuration } from './configuration';
+import { Configuration, getSaveInstrumentKey } from './configuration';
 import { ref } from '@vue/composition-api';
+import { CKO_PAYMENT_TYPE } from './helpers';
 import useCkoCard from './useCkoCard';
 
 const error = ref(null);
 const availableMethods = ref([]);
+const contextId = ref(null);
```
Qrzy, line 11: Type here would be helpful

packages/checkout-com/src/useCko.ts
```diff
@@ -20,26 +22,36 @@ interface PaymentMethodsConfig {
...
-  PAYPAL
-}
+const selectedPaymentMethod = ref(CKO_PAYMENT_TYPE.NOT_SELECTED);
 
-const selectedPaymentMethod = ref(PaymentMethod.NOT_SELECTED);
+const setSavePaymentInstrument = (newSavePaymentInstrument: boolean) => {
+  localStorage.setItem(getSaveInstrumentKey(), JSON.stringify(newSavePaymentInstrument));
+};
+const loadSavePaymentInstrument = () => (localStorage.getItem(getSaveInstrumentKey())
+  ? JSON.parse(localStorage.getItem(getSaveInstrumentKey()))
```
Qrzy, line 31: Forcing a oneliner here is not the best approach, especially when you load something from the storage.
Assign the result to variable and then check it to avoid loading it again.

packages/checkout-com/README.md
```diff
@@ -1,48 +1,118 @@
...
-interface {
-    error: Error | null,
-    submitDisabled: Boolean,
-    submitForm: async(),
-    makePayment: async({ cartId }),
-    initCardForm: (): void
+    selectedPaymentMethod: CKO_PAYMENT_TYPE,
+    savePaymentInstrument: boolean,
+    storedPaymentInstruments: any[],
+    submitDisabled: boolean,
```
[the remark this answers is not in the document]
Fifciu: You are right `submitDisabled` is:
```
    submitDisabled: ComputedRef<boolean>,
```
But `savePaymentInstrument` is just a `boolean`

packages/checkout-com/src/useCko.ts
Fifciu: It is string. However, you do not have access to it.

packages/checkout-com/README.md
```diff
@@ -72,57 +142,134 @@ const initForm = (initMethods: PaymentMethods = null, config: PaymentMethodsConf
...
+    removePaymentInstrument,
+    storedPaymentInstruments,
+    submitDisabled,
+    error
+} = useCko();
+```
+
+Currently, these are available payment methods:
+```ts
+enum CKO_PAYMENT_TYPE {
```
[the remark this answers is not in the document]
Fifciu: You are right. Proof: https://google.github.io/styleguide/javascriptguide.xml?showone=Naming#Naming

andrzejewsky:
please neither modify the original CT theme nor core one, remove it and it's ready to merge

PR #4658, opened 2020-07-22
Next: Update nuxt theme module

### LINKED ISSUE #4652: Update nuxt-theme module
We should make `nuxt-theme` module work in projects generated through `createProject`. To make that happen we should split core development and non-core development properties and behaviors.

1. The module config should look as following: 
```js
{
 ['@vue-storefront/nuxt-theme', {
  routes: true, // if true adds default routing to the app, defaults to true
  generate: { // works as current options, defaults to false
    replace: {
       apiClient: '@vue-storefront/commercetools-api',
       composables: '@vue-storefront/commercetools'
    }
  }
}]
```
2 Change `options.x` in EJS from a theme module to match new positions in config (eg `options.apiClient` is now `options.generate.replace.apiClient` )
3. (unrelated) Rename `theme-module` folder to `nuxt-theme-module` for consistency with `nuxt-module` dir
4. Add routes to the project only when `routes` property is set to `true`
5. Do not generate files in `theme`, do not change project dirs, to `theme`, do not watch files via chokidar if `generate` option is set to `false`
6. Generate following configs in `nuxt-config.js` 
- core dev
```js
 ['@vue-storefront/nuxt-theme', {
  generate: { // works as current options, defaults to false
    replace: {
       apiClient: '@vue-storefront/commercetools-api',
       composables: '@vue-storefront/commercetools'
    }
  }
}]
```
- generated project
```
 ['@vue-storefront/nuxt-theme']
```
7. Check if cli-generated project is working within a monorepo (packages are not released so you have to test here). If not write about issues.
8. Remove `global.coreDev` from `@vue-storefront/nuxt-module` as condition and replace it with `generate` option

Fifciu:
## How to test
1. Prepare name for your create project directory. Let it be `new-project-test`
2. Open `package.json`
3. Add to scripts:
```
"test-cli": "cd packages/core/cli/src/scripts/createProject && ts-node index.ts"
```
4. Add to `workspaces.packages`:
```
"packages/new-project-test"
```
5. Open `packages/core/cli/src/scripts/createProject/index.ts`:
6. Add at the very bottom: 
```
createProject('commercetools', '../../../../../new-project-test')
```
7. Run `yarn test-cli` from root directory
8. `packages/new-project-test` should appear
9. Change name in package.json of new package to prevent conflict with existing in monorepo theme
10. Run `yarn` in root directory
11. `yarn dev` inside new package should work

## Things I tested
1. Unit tests
2. Core development with CT
3. Client development with created CT project

Closes #4652


packages/core/cli/__tests__/createProject/uncommentProjectOnly.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+jest.mock('fs', () => ({
+  readFileSync: () => fileContent,
+  writeFileSync: jest.fn()
+}));
+
+import { writeFileSync } from 'fs';
+
+describe('[vsf-next-cli] uncommentProjectOnly', () => {
+  it('removes comments from project on', () => {
```
filrak, line 16: i think you missed something here ;)

packages/core/cli/src/scripts/createProject/index.ts
```diff
@@ -16,6 +17,7 @@ async function createProject(integration: string, targetPath: string): Promise<v
   const absoluteTargetPath = path.join(__dirname, targetPath);
   const nuxtConfigPath = path.join(absoluteTargetPath, 'nuxt.config.js');
   await removeMagicCommentsFromFile(nuxtConfigPath);
```
filrak, line 19: **suggestion** shouldn't both (with `uncommentprojectOnly` be a part of a same fn? or at least same file

packages/core/nuxt-theme-module/index.js
```diff
@@ -19,58 +19,78 @@ const log = {
 module.exports = async function DefaultThemeModule(moduleOptions) {
   log.info(chalk.green('Starting Theme Module'));
 
-  const agnosticThemeDir = path.join(__dirname, 'theme');
-  const projectLocalThemeDir = this.options.buildDir.replace('.nuxt', '.theme');
+  if (typeof moduleOptions.routes === 'undefined') {
```
filrak, line 22: **code organization** I think it would be better to just have a default config and merge it with provided one (via `lodash/merge`). Otherwise we will have to make this check for each "default" property value

packages/core/nuxt-theme-module/index.js
```diff
@@ -19,58 +19,78 @@ const log = {
...
+            apiClient: moduleOptions.generate.replace.apiClient,
+            composables: moduleOptions.generate.replace.composables
+          }
+        }
       });
   };
+  let projectLocalThemeDir;
+  let themeDirectoriesPaths;
+
+  if (moduleOptions.generate) {
```
filrak, line 47: **code organization** how about externalizing al generate-related code to separate file and just using `apply` || `call` to pass the context and `moduleOptions`  (+ other required) as arguments

packages/core/nuxt-theme-module/index.js
Fifciu: I agree

packages/core/cli/src/scripts/createProject/index.ts
Fifciu: For file, ok but I am not sure about the function

filrak: I meant that dealing with magic comments should in createProject can be a single step 9represented as fn) but this step could contain 2 functions. However you do this my point is that these 2 features should be in the same place for maintainability reasons. If one will change the second one will most likely have to change as well

packages/core/cli/__tests__/createProject/uncommentProjectOnly.spec.ts
Fifciu: Good eye! I've just changed it to: `uncomments parts inside "project only" comments`

packages/core/cli/src/scripts/createProject/index.ts
Fifciu: I decided to merge these 2 files into `processMagicComments.ts` which contains `processMagicComments` function which just runs these 2

packages/core/nuxt-theme-module/index.js
Fifciu: Done

packages/core/nuxt-theme-module/index.js
Fifciu: I separated it and load only when `moduleOptions.generate`

packages/checkout-com/nuxt/index.js
```diff
@@ -3,9 +3,7 @@ import path from 'path';
 export default function CheckoutComModule(moduleOptions) {
```
filrak, line 3: does it have anything to do with CKO?

packages/core/cli/src/scripts/createProject/processMagicComments.ts
```diff
@@ -0,0 +1,29 @@
+import fs from 'fs';
+
+const removeMagicComments = async (absoluteFilePath: string): Promise<void> => {
```
filrak, line 3: `removeProjectOnly` would be more descriptive ;)

Fifciu: Are you sure? It is about `// core-development-start & end`. Below we have one from removing `Project only`

packages/checkout-com/nuxt/index.js
Fifciu: Damn. Looks like I switched branch to old one while I was working on the next task by mistake. I will fix it first thing tomorrow

Fifciu: Fixed

packages/core/nuxt-theme-module/package.json
```diff
@@ -18,5 +18,10 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
+    "@vue/composition-api": "^1.0.0-beta.6",
+    "lodash.merge": "^4.6.2",
+    "vue": "^2.6.11"
```
filrak, line 25: for what we need vue here?

Fifciu:
## Todo
- [ ] Readme

PR #4678, opened 2020-07-24
Add nuxt composition api

andrzejewsky:
closes: https://github.com/DivanteLtd/vue-storefront/issues/4668


packages/about-you/theme/nuxt.config.js
```diff
@@ -27,6 +27,7 @@ export default {
   buildModules: [
     // to core
     '@nuxt/typescript-build',
+    'nuxt-composition-api',
```
filrak, line 30: should be added via `@vue-storefront/nuxt`

PR #4689, opened 2020-07-29
Next: Missing features cko

### LINKED ISSUE #4656: [Next] Missing features in basic checkout.com integration
- [x] Nuxt module passes only `publicKey` to setup function, thus it's not possible to set custom styles, webhook URL or tokenized card key (see https://github.com/DivanteLtd/vue-storefront/blob/d572e03c37454bea2d623ed9dd1d6a4bc7e9cd3f/packages/checkout-com/src/configuration.ts#L36 and https://github.com/DivanteLtd/vue-storefront/blob/d572e03c37454bea2d623ed9dd1d6a4bc7e9cd3f/packages/checkout-com/nuxt/plugin.js#L5)

- [x] Config is missing the `localization` option that allows to set custom placeholders.

- [x] Let's allow to pass all the options above also through the `initForm` function - this is important especially for localization.

- [x] There are no sourcemaps in the package, please add them via rollup config.

- [x] Sdk will be autoloaded based on the config

- [x] Let's allow to fetch all available payment methods from cko

- [x] Readme update

Fifciu:
Closes #4656


packages/checkout-com/src/useCkoCard.ts
```diff
@@ -14,7 +14,15 @@ const setCardToken = (token) => localStorage.setItem(getCardTokenKey(), token);
 const removeCardToken = () => localStorage.removeItem(getCardTokenKey());
 
 const useCkoCard = () => {
-  const makePayment = async ({ cartId }) => {
+  const fetchAvailableMethods = async (reference) => {
```
Qrzy, line 17: I'm not sure if I get this part right.
Is the return value of this call locale-independent? I can't find the context creation part in cko docs, thus the question.
Either way, wouldn't it be better to create this context when `useCkoCard` is called and returnavailable methods as readonly prop? 

By the way - considering this one bit is added - is this composable still card-related only or should it be renamed from `useCkoCard` to `useCko`?

packages/checkout-com/nuxt/index.js
```diff
@@ -1,13 +1,39 @@
...
+  const paymentMethods = {
+    ...defaultPaymentMethods,
+    ...(moduleOptions.paymentMethods ? moduleOptions.paymentMethods : {})
+  };
+
+  for (const [paymentMethod, enabled] of Object.entries(paymentMethods)) {
+    if (enabled && paymentMethodSdk[paymentMethod] && !isScriptInArray(scripts, paymentMethodSdk[paymentMethod])) {
+      scripts.push({
+        src: paymentMethodSdk[paymentMethod]
+      });
```
Qrzy, line 33: Shouldn't this have at least `async` attribute?

Fifciu: Sure it could

packages/checkout-com/nuxt/index.js
```diff
@@ -1,13 +1,39 @@
...
     src: path.resolve(__dirname, './plugin.js'),
-    options: {
-      publicKey: moduleOptions.publicKey
-    }
+    options: moduleOptions
   });
 
+  const paymentMethods = {
+    ...defaultPaymentMethods,
+    ...(moduleOptions.paymentMethods ? moduleOptions.paymentMethods : {})
```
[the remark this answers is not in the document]
Fifciu: Right

packages/checkout-com/src/useCkoCard.ts
Fifciu: We need billing address on the server to create a context. So I think autosending create context request might be wrong way for some devs.

About naming, yes, maybe we should separate them to 2 composables `useCkoCard` and `useCko` - for fetchAvailableMethods.

Tell us what do you think @andrzejewsky

Qrzy: One more suggestion:

From DX perspective it would be good to get here api similar to `useProduct` for example.
Meaning `useCko` would return readonly `availableMethods` and a method `loadMethods` that would populate the former.
This way dev wouldn't have to care about assigning the result of fetching available methods.
What do you think?

andrzejewsky: first thing - we shouldn't load payment methods inside of composable that means payment method  itself 😃 @Fifciu let's discuss about it on the call

Fifciu:
Guys, I was thinking about the whole integration and I had to change a few things. Please rereview it :D

packages/checkout-com/README.md
```diff
@@ -6,51 +6,101 @@
...
 ```js
-import { useCkoCard } from '@vue-storefront/checkout-com';
+import { useCko, useCkoCard } from '@vue-storefront/checkout-com';
 ```
 
-2. `useCkoCard` returns:
+2. `useCko` returns:
+```ts
+interface {
+    availableMethods: { name: string, [key: string]: any },
```
Qrzy, line 18: This collides with the last lines of this README, that say `availableMethods` is just an array of strings.

packages/checkout-com/src/useCko.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const selectedPaymentMethod = ref(PaymentMethod.NOT_SELECTED);
+
+const useCko = () => {
+  const loadAvailableMethods = async (reference) => {
+    try {
+      const response = await createContext({ reference });
+      availableMethods.value = [
+        ...response.data.apms,
+        { name: 'card' }
```
Qrzy, line 39: And here yet another format. So finally, `availableMethods` is:
```ts
{ name: string, [prop]:string }[]
```
Correct?

packages/checkout-com/README.md
Fifciu: Ok, I fixed it at the bottom of the readme

packages/checkout-com/src/useCko.ts
Fifciu: It is:
```
{ name: string, [key: string]: any }
```

packages/checkout-com/README.md
```diff
@@ -6,51 +6,101 @@
...
-const { submitForm, submitDisabled, initForm } = useCkoCard();
+const { cart } = useCart();
+const { setBillingDetails } = useCheckout();
+const { submitDisabled } = useCkoCard();
+const { initForm, loadAvailableMethods, availableMethods } = useCko();
+```
+
+3. `setBillingDetails` to save billing address. So you will be able to fetch `availableMethods` which base on your billing address (server-side)
+4. Run `loadAvailableMethods` - it will return `interface { id, apms: Array<any> }` and set `apms` inside `availableMethods`
+5. Execute `initForm`. It mounts different payment handlers depends on arguments (check details below). If you are calling it after load component - **use `onMounted` to make sure DOM Element where it will be mounted already exists**.
```
filrak, line 45: where does this dom elements exist? where we're adding this

PR #4693, opened 2020-07-29
add ssrRef for category and product

andrzejewsky:
closes: https://github.com/DivanteLtd/vue-storefront/issues/4669 https://github.com/DivanteLtd/vue-storefront/issues/4670

### DIFF
--- a/packages/core/core/src/factories/useCategoryFactory.ts
+++ b/packages/core/core/src/factories/useCategoryFactory.ts
@@ -1,6 +1,6 @@
 import { UseCategory } from '../types';
-import { useSSR } from '../utils';
-import { ref, Ref, computed } from '@vue/composition-api';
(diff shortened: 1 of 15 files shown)


packages/core/core/src/factories/useCategoryFactory.ts
```diff
@@ -1,6 +1,6 @@
 import { UseCategory } from '../types';
-import { useSSR } from '../utils';
-import { ref, Ref, computed } from '@vue/composition-api';
+import { Ref, computed } from '@vue/composition-api';
+import { ssrRef } from 'nuxt-composition-api';
```
filrak, line 3: this (and all others) has to be abstracted to work on non-nuxt environments as well.  We can also just use `if` to check if current env is nuxt or not and use `ssrRef` or `ref`

PR #4711, opened 2020-08-06
[Docs] Getting Started about Nuxt Modules

Fifciu:
Added `Nuxt installation` part about nuxt modules for both AYC & CT

### DIFF
--- a/packages/core/docs/core/getting-started.md
+++ b/packages/core/docs/core/getting-started.md
@@ -30,7 +30,17 @@ During the installation process you will be asked questions about:
 
 Based on provided answers Vue Storefront CLI will generate a project already integrated with the services that you have chosen.
 
-### Manual installation
+### Nuxt project installation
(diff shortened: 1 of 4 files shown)


packages/core/docs/core/getting-started.md
```diff
@@ -30,11 +30,17 @@ During the installation process you will be asked questions about:
 
 Based on provided answers Vue Storefront CLI will generate a project already integrated with the services that you have chosen.
 
-### Manual installation
+### Nuxt project installation
```
filrak, line 34: worth mentioning `@vue-storefront/nuxt` module as well

Fifciu: Ok, I mentioned all 4 modules. However, we do not have Shopify nuxt module. Should I add it? If so, should it be part of this PR or another one? @filrak

PR #4714, opened 2020-08-07
Next: CLI - Create Project

### LINKED ISSUE #4703: Create a CLI tool for VSF Next
1. Create a new project in `core` folder called `cli`
2. The CLI should be runnable via `init <projectname>` (`projectname` should determine npm project name and folder)
3. After that you should see a list of available backends (`inquirer.js`) https://www.npmjs.com/package/inquirer
4. Once the backend is chooses generate a project based on  agnostic and integration themes

Fifciu:
closes #4703 
Command: `yarn cli init <projectName>`


packages/core/cli/src/commands/init.ts
```diff
@@ -0,0 +1,25 @@
+import log from '../utils/log';
+import getIntegrationsFromPackage from '@vue-storefront/cli/src/utils/getIntegrationsFromPackage';
+import inquirer from 'inquirer';
+import createProject from '../scripts/createProject';
+import path from 'path';
+
+export default async ([projectName]) => {
+  if (!projectName) {
```
filrak, line 8: i think that could be just a first question if not provided (like in vue cli)

packages/core/cli/src/utils/getIntegrationsFromPackage.ts
```diff
@@ -0,0 +1,23 @@
+import path from 'path';
+import fs from 'fs';
+
+const notAnIntegration = [
```
filrak, line 4: I think it would be safer to have inclusive list, not exclusive - for external maintainers its not necessarily obvious that adding any other VSF package will make it an integration

packages/core/cli/src/commands/init.ts
Fifciu: Just added it with unit test

packages/core/cli/src/utils/getIntegrationsFromPackage.ts
Fifciu: Ok, now it is just a:
```ts
export default (): string[] => [
  'about-you',
  'commercetools'
];
```

Fifciu:
Added readme

packages/core/cli/src/utils/getIntegrationsFromPackage.ts
```diff
@@ -0,0 +1,4 @@
+export default (): string[] => [
```
filrak, line 1: AYC shouldn't be available yet (its not working correctly). I'd put commercetools only for now until we check if shopify is working, also filename is not accurate anymore ;)

PR #4739, opened 2020-08-13
Next: CKO hotfix cko params

Fifciu:
Fix for sending configuration to `Frames`. Currently, we need double nested `card` attribute. With this fix, it will work as it supposed to


andrzejewsky (approved):
@Fifciu please fix test / build and it's ready to merge

PR #4753, opened 2020-08-19
SSR implementation

andrzejewsky:


packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -38,20 +38,15 @@ interface UseCartFactory<CART, CART_ITEM, PRODUCT, COUPON> {
...
   factoryParams: UseCartFactoryParams<CART, CART_ITEM, PRODUCT, COUPON>
 ): UseCartFactory<CART, CART_ITEM, PRODUCT, COUPON> => {
-  const appliedCoupon: Ref<COUPON | null> = ref(null);
-  const loading: Ref<boolean> = ref<boolean>(false);
-  const cart: Ref<CART> = ref(null);
-  let isInitialized = false;
 
   const setCart = (newCart: CART) => {
-    cart.value = newCart;
+    getShared('useCart-cart').value = newCart;
```
filrak, line 43: I really don't like this construction. Basing refs on keys is nether close to the native implementation nro DX-friendly (no autosuggestion, no clear information on from where the import is coming etc).

Also it's misleading since getter is implying a read-only property

 Can we somehow change it to regular imports?

filrak (changes requested):
Even though I like the solution the APIs are adding a lot of unnecessary complexity that's far away from a standard way of using CAPI 9therefore it will be harder to migrate to native APIs once we will move to Vue 3).

- `getShared` is key-based which is a terrible DX, also API is misleading
- `ssrRef` will have a name collisions with Nuxt CAPI

packages/core/nuxt-theme-module/index.js
```diff
@@ -35,6 +36,10 @@ module.exports = async function DefaultThemeModule(moduleOptions) {
     });
   }
 
+  this.extendBuild(config => {
```
filrak, line 39: why its moved from nuxt module to theme module? theme module is optional for projects that are not using default theme

andrzejewsky: I remember that something didn't work without that - I'll check if it's still necessary

packages/core/core/src/factories/useCartFactory.ts
andrzejewsky: yeah.. don't like it too - I have an idea for having just `sharedRef` - so let's improve

Fifciu (changes requested):

filrak (changes requested):

PR #4768, opened 2020-08-22
fix mem leak in dynamic config

tomasz-duda:


pkarw (approved):
Great job thanks!

RakowskiPrzemyslaw (approved):

Fifciu:
I just did simple test with `dev:inspect` and looks like it works. Good job!

Fifciu (approved):

core/scripts/server.ts
```diff
@@ -280,47 +282,43 @@ app.get('*', (req, res, next) => {
...
-          })
+          }))
         }
       })
-    } else {
-      config = require('config') // reload config
-      dynamicCacheHandler()
+      await Promise.all(configPromises)
+
+      if (!config.server.dynamicConfigReloadWithEachRequest) {
```
gibkigonzo, line 313: I understand it as if it is true then config will be reloaded with each request. But we check here negation and then reload config. So shouldn't it be 
```
      if (config.server.dynamicConfigReloadWithEachRequest) {
        globalContextConfig = config.util.extendDeep({}, requestContextConfig)
      }
```
and do we need extra config ? 🤔 Shouldn't `dynamicConfigReload ` handle that?

gibkigonzo (approved):

PR #4787, opened 2020-08-27
Draft: CKO Paypal MVP

Fifciu:
Still have to:
- [x] Write readme
- [x] Merge and rebase: https://github.com/DivanteLtd/vue-storefront/pull/4748
- [x] Check for Customer


packages/checkout-com/nuxt/proxyMiddleware.js
```diff
@@ -33,6 +33,9 @@ const getStoredMethods = async ({ customerId }) => {
     return data;
   } catch (err) {
     console.log(err);
+    if (err.response) {
+      console.log(err.response.data);
```
Qrzy, line 37: Again, I think it's better to `console.error` errors.

yarn-error.log
```diff
@@ -0,0 +1,16171 @@
+Arguments:
```
Qrzy, line 1: Let's get rid of this one. :)

Qrzy (approved):

packages/checkout-com/README.md
```diff
@@ -164,7 +164,10 @@ const payment = await makePayment({ cartId: cart.value.id, email: user.value &&
 // If you've already loaded available payment methods with same useCko composable instance
 const payment = await makePayment();
 
-if (!payment) return;
+if (error.value) {
+    console.log(error.value);
```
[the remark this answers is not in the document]
Fifciu: Also changed to `error.value.message` as error.value is instance of `Error`

packages/checkout-com/src/useCko.ts
```diff
@@ -103,16 +102,11 @@ const useCko = () => {
...
-      };
     } else if (selectedPaymentMethod.value === CkoPaymentType.PAYPAL) {
-      finalizeTransactionFunction = () => {
-        console.log('Making transaction with PayPal...');
-      };
+      finalizeTransactionFunction = makePaypalPayment;
+      localError = paypalError;
     } else {
-      error.value = 'Not supported payment method';
+      error.value = new Error('Not supported payment method');
```
andrzejewsky, line 109: why you are returning a new error instead of string?

andrzejewsky (changes requested):

packages/checkout-com/src/useCko.ts
Fifciu: When we have `try{...}catch(e){error.value=e;}` parts - `e` is instance of `Error`. I wanted to unify errors in whole package

andrzejewsky (approved):

PR #4821, opened 2020-09-02
Reduce initial client-side bundle-size by lazy-loading `i18n` translations

### LINKED ISSUE #4813: Reduce initial client-side bundle-size by lazy-loading `i18n` translations

If you  are using a VSF shop with a lot of store-views, the `app.js` inside the bundle will automatically blow up because of the many different translation files it's possibly needs.

Its' caused by the dynamic import inside [`core/i18n/index.ts`](https://github.com/DivanteLtd/vue-storefront/blob/7ba4b91020f64333ef5f85c725edab7e63d1c9ec/core/i18n/index.ts#L52-L55). As a [dynamic expression import](https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import) is used, webpack needs to include all possible JSON files beneath the import path inside the bundle, to be able to load them later. If you have like 10 different languages, this leads to 9 unneeded, but loaded, language files and a package-size increase of around 500k+ (depending on the number of translations you are using) with a lot of initially unused code.
As we never know how much and which languages are needed at build-time this is a common way to approach but brings this flaw with it.

I've already tried to chunk the language includes (using webpack) to decrease the initial package-size of `app.js` but the unneeded languages will still be loaded in prefetch/preload requests and would still cause unused code.

I would suggest that we include an option to use `fetch` to load the languages (instead of `import`), this way users can reduce their package size and only load languages they actually need. I've already tested a solution where I copied the files using `CopyWebpackPlugin` and loaded them using `isomorphic-fetch` – these are problems and drawbacks to consider:
* We'd make two requests (on SSR and CSR) to fetch the JSON file – maybe they should be in state-management or a global to save this request when the language is already loaded.
* For SSR request we need an absolute URL to use the `fetch`, therefore you need another config variable including a base-path.

I could also imagine to use this issue to build some kind of translation handler to include your custom translation sources, like an CMS or custom API endpoint. This way we could achieve the above and add some flexibility.

What do you say about this? – any better Ideas?

**NOTE:** We are not using `bundleAllStoreviewLanguages` in this usecase ;)

## Version of Vue Storefront

- [x] Vue Storefront
- [ ] Vue Storefront Next

- [x] YES
- [ ] NO

cewald:
### Related Issues

closes #4813

> If you are using a VSF shop with a lot of store-views, the app.js inside the bundle will automatically blow up because of the many different translation files it's possibly needs.

More detailed informations are in the issue #4813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.13.0] - UNRELEASED
+
+### Added
+- Reduce initial client-side bundle-size by lazy-loading `i18n` translations - @cewald (#4821)
+
 ## [1.12.2] - 2020.07.28
 
 ### Added
@@ -15,7 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   More info about helmet configuration https://helmetjs.github.io/docs/
 - Add config `users.tokenInHeader` which allows to send token in header instead in query. Require to set on true same config in vsf-api.
 - Make calculation of bundled products price by options optional - @cewald (#4556)
-- Reduce initial client-side bundle-size by lazy-loading `i18n` translations - @cewald (#4821)
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   More info about helmet configuration https://helmetjs.github.io/docs/
 - Add config `users.tokenInHeader` which allows to send token in header instead in query. Require to set on true same config in vsf-api.
 - Make calculation of bundled products price by options optional - @cewald (#4556)
+- Reduce initial client-side bundle-size by lazy-loading `i18n` translations - @cewald (#4821)
```
gibkigonzo, line 18: can you move it to ?
```
## [1.13.0] - UNRELEASED

### Added
- Reduce initial client-side bundle-size by lazy-loading `i18n` translations - @cewald (#4821)
```

cewald: Latest was on 1.12.2 – I added 1.13.0 or is there another integration branch?

gibkigonzo (approved):

CHANGELOG.md
gibkigonzo: its perfect 👌 thank you

Issue #4616, opened 2020-07-13
appliedCoupon is null if you load a cart that already contains applied coupon (discountCode)


EugeneHlushko:
## Current behavior

Currently, appliedCoupon is only set if you had applied it after loading a cart

## Expected behavior

* appliedCoupon is available also if you had loaded a cart that has an applied discount code
* need to change code in [`removeCoupon`](https://github.com/DivanteLtd/vue-storefront/blob/next/packages/core/core/src/factories/useCartFactory.ts#L134) to make it work to:

```diff
- coupon: appliedCoupon.value
+ coupon: appliedCoupon
```

Because api-client package expects full coupon object and not code string value

* [`coupon: computed(() => appliedCoupon.value)` from `useCart`](https://github.com/DivanteLtd/vue-storefront/blob/next/packages/core/core/src/factories/useCartFactory.ts#L152) should be fixed by fixing `appliedCoupon`

## Steps to reproduce the issue

* Add a coupon to your cart, reload a page, see appliedCoupon being null although cart has it under `cart.value.discountCodes[0]`

## Can you handle fixing this bug by yourself?
* [ ] YES
* [ ] NO

## Which [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html) state this refers to? Info for developer.
Pick one option.

* [ ] This is a bug report for test version on https://test.storefrontcloud.io - In this case Developer should create branch from `develop` branch and create Pull Request `2. Feature / Improvement` back to `develop`.
* [ ] This is a bug report for current Release Candidate version on https://next.storefrontcloud.io - In this case Developer should create branch from `release` branch and create Pull Request `3. Stabilisation fix` back to `release`.
* [ ] This is a bug report for current Stable version on https://demo.storefrontcloud.io and should be placed in next stable version hotfix - In this case Developer should create branch from `hotfix` or `master` branch and create Pull Request `4. Hotfix` back to `hotfix`.

## Environment details

* Browser: 
* OS: 
* Node: 
* Code Version: 

## Additional information

┆Issue is synchronized with this [Jira Story](https://vsf.atlassian.net/browse/NEXT-185) by [Unito](https://www.unito.io/learn-more)

PR #4850, opened 2020-09-09
Next: Applied coupons fixes

Fifciu:
I made `coupon` from `useCart` persistent. It has proper value in CT after refresh. It solves #4616 in 100%.

`coupon` looks like:
image

I changed the call signature for applyCoupon & removeCoupon to make it work properly

┆Issue is synchronized with this [Jira Story](https://vsf.atlassian.net/browse/NEXT-270) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/packages/core/docs/commercetools/changelog.md
+++ b/packages/core/docs/commercetools/changelog.md
@@ -9,3 +9,5 @@
 
 * **factories**: refactored setup using apiClientFactory (#4777)
 * **factories**: updated return signature of loadCart ([#4616](https://github.com/DivanteLtd/vue-storefront/issues/4616))
+* **factories**: updated call arguments of applyCoupon ([#4616](https://github.com/DivanteLtd/vue-storefront/issues/4616))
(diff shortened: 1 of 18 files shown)


filrak (commented):
the comvention for `currentCoupon` is to be named 'coupon' in other places (`removeCoupon`). Perhaps its better to name it `currentCoupon` though in there. In that case you should also correct all the integrations that are making use of it and add it to the changelog

packages/core/docs/commercetools/changelog.md
```diff
@@ -7,4 +7,7 @@
 
 ### Changes
 
-- refactored setup using apiClientFactory (#4777)
+* **factories**: refactored setup using apiClientFactory (#4777)
+* **factories**: updated return signature of loadCart ([#4616](https://github.com/DivanteLtd/vue-storefront/issues/4616))
```
filrak, line 11: if its a single PR lets make it a single row

filrak (approved):

PR #4855, opened 2020-09-11
Next: Applied coupons fixes - different way

Fifciu:
Applied coupons fixes - different way

┆Issue is synchronized with this [Jira Story](https://vsf.atlassian.net/browse/NEXT-273) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/packages/about-you/api-client/package.json
+++ b/packages/about-you/api-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vue-storefront/about-you-api",
-  "version": "0.1.1",
+  "version": "0.1.0",
   "private": true,
   "sideEffects": false,
   "main": "lib/index.cjs.js",
--- a/packages/about-you/api-client/package.json
+++ b/packages/about-you/api-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vue-storefront/about-you-api",
-  "version": "0.1.1",
+  "version": "0.1.0",
   "private": true,
   "sideEffects": false,
   "main": "lib/index.cjs.js",
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "@aboutyou/backbone": "^7.0.0",
-    "@vue-storefront/core": "^2.0.6"
+    "@vue-storefront/core": "^2.0.4"
   },
   "files": [
     "lib/**/*"
--- a/packages/about-you/api-client/package.json
+++ b/packages/about-you/api-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vue-storefront/about-you-api",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "private": true,
   "sideEffects": false,
   "main": "lib/index.cjs.js",
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "@aboutyou/backbone": "^7.0.0",
-    "@vue-storefront/core": "^2.0.4"
+    "@vue-storefront/core": "^2.0.6"
   },
   "files": [
     "lib/**/*"
--- a/packages/commercetools/composables/src/useCart/index.ts
+++ b/packages/commercetools/composables/src/useCart/index.ts
@@ -8,12 +8,13 @@ import {
 import { ProductVariant, Cart, LineItem } from './../types/GraphQL';
 import loadCurrentCart from './currentCart';
 import { useCartFactory, UseCartFactoryParams} from '@vue-storefront/core';
+import { getCoupon } from '../getters/cartGetters';
 
 const getBasketItemByProduct = ({ currentCart, product }) => {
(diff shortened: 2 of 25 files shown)


Qrzy:
Is introducing `AgnosticCoupon` a good idea here? The coupon may include additional informations that would not be available unless developer makes his own getter.

packages/core/core/src/types.ts
```diff
@@ -255,6 +254,12 @@ export interface AgnosticCategoryTree {
   [x: string]: unknown;
 }
 
+export interface AgnosticCoupon {
+  id: string;
+  name: string;
+  code: string;
```
andrzejewsky, line 260: please add `value` and set it to null in CT for now, but please create an issue for that, to implement that field in CT

andrzejewsky (approved):

packages/core/core/src/types.ts
Fifciu: Done

filrak:
conflicts in changelog @Fifciu

filrak:
@Qrzy dev will still have access to the cart object

Fifciu:
Gonna fix it tomorrow @Fifciu

EugeneHlushko:
#4616

packages/commercetools/composables/src/useCart/index.ts
```diff
@@ -38,15 +39,16 @@ const params: UseCartFactoryParams<Cart, LineItem, ProductVariant, any> = {
     const updatedCart = await apiApplyCartCoupon(currentCart, coupon);
     return { updatedCart: updatedCart.data.cart, updatedCoupon: coupon };
   },
-  removeCoupon: async ({ currentCart, coupon }) => {
-    const updatedCart = await apiRemoveCartCoupon(currentCart, coupon);
+  removeCoupon: async ({ currentCart }) => {
+    const { id: couponId } = getCoupon(currentCart);
```
filrak, line 43: what if we will change the way getter works? I don't think it's safe to use getters inside composable factories - they're prone to changes

packages/about-you/api-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vue-storefront/about-you-api",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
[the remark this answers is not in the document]
Fifciu: Indeed. I will revert changes in package.json's

packages/commercetools/composables/src/useCart/index.ts
Fifciu: So how should it been done in your opinion? I thought it would be really good solution. Wdyt about it @andrzejewsky ?

andrzejewsky: if you want to use the same logic in both geetter and here, just move it to the separated file / function and re-use in both places: one in getter, second here - because we can't use getters inside of composables, that breaks our api-design, getters are designed to be used in the themes, providing agnostic-way,

packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -22,29 +22,27 @@ export type UseCartFactoryParams<CART, CART_ITEM, PRODUCT, COUPON> = {
   applyCoupon: (params: {
     currentCart: CART;
     coupon: string;
-  }) => Promise<{ updatedCart: CART; updatedCoupon: COUPON }>;
+  }) => Promise<{ updatedCart: CART }>;
   removeCoupon: (params: {
```
filrak, line 26: why you removed `coupon` property from `removeCoupon`? How will we know which one we want to remove?

packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -124,11 +121,9 @@ export const useCartFactory = <CART, CART_ITEM, PRODUCT, COUPON>(
       try {
         loading.value = true;
         const { updatedCart } = await factoryParams.removeCoupon({
-          currentCart: cart.value,
-          coupon: appliedCoupon.value
+          currentCart: cart.value
```
filrak, line 124: how do you know which coupon you're removing?

filrak (changes requested):
You have inconsistency in `removeCoupon`

You want it to accept a param in public API but factory is not accepting any param

packages/core/core/src/factories/useCartFactory.ts
Fifciu: `getCouponFromCart` function will tell that.

packages/core/core/src/factories/useCartFactory.ts
Fifciu: `getCouponFromCart` function will tell that.

PR #4883, opened 2020-09-17
CLI Versions bump

Fifciu:
CLI Versions bump


andrzejewsky (approved):

Fifciu:
I also fixed `prepublish` command - now it also runs `tsc` before publish. Earlier, I probably forgot about building - which caused problems on some node versions

filrak (approved):

PR #4888, opened 2020-09-18
[CT] Fix updateUser method

### LINKED ISSUE #4863: updating user in the composable breaks the user value
## Current behavior
When updateUser is called its running factoryparams method and assigns the returned value to the user.value.
https://github.com/DivanteLtd/vue-storefront/blame/next/packages/commercetools/composables/src/useUser/factoryParams.ts#L32

But, instead of the updated user object it returns an object with these keys: 
```
{currentUser: ..., updatedUserData: ...}
```

Here: https://github.com/DivanteLtd/vue-storefront/blob/next/packages/core/core/src/factories/useUserFactory.ts#L35

## Expected behavior

It will do the update properly.

See description.

## Version of Vue Storefront

- [ ] Vue Storefront
- [x] Vue Storefront Next

- [ ] YES
- [x] NO

## Which [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html) state this refers to? Info for developer. (doesn't apply to Next)
Pick one option.

## Environment details

- Browser: 
- OS: 
- Node: 
- Code Version:

filipsobol:
Closes #4863


andrzejewsky (approved):
@filipsobol update changelog

packages/commercetools/api-client/src/helpers/customer/index.ts
```diff
@@ -0,0 +1,11 @@
+export const changeCustomerEmail = (email: string) => ({
```
filrak, line 1: for what we need those? Is this a convention in the api client? I don't think these functions bring any value by being externalized, they're one-liners

andrzejewsky: @filrak this is an internal helper, it's not exposed to the external world

andrzejewsky: @filrak they are action creator actually here is the other example: https://github.com/DivanteLtd/vue-storefront/blob/next/packages/commercetools/api-client/src/helpers/cart/actions.ts, @filipsobol please ad *Action ad the end.

filrak: i know its a private api but I'm askig about the reason behind externalziing a one line fn into a separate file?

filipsobol: IMO in this case it only overcomplicates things.

PR #4911, opened 2020-09-24
Logger implementation

andrzejewsky:
### Related Issues

closes https://github.com/DivanteLtd/vue-storefront/pull/4578

Logging system implementation which can be useful to track the errors by external providers or even debuging.

### Usage

If you are using our nuxt-module, the logger is being registered out of the box. You have also available config options that can override the default logger

```
    ['@vue-storefront/nuxt', {
      coreDevelopment: true,
      logger: { // new section here
        verbosity: 'error' // verbosity
        // anything else you want to override, debug, info, error, warn
      },
      useRawSource: {
        dev: [
          '@vue-storefront/commercetools',
          '@vue-storefront/core'
        ],
        prod: [
          '@vue-storefront/commercetools',
          '@vue-storefront/core'
        ]
      }
    }],
```

In case when you don't use our module, you have function `registerLogger` in the core, so adding a new logger implementation look like this:

```
registerLogger(logger: VSFLogger, verbosity: string)
```

```js
import { registerLogger } from '@vue-storefront/core'

registerLogger({
  debug: (message: any, ...args) => console.debug('[VSF][debug]', message, ...args),
  info: (message: any, ...args) => console.info('[VSF][info]', message, ...args),
  warn: (message: any, ...args) => console.warn('[VSF][warn]', message, ...args),
  error: (message: any, ...args) => console.error('[VSF][error]', message, ...args),
}, 'error');
```

## Custom logger

As we are handling verbosity by ourselves, sometimes there is a need to implement your own solution. To do that you can provide totally your own logger by passing a function instead:

```
const logger = (verbosity) => {
 // current verbosity available here
 
 return {
   debug: (message: any, ...args) => console.debug('[VSF][debug]', message, ...args),
   info: (message: any, ...args) => console.info('[VSF][info]', message, ...args),
   warn: (message: any, ...args) => console.warn('[VSF][warn]', message, ...args),
   error: (message: any, ...args) => console.error('[VSF][error]', message, ...args),
 }
}

registerLogger(logger, 'error');
```

When it comes to our nuxt module, there is special field to do that (`customLogger`):

```
    ['@vue-storefront/nuxt', {
      coreDevelopment: true,
      logger: { // new section here
        verbosity: 'error' // verbosity
        customLogger: logger // your own implementation
      },
      useRawSource: {
        dev: [
          '@vue-storefront/commercetools',
          '@vue-storefront/core'
        ],
        prod: [
          '@vue-storefront/commercetools',
          '@vue-storefront/core'
        ]
      }
    }],
```


packages/core/core/src/factories/apiClientFactory.ts
```diff
@@ -13,15 +14,21 @@ export function apiClientFactory<ALL_SETTINGS, CONFIGURABLE_SETTINGS>(factoryPar
       settings = merge(factoryParams.defaultSettings, config);
       factoryParams.onSetup(settings);
 
+      console.log('TEST', Logger);
```
filrak, line 17: 🗡️

packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -44,6 +45,8 @@ export const useCartFactory = <CART, CART_ITEM, PRODUCT, COUPON>(
     const cart: Ref<CART> = sharedRef(null, 'useCart-cart');
 
     const addToCart = async (product: PRODUCT, quantity: number, customQuery?: CustomQuery) => {
+      Logger.debug('userCart.addToCart', { product, quantity });
```
filrak, line 48: `useCart` not `userCart`

packages/commercetools/theme/nuxt.config.js
```diff
@@ -32,6 +32,9 @@ export default {
     '@nuxt/typescript-build',
     ['@vue-storefront/nuxt', {
       coreDevelopment: true,
+      logger: {
```
filrak, line 35: `displayErrorsOnProd` will be valuable here as well

packages/core/core/src/types.ts
```diff
@@ -443,3 +443,11 @@ export interface AgnosticFacetSearchParams {
   metadata?: any;
   [x: string]: any;
 }
+
+export interface AgnosticLogger {
```
filrak, line 447: `Logger` is enough imho or `VSFLogger` it's not agnostic type ;)

packages/core/core/src/types.ts
```diff
@@ -443,3 +443,11 @@ export interface AgnosticFacetSearchParams {
...
+  debug(message?: any, ...args: any): void;
+  info(message?: any, ...args: any): void;
+  warn(message?: any, ...args: any): void;
+  error(message?: any, ...args: any): void;
+  verbosity: string;
```
filrak, line 452: you mean `defaultVerbosity` ? should that be even in this interface? imho its more for `registerLogger`

packages/core/core/src/utils/logger/index.ts
```diff
@@ -0,0 +1,52 @@
+import { AgnosticLogger } from './../../types';
```
filrak, line 1: how about having default logger in `utils/logger/default` seems more readable

packages/commercetools/theme/nuxt.config.js
andrzejewsky: I've introduces just verbosity `none` which means _don't log anything_ displaying errors etc. is depending on the logger you provide, eg. you are using sentry - it's your implementation of what you want to display and what now. Verbosity level means _which functions will be called_ (debug, error, info, debug)

packages/core/core/src/factories/useUserFactory.ts
```diff
@@ -55,41 +64,57 @@ export const useUserFactory = <USER, UPDATE_USER_PARAMS, REGISTER_USER_PARAMS ex
...
         user.value = null;
       } catch (err) {
+        Logger.error('useUserFactory.err');
+
         throw new Error(err);
       }
     };
 
     const changePassword = async (currentPassword: string, newPassword: string) => {
+      Logger.debug('useUserFactory.changePassword', { currentPassword, newPassword });
```
samriley, line 95: should the passwords be logged in full? would it make sense to instead mask them ie show only first and last character. you can tell that the correct value is passed without accidentally logging people passwords

packages/core/core/src/factories/useUserShippingFactory.ts
```diff
@@ -40,20 +40,26 @@ export const useUserShippingFactory = <ADDRESS>(
     const readonlyDefaultAddress: Readonly<ADDRESS> = unref(defaultAddress);
 
     const addAddress = async (address: ADDRESS) => {
+      Logger.debug('useUserShipping.addAddress', address);
```
samriley, line 43: probably the same comment about paswords applies here - should pii data be masked to prevent accidental leaking

packages/commercetools/theme/nuxt.config.js
filrak: yeah, i also realized you can just do `varbosity: isProd ? "error" : "all"`

packages/core/nuxt-module/lib/module.js
```diff
@@ -43,6 +44,13 @@ module.exports = function VueStorefrontNuxtModule (moduleOptions) {
   this.addPlugin(path.resolve(__dirname, 'plugins/ssr.js'))
   log.success('Installed VSF SSR plugin');
 
+  this.addPlugin({
```
filrak, line 47: shouldn't we run this only if custom logger is provided?

filrak (changes requested):
missing docs

packages/core/nuxt-module/lib/module.js
andrzejewsky: nope, because you need to run logger configuration for example for verbosity etc. but.... I  think you mean to load this plugin only if section logger in the config is provided and in that case - yes, will do

filrak: yup, exactly. default logger should be applied on the library level ;)

packages/core/docs/general/logging.md
```diff
@@ -0,0 +1,102 @@
+# Logging
+
+A logging error, debugging information, and warnings are a very common case during development and future monitoring of the application. In the Vue Storefront core, you can catch everything you need to screen what is happening inside of your app.
```
filrak, line 3:
```suggestion
In Vue Storefront we're providing all the debugging information, warnings, and errors out of the box so you won't miss anything that happens inside your application. Sometimes you have to pass these data to some third-party monitoring services. You can do this through a custom Logger.
 ```

filrak (approved):

PR #4917, opened 2020-09-25
Creating user session token on-demand

andrzejewsky:
We used to create an anonymous token (user session) each time user visited the page which can end up with so many opened user sessions. 

This PR introduces creating that token only on-demand when we really need to open user session eg. when we interact with the cart or user profile

### DIFF
--- a/packages/commercetools/composables/src/useCart/index.ts
+++ b/packages/commercetools/composables/src/useCart/index.ts
@@ -15,6 +15,7 @@ const getBasketItemByProduct = ({ currentCart, product }) => {
   return currentCart.lineItems.find((item) => item.productId === product._id);
 };
 
+/** returns current cart or creates new one **/
 const getCurrentCart = async (currentCart) => {
   if (!currentCart) {
     return loadCurrentCart();
--- a/packages/commercetools/composables/src/useCart/index.ts
+++ b/packages/commercetools/composables/src/useCart/index.ts
@@ -13,31 +13,49 @@ const getBasketItemByProduct = ({ currentCart, product }) => {
   return currentCart.lineItems.find((item) => item.productId === product._id);
 };
 
+const getCurrentCart = async (currentCart) => {
+  if (!currentCart) {
+    return loadCurrentCart();
+  }
+
+  return currentCart;
+};
+
 const params: UseCartFactoryParams<Cart, LineItem, ProductVariant, any> = {
   loadCart: async (customQuery?: CustomQuery) => {
     return await loadCurrentCart(customQuery);
   },
   addToCart: async ({ currentCart, product, quantity }, customQuery?: CustomQuery) => {
-    const { data } = await apiAddToCart(currentCart, product, quantity, customQuery);
+    const loadedCart = await getCurrentCart(currentCart);
+
+    const { data } = await apiAddToCart(loadedCart, product, quantity, customQuery);
     return data.cart;
   },
   removeFromCart: async ({ currentCart, product }, customQuery?: CustomQuery) => {
-    const { data } = await apiRemoveFromCart(currentCart, product, customQuery);
+    const loadedCart = await getCurrentCart(currentCart);
+
+    const { data } = await apiRemoveFromCart(loadedCart, product, customQuery);
     return data.cart;
   },
   updateQuantity: async ({ currentCart, product, quantity }, customQuery?: CustomQuery) => {
-    const { data } = await apiUpdateCartQuantity(currentCart, { ...product, quantity }, customQuery);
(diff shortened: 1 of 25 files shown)


packages/commercetools/composables/src/useCart/index.ts
```diff
@@ -13,31 +15,55 @@ const getBasketItemByProduct = ({ currentCart, product }) => {
   return currentCart.lineItems.find((item) => item.productId === product._id);
 };
```
filrak, line 17: For mouse-over explanation
```suggestion
/** returns current cart or creates new one **/
```

packages/core/nuxt-theme-module/theme/components/AppHeader.vue
```diff
@@ -65,21 +65,32 @@ export default {
 
     const accountIcon = computed(() => isAuthenticated.value ? 'profile_fill' : 'profile');
 
-    const onAccountClicked = () => {
-      isAuthenticated && isAuthenticated.value ? root.$router.push('/my-account') : toggleLoginModal();
+    const handleSidebarOpen = async () => {
+      await loadCart();
```
filrak, line 69: IMHO this should work differently. We cant have logic that is required for sidebar to work outside of the sidebar.
It could be either:
- watcher in the sidebar that loads the cart when `isOpen` becomes true
- fn in `onMounted` hook and sidebar will become async and lazily loaded

packages/core/nuxt-theme-module/theme/components/AppHeader.vue
```diff
@@ -65,21 +65,32 @@ export default {
...
+      toggleCartSidebar();
+    };
+
+    const handleAccountClick = async () => {
+      await loadUser();
```
filrak, line 74: same here, you can't assume that its the only place account is entered from, this should work on every occasion when we navigate to user profile. 

Both auth check and loading user should be handled in my account page

packages/core/nuxt-theme-module/theme/components/AppHeader.vue
```diff
@@ -65,21 +65,32 @@ export default {
...
+
+      if (isAuthenticated.value) {
+        return root.$router.push('/my-account');
+      }
+
+      toggleLoginModal();
     };
 
     onSSR(async () => {
+      await loadUser();
```
filrak, line 84: why you do this here and then in `handleAccountClick` (same for cart)

packages/core/nuxt-theme-module/theme/pages/Category.vue
```diff
@@ -274,9 +274,13 @@ export default {
 
     onSSR(async () => {
       await search(th.getFacetsFromURL());
-      await loadCart();
     });
 
+    const handleAddToCart = async (product, qty) => {
```
filrak: already commented on that

filrak (changes requested):

packages/core/core/src/types.ts
```diff
@@ -54,7 +54,7 @@ export interface UseUser
   changePassword: (
     currentPassword: string,
     newPassword: string) => Promise<void>;
-  refreshUser: () => Promise<void>;
+  loadUser: () => Promise<void>;
```
filrak, line 57:
```suggestion
  load: () => Promise<void>;
```

packages/core/nuxt-theme-module/theme/components/AppHeader.vue
```diff
@@ -65,20 +65,24 @@ export default {
 
     const accountIcon = computed(() => isAuthenticated.value ? 'profile_fill' : 'profile');
 
-    const onAccountClicked = () => {
-      isAuthenticated && isAuthenticated.value ? root.$router.push('/my-account') : toggleLoginModal();
+    const handleAccountClick = async () => {
+      if (isAuthenticated.value) {
```
filrak, line 69: should be done on the user page (redirect back and show modal)

filrak (changes requested):
would e nice to mention about how it works in docs as well

andrzejewsky:
Ref: https://github.com/DivanteLtd/vue-storefront/issues/4927

filrak (approved):

packages/core/docs/contributing/changelog.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 ## 2.1.0 (not released)
 
+- renamed refreshUser to load in useUser (#4917)
```
filrak, line 4:
```suggestion
- renamed `refreshUser` to `load` in `useUser`, user shouldn't be automatically loaded now (#4917)
```

Issue #4903, opened 2020-09-23
Maintenance mode


pablogiralt:
## What is the motivation for adding / enhancing this feature?

In causes where downtime on a VSF site is needed for maintenance tasks, there is currently no way to block the access to the front-end for users.
It has happened to me that when doing necessary maintenance in magento backend and setting it to maintenance mode, the front-end obviously is still accesible by users but it does not work (cart, checkout, etc...).  This is a bad experience for users accessing the site at that time.

## What are the acceptance criteria 

* [ ] Add a setting to config to enable/disable maintenance mode
* [ ] Users are redirected to a maintenance page when maintenance mode is enabled.
* [ ] Maintenance mode should work on any app route and if routes are loaded client-side or server-side

## Version of Vue Storefront

* [x] Vue Storefront
* [x] Vue Storefront Next
(could be useful for both)

## Can you complete this feature request by yourself?
* [ ] YES
* [x] NO

I could handle layout and styling for the maintenance page, but not the logic

## Additional information

How to handle this in Nuxt: https://medium.com/code4mk-org/nuxt-maintenance-mode-48e158c429e2

bloodf:
This issue has been closed due to inactivity. 

If you want to re-open the issue, please re-create the issue, and post newer information on this problem.

PR #4937, opened 2020-09-29
handle redirection on ssr 4647

### LINKED ISSUE #4647: handle redirection on client and on server

Right now if we want to redirect user we need to check if he is on client or server side. In first case we use router and in second we need to get access to server context and then we can make redirection in express app.
For example;
```
      const newUrl = '/women/tops-women/tees-women/tees-25/karissa-v-neck-tee-1529.html?childSku=WS10-XS-Green'
      AsyncDataLoader.push({
        execute: async ({ context }) => {
          if (context && !context.url.includes(newUrl)) {
            context.server.response.redirect(301, newUrl)
          }
        }
      })
```
or same thing in `asyncData`.
I think this is a common problem which should be resolved by service or helper in core.

Also right now when we make redirection on server we can get error `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client`. That is because in server.ts we still call res.send().

Last thing is that we repeat data fetch in each request. I don't know if this can be solved.
For example:
Lets say that user goes to product page then he needs to fetch product, menu, breadcrumbs (if menu is lazy loaded), attributes...
Now at the end of that we have enough data to decide if he needs to be redirected to other page. After redirection it goes once again => fetch product, menu...

If we have redis then maybe we can add vuex data from redirected request to redis. After redirection we can check if there is data in redis (and remove it, this data should exist only for next request) and we can add it as initial vuex state. In that case we don't need to make requests again.

- [ ] add helper or service which will handle redirection on ssr and csr
- [ ] handle redirection in server.ts after `renderer.renderToString(context)` is finished
- [ ] (optional) keep vuex state between requests

- [x] YES
- [ ] NO

Pick one option.

gibkigonzo:
### Related Issues

closes #4647
related #4903 - we can use redirect here

 [x] - add helper or service which will handle redirection on ssr and **csr** (I didn't handle it on client side, because I think there is no need to. It's easy to be done and there are no problems with it. We can use router.push/router.replace in any place)
 [x] - handle redirection in server.ts after renderer.renderToString(context) is finished
 [ ] - (optional) keep vuex state between requests - TODO

### DIFF
--- a/core/scripts/server.ts
+++ b/core/scripts/server.ts
@@ -194,9 +194,9 @@ app.get('*', (req, res, next) => {
     }
     const context = ssr.initSSRRequestContext(app, req, res, config)
     renderer.renderToString(context).then(output => {
-      if (context.redirect.isPending()) {
-        console.log(`Redirection from ${context.url}`)
-        context.redirect.resolve()
+      if (context.server._redirect.isPending()) {
+        console.log(`redirect from [${context.url}]`)
+        context.server._redirect.resolve()
         return
       }
       if (!res.get('content-type')) {
--- a/core/scripts/server.ts
+++ b/core/scripts/server.ts
@@ -198,7 +198,7 @@ app.get('*', (req, res, next) => {
       if (context.server._redirect.isPending()) {
         console.log(`redirect from [${context.url}]`)
         // it should have arguments setup, we just need to call it
-        context.server._redirect.resolve.call(null)
+        context.server._redirect.resolver.call(null)
         return
       }
       if (!res.get('content-type')) {
--- a/core/scripts/server.ts
+++ b/core/scripts/server.ts
@@ -194,6 +194,11 @@ app.get('*', (req, res, next) => {
     }
     const context = ssr.initSSRRequestContext(app, req, res, config)
     renderer.renderToString(context).then(output => {
+      if (context.redirect.isPending()) {
+        console.log(`Redirection from ${context.url}`)
+        context.redirect.resolve()
+        return
+      }
       if (!res.get('content-type')) {
         res.setHeader('Content-Type', 'text/html')
       }
--- a/core/scripts/server.ts
+++ b/core/scripts/server.ts
@@ -1,4 +1,5 @@
 import { serverHooksExecutors } from '@vue-storefront/core/server/hooks'
+import { Context } from './utils/types';
 
 let config = require('config')
 const path = require('path')
@@ -192,11 +193,12 @@ app.get('*', (req, res, next) => {
       res.status(202).end(HTMLContent)
       return next()
     }
-    const context = ssr.initSSRRequestContext(app, req, res, config)
+    const context: Context = ssr.initSSRRequestContext(app, req, res, config)
     renderer.renderToString(context).then(output => {
       if (context.server._redirect.isPending()) {
         console.log(`redirect from [${context.url}]`)
-        context.server._redirect.resolve()
+        // it should have arguments setup, we just need to call it
+        context.server._redirect.resolve.call(null)
         return
       }
       if (!res.get('content-type')) {
--- a/core/scripts/utils/ssr-redirect.ts
+++ b/core/scripts/utils/ssr-redirect.ts
@@ -1,9 +1,6 @@
 import { Response } from 'express'
 import { Context } from './types'
 
-// Redirection resolver type
-export type SsrRedirectResolve = (code: number, path: string) => void
-
 function createIsPending (context: Context): () => boolean {
   return () => !!context.server._redirect.pendingPath
 }
@@ -13,6 +10,8 @@ function createIsPending (context: Context): () => boolean {
  */
 function createResolveMethod (context: Context, expressResponse: Response) {
   return (code: number, path: string) => {
+    if (!context.server._redirect.isPending()) return
+
     // normal express redirection
     expressResponse.redirect(code, path)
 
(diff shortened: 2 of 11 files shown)


core/scripts/server.ts
```diff
@@ -194,6 +194,11 @@ app.get('*', (req, res, next) => {
     }
     const context = ssr.initSSRRequestContext(app, req, res, config)
     renderer.renderToString(context).then(output => {
+      if (context.server._redirect.isPending()) {
```
Fifciu, line 197: It looks good. However, what would happen if we want to redirect the user from the cached page? I think we should also support this case

core/scripts/utils/ssr-redirect.ts
```diff
@@ -0,0 +1,51 @@
...
+  // lock path
+  if (typeof path !== 'string') throw new Error(`Redirection path needs to be a string`)
+  this.server._redirect.pendingPath = path
+
+  this.server._redirect.resolve = resolve.bind(this, path, options || {})
+}
+
+export function createContextRedirection (): SsrRedirect {
+  this.server._redirect = {
+    isPending: isPending.bind(this)
```
Fifciu, line 47: In this moment `this.server._redirect` still equals `null` right? If so wouldn't `return !!this.server._redirect.pendingPath` throw `Error: pendingPath of null`?

core/scripts/utils/ssr-redirect.ts
```diff
@@ -0,0 +1,51 @@
...
+/*
+ * Resolve redirection only once per request.
+ */
+function resolve (path: string, {
+  code = 302
+}: RedirectOptions) {
+  // normal express redirection
+  this.server.response.redirect(code, path)
+  // clear temp object
+  Object.keys(this.server._redirect).forEach((key) => delete this.server._redirect[key])
```
Fifciu, line 21: Couldn't we just use?:
```js
this.server._redirect = {};
```

Fifciu (changes requested):
Please check my suggestions/questions. I also recommend to write:
- Simple doc - how to redirect with asyncData/AsyncDataLoader
- Unit tests for `ssr-redirect.ts`

core/scripts/server.ts
gibkigonzo: In that case you need to clear cache. `dynamicRequestHandler` will not be resolved for cached pages. So if you changed urls, you need to clear cache and apply logic in asyncData that will handle redirection. I don't see other way :thinking:

core/scripts/utils/ssr-redirect.ts
gibkigonzo: `this.server._redirect.pendingPath` it's not possible to call isPending before context is created - but it's outdated

core/scripts/utils/ssr-redirect.ts
gibkigonzo: I used assign to null - outdated

gibkigonzo:
**update** - added proxy for express reponse object to match api
Todo: tests + docs

core/scripts/server.ts
Fifciu: Couldn't we apply a similar mechanism to dynamicCacheHandler?

gibkigonzo:
`[ ] - (optional) keep vuex state between requests ` - we can't initialize state with values, because after store initialization modules override state. So it can't be done

core/scripts/utils/ssr-redirect.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!context.server._redirect) {
+    console.warn(`Something went wrong while creating vsf redirection handler. Fallback to normal express redirect`)
+    return expressResponse
+  }
+
+  const ProxyConstructor = Proxy || require('proxy-polyfill/src/proxy')
+
+  // returns proxy for Express Reponse object
+  return new ProxyConstructor(expressResponse, {
+    get (target, propKey) {
```
Fifciu, line 87: Couldn't we do it like that:
```js
get (target, propKey) {
      // 'redirect' can be called multiple times and we want to reuse same method
      if (propKey === 'redirect') {
        return context.server._redirect.handler
      }

      const originalMethod = target[propKey]
      return originalMethod;
    }
```

core/scripts/utils/types/index.ts
```diff
@@ -12,12 +12,45 @@ export interface Context {
...
+   * it will play nicely with mounted apps, relative paths,
+   * `"back"` etc.
+   *
+   * Examples:
+   *
+   *    res.redirect('/foo/bar');
+   *    res.redirect('http://example.com');
+   *    res.redirect(301, 'http://example.com');
+   *    res.redirect('http://example.com', 301);
+   *    res.redirect('../login'); // /blog/post/1 -> /blog/login
```
Fifciu, line 51: Just wanted to say, it is so cool case. Glad it is supported!

Fifciu (changes requested):

core/scripts/utils/ssr-redirect.ts
```diff
@@ -85,13 +85,12 @@ export function createRedirectProxy (context: Context, expressResponse: Response
   // returns proxy for Express Reponse object
```
Fifciu, line 85: Reponse in comment! Fix it if you want, PR looks great right now.

Fifciu (approved):

core/scripts/utils/ssr-redirect.ts
gibkigonzo: done

PR #4950, opened 2020-10-01
Update faceting mechanism

andrzejewsky:
Extending the faceting mechanism:

- add search phase support
- add counting support
- add root category support


packages/core/nuxt-theme-module/theme/composables/useUiHelpers/index.ts
```diff
@@ -11,31 +11,36 @@ const useUiHelpers = () => {
...
 
   // eslint-disable-next-line
   const switchItemsPerPage = (itemsPerPage) => {
-    console.warn('[VSF] please implement createThemeHelpers.switchItemsPerPage.');
+    console.warn('[VSF] please implement useUiHelpers.switchItemsPerPage.');
+  };
+
+  // eslint-disable-next-line
+  const switchSearchPhase = (phase: string) => {
+    console.warn('[VSF] please implement useUiHelpers.switchItemsPerPage.');
```
filrak, line 38: image
search phase or items per page? ;>

packages/core/nuxt-theme-module/theme/composables/useUiHelpers/index.ts
```diff
@@ -11,31 +11,36 @@ const useUiHelpers = () => {
...
   };
 
   // eslint-disable-next-line
   const switchFilters = (filters) => {
-    console.warn('[VSF] please implement createThemeHelpers.switchFilters.');
+    console.warn('[VSF] please implement useUiHelpers.switchFilters.');
   };
 
   // eslint-disable-next-line
   const switchItemsPerPage = (itemsPerPage) => {
```
filrak, line 32: isn't that more like a `change` instead of switch? talking purely about naming (and doesn't it apply to all fns here?

filrak (changes requested):

packages/core/nuxt-theme-module/theme/composables/useUiHelpers/index.ts
andrzejewsky: I used word switch because it's related to the URL but can be `change` too

packages/core/core/src/types.ts
```diff
@@ -434,6 +435,8 @@ export interface FacetSearchResult<S> {
 
 export interface AgnosticFacetSearchParams {
   categorySlug?: string;
+  rootCatSlug?: string;
+  phase?: string;
```
filrak, line 439: term

filrak (approved):

PR #4959, opened 2020-10-02
Make isUserSession configurable

andrzejewsky:
Usage:

```
setup({
 handleIsTokenUserSession: (token) => {
  // ....
  return true / false
 }
})
```

### DIFF
--- a/packages/core/docs/commercetools/api-client.md
+++ b/packages/core/docs/commercetools/api-client.md
@@ -31,7 +31,7 @@ setup({
 ```
 **`setup`** accepts following properties:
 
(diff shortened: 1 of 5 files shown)


filrak (changes requested):
DOCS

packages/core/docs/commercetools/api-client.md
```diff
@@ -117,6 +117,25 @@ const serverMiddleware = async ({ app }) => {
 export default serverMiddleware;
 ```
 
+#### Checking the user session
```
filrak, line 120: add this to this big interface at the top

filrak (changes requested):

PR #4962, opened 2020-10-05
Block guest access to user profile

filipsobol:
This PR blocks acess to user profile pages for guests.

- [X] RC version (https://next.storefrontcloud.io) - this is a stabilisation fix for Release Candidate of Vue Storefront. I've created branch from `release` branch and want to merge it back to `release`

- [ ] I've updated the [Upgrade notes](https://github.com/DivanteLtd/vue-storefront/blob/develop/docs/guide/upgrade-notes/README.md) and [Changelog](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md) on how to port existing Vue Storefront sites with this new feature

### DIFF
--- a/packages/core/nuxt-theme-module/theme/pages/MyAccount.vue
+++ b/packages/core/nuxt-theme-module/theme/pages/MyAccount.vue
@@ -54,7 +54,6 @@ import MyNewsletter from './MyAccount/MyNewsletter';
 import OrderHistory from './MyAccount/OrderHistory';
 import MyReviews from './MyAccount/MyReviews';
 
-// TODO: protect this route: https://github.com/DivanteLtd/next/issues/379
 export default {
   name: 'MyAccount',
   components: {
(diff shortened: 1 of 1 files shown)


filrak (approved):

packages/core/nuxt-theme-module/theme/pages/MyAccount.vue
```diff
@@ -68,6 +67,17 @@ export default {
     OrderHistory,
     MyReviews
   },
+
```
andrzejewsky, line 70: what about this:
https://github.com/DivanteLtd/vue-storefront/blob/next/packages/commercetools/theme/middleware/checkout.js

I think you can do something similar and we can have just one middleware in the CT theme, it needs a bit refactoring, but both require user so:

- rename checkout.js to the eg. access-guard.js
- move checkout logic to the separate file or just function above
- implement the logic for my account section in another function
- use both in the middleware

in this case, you will benefit from the user call which actually is being called even now

andrzejewsky (changes requested):

packages/core/nuxt-theme-module/theme/pages/MyAccount.vue
filipsobol: Currently this middleware is part of the core, meaning that it works in all implementations. If we move it to CT, other integrations will lose it, unless we repeat it for every one of them.

andrzejewsky (approved):

PR #5028, opened 2020-10-15
Remove checkout shipping defaulting

andrzejewsky:
### Related Issues
closes https://github.com/DivanteLtd/vue-storefront/issues/5026


packages/commercetools/api-client/src/api/deleteCart/defaultMutation.ts
```diff
@@ -0,0 +1,12 @@
+import gql from 'graphql-tag';
+import { CartFragment } from '../../fragments';
+
+export default gql`
+  ${CartFragment}
+
+  mutation deleteCart($id: String!, $version: Long!, $locale: Locale!, $acceptLanguage: [Locale!]) {
```
EugeneHlushko, line 7: I think the cart here will get its status set to `ordered` and wont be queried as an `active cart` anymore via getMe endpoint

packages/commercetools/composables/src/useCheckout/createPlaceOrder.ts
```diff
@@ -9,12 +9,13 @@ const createPlaceOrder = ({ factoryParams, cartFields, setCart }) => async () =>
   const { id, version } = cartFields.cart.value;
 
   const orderResponse = await createMyOrderFromCart({ id, version });
-  const cartResponse = await createCart();
-  setCart(cartResponse.data.cart);
-  initFields(cartResponse.data.cart);
+  const { order } = orderResponse.data;
+
+  const cartResponse = await deleteCart(order.cart.id, order.cart.version);
+  setCart(null);
```
EugeneHlushko, line 15: When placing an order from lets say order review page, you will call `await placeOrder()` after which you would want to redirect someone to lets say success page, but, when that is done, your current page will throw errors in different components saying it cannot read its data because cart is `null` and before the redirect your page will flash because it will become empty for a moment because the cart is `null`

Fifciu (approved):
I see nothing more than @EugeneHlushko said

EugeneHlushko (approved):

filrak (approved):

PR #5080, opened 2020-10-27
Added support to configure custom auth header

### LINKED ISSUE #5078: Add support for custom auth header
We added support to pass token in header instead of query param at [PR](https://github.com/DivanteLtd/vue-storefront/pull/4626 )

In this PR we used Authorization header along with  Bearer scheme, but this will not work if server support other than Authorization header for token authentication. 

Proposal to add tokenHeaderName in config.user property, default value will be "Authorization", if tokenInHeader set to true then getPayload method will use the tokenHeaderName else not.

dixitdeepak:
### Related Issues

closes #5078

Added support to configure custom auth header. User can define the auth header name in config file, this will work with tokenInHeader configuration. If tokenInHeader set to true, getPayload will use the tokenHeaderName and send token along with it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for boolean filter aggregations in ES7 - @cewald (#4887)
 - Remove vue-lazyload from core - @jahvi (#5045)
 - Remove unnecessary async and Logger import - @jahvi (#5039)
+- Added support to configure custom auth header name - dixitdeepak (#5078)
 
 ## [1.12.2] - 2020.07.28
 
--- a/config/default.json
+++ b/config/default.json
@@ -752,7 +752,8 @@
     "me_endpoint": "/api/user/me?token={{token}}",
     "refresh_endpoint": "/api/user/refresh",
     "allowModification": ["firstname", "lastname", "email", "addresses"],
-    "tokenInHeader": false
+    "tokenInHeader": false,
+    "tokenHeaderName": "Authorization"
   },
   "stock": {
     "synchronize": true,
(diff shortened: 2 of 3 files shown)


Fifciu (changes requested):
Amazing job! It looks great. However, I would like to propose another functionality. We should make adding prefix like "Bearer" optional. E.g. let's add new option in config which contains `prefix`. If it is empty - do not add any prefix. Otherwise, add it before token. The default value should be probably just a `Bearer`.

dixitdeepak:
Thanks for review @Fifciu, Let me update code

dixitdeepak:
I'd like to use tokenAuthenticationScheme instead of prefix
https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Authentication_schemes

Fifciu:
Sound reasonable. But there might be some custom shapes like `AWS4-HMAC-SHA256` which is not a part of the table here (https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml). I would like to support more than less shapes. 

How do you see the implementation?

dixitdeepak:
Updated code, used tokenAuthScheme in default.json config file, default value will be `Bearer`

Fifciu (approved):
Ok, looks cool thanks

dixitdeepak:
> Sound reasonable. But there might be some custom shapes like `AWS4-HMAC-SHA256` which is not a part of the table here (https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml). I would like to support more than less shapes.
> 
> How do you see the implementation?

Frankly speaking, I haven't got a chance to work with custom shapes.
General pattern for Authorization header is like

`Authorization: <type> <credentials>`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

So I think we can use any type with the current implementation, but if we want to prepare customize token then we need to do customization in token preparation work .

gibkigonzo (approved):

PR #5161, opened 2020-11-10
Cache mechanism

andrzejewsky:
ref: https://github.com/DivanteLtd/vue-storefront/issues/5139


packages/commercetools/theme/exampleCache.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable */
+
+const CacheDriver = (options, invalidators) => {
+  console.log('driver called', options);
+
+  return {
+    invoke: async ({ getTags, render }) => {
+
+      const res = await render();
```
pkarw, line 9: What's the format returned from this function? I mean the cache driver should have full access to the rendered body + headers in order to store it - then when it's a hit to set the headers from cache. Some drivers (like AKAMAI) won't store anything but just operate on the headers (setting proper cache headers)

packages/core/nuxt-module/lib/cache/module.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  createRenderer(async (route, context, render) => {
+    const driver = cache.getCacheDriver()
+    const getTags = () => {
+      if (context.req.vsfCache) {
+        return Array.from(context.req.vsfCache.tags)
+      }
+
+      if (context.req.headers['cache-tags']) {
```
pkarw, line 63: I suppose that the headers management should be 100% done in the cache driver - bc. each CDN/proxy requires different naming strategy and convention

pkarw (changes requested):
Maybe this review is not 100% apropriate but I don't see all required elements:
- driver should be able to control the output headers and has access to input headers (set by the CDN/Proxy on intercepting the request prior to process it in the Nuxt)
- don't see where actually the driver is checking if the page is hit or miss (for drivers storing the content in Redis, files or other storage) - drivers that gonna work with CDNs will always return MISS (bc. cache storage is subject to CDN responsibility) - but they for example gonna map the request URL / tags and store it in redist for the further need of purging cache by URLS not tags (for example when using nginx as a caching proxy)

packages/commercetools/theme/exampleCache.js
andrzejewsky: @pkarw invoke has the following signature:
```
async ({  route, context, getTags, render })
```

inside of the context you have access to `req` and `res` which means you have full access to HTTP headers, so you can modyfi them eg. by adding your own headers or even manipulate the response (depending on the needs)

andrzejewsky: when it comes to `render` it mostly returns rendered markup

packages/core/nuxt-module/lib/cache/module.js
andrzejewsky: actually, that's the reading cache tags from the request, you may right.. maybe we should add it to the driver too and maybe call `getCacheTagsFromRequest` or something

packages/commercetools/theme/exampleCache.js
```diff
@@ -0,0 +1,20 @@
...
+      console.log('invoke called 2', getTags());
+
+      return res;
+    },
+    invalidate: ({ req, res }) => {
+      console.log('invalidate called', req);
```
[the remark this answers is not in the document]
andrzejewsky: yes you have, look at the argument on the top

packages/commercetools/theme/nuxt.config.js
```diff
@@ -39,6 +39,13 @@ export default {
     '@nuxt/typescript-build',
     ['@vue-storefront/nuxt', {
       coreDevelopment: true,
+      cache: {
+        server: {
+          invalidateEndpoint: '/cache-invalidate',
+          invalidators: ['./invalidator'],
+          driver: ['./exampleCache', { test: 1 }]
```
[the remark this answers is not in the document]
andrzejewsky: sure, it's just an example and PR draft will remove all of the console logs, examples, drivers, configs etc.

andrzejewsky:
NOTES:
- fallback for rendering
- multiple drivers
- remove `context.req.headers['cache-tags']`

packages/commercetools/theme/exampleCache.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable */
+
+const CacheDriver = (options, invalidators) => {
+  console.log('driver called', options);
```
filrak, line 4: !

PR #5178, opened 2020-11-12
Context implementation

andrzejewsky:
### Summary:
* provided `useContext` abstraction - by default our nuxt module will add `useContext` from nuxt library, but you can provide your own implementation
* each factory method has context as the first argument so you don't need to import `getSettings` anymore
* each API client function is stateless - it requires a context too wit the settings as you can't use `getSettings` (as singleton) due to vue instance hydration reasons
* setup function returns an object of the configuration
* apiClientFactory doesn't store anything in the memory

## Usage in plugins
Settings will be available under the `vsfSettings` key in context. To provide this, please use the following example:

```js
const myPlugin = (app, inject) => {
 inject('vsfSettinfgs', setup(...))
}
```

we do the same thing for comemrcetools

## Usage in factory params
The first argument becomes always a context with available key `$vsfSettings`.

```js
const factoryParams = {
 search (context, params) {
    // context.$vsfSettings - current settings of the app
    return getProduct(context, params)
  }
}
```

## Working with api client
API client functions are stateless, you have to provide a context each time you use some api function all

```js
const $vsfSettings = setup({ configuration here }) // setup returns a configuration
const context = { $vsfSettings } // creating a context

const products = getProducts(context, { id: 'aabbcc' });
const categories = getCategory(context, { id: 'ccddee' });
```

## getSettings form API client
You can still use `getSettings `from the API client, it uses `useContext` underneath, but please keep in mind that you have to be inside of the setup() method with the existing vue instance. As an example, it won't work with SSR if you have many awaiters

## useContext
You can also access current context by using `useContext` from core

```js
import { useContext } from '@vue-storefront/core'

const context = useContext();

context.$vsfSettings // settings here
```

## Provide context implementation
You are able to provide your own context implementation. 

```js
import { configureContext } from '@vue-storefront/core'
import { useContext } from '@nuxtjs/composition-api';

const contextPlugin = () => {
  configureContext({ useContext });
};

export default contextPlugin;
```

By using our nuxt modules, `useContext` becomes an implementation that nuxt.js provides out of the box.

┆Issue is synchronized with this [Jira Story](https://vsf.atlassian.net/browse/NEXT-379) by [Unito](https://www.unito.io/learn-more)


andrzejewsky:
```js
const { setup, apiClientMethodFactory } = apiClientFactory('ct')

// 1. example api client function
function getProduct (params) {
  return apiCall(this.ct.url, params)
}

// apiClientMethodFactory():
//    import { config } from './index'
//    getProduct.raw = getProduct (*) (second bind can override previous one?)
//    export default getProduct.bind(config)
export default apiClientMethodFactory(getProduct)


// 2. Core
const testFactory = (factoryParams) => {

  const useTest = () => {
    // injectContext: 
    const context = useContext(); // 1
    factoryParams.settings = context.$vsfSettings; // ct
    factoryParams.api.getProduct.bind(context.$vsfSettings)
    const newFactoryParams = injectSettings(factoryParams)
  
    const search = async (params) => {
      return newFactoryParams.search(params);
    }
  
    return {
      search
    }
  }

  return useTest;
}



// 3. factory params
const factoryParams = {
  search (params) {
    // this.settings
    // this.api.getProduct
    return this.api.getProduct(params)
  },
  api: {
    getProduct
  }
}

const useTest = testFactory(factoryParams)


// 4. nuxt configuration (plugin)

const configPlugin = (_, inject) => {
  configureIntegration({}, inject)
  const { config, tag } = setup()
  // inject('vsf', setup({ ... }))
  // vsf.ct
}

// 5. using clean api client

setup({}) // store in global object 

getProduct(params)
```

PR #5197, opened 2020-11-17
Context implementation

andrzejewsky:
- `getSettings` works as previously but it reads configuration from the context
- to register an integration you have to use `registerIntegration`

```js
import { registerIntegration } from '@vue-storefront/core';
import { setup } from '@vue-storefront/some-integration';

export default registerIntegration(({ app, vsf }) => {
   // it's a regular nuxt plugin 
   vsf.configure(setup({ config here ))
})
```
of course `@vue-storefront/comercetools/nuxt` does the job for you
- calling `apiClientFactory` requires and extra param called `tag` - it's a string identifier of the configuration in the context (we can have many configurations in the context, so we need a tag them)

### DIFF
--- a/packages/core/core/src/factories/apiClientFactory.ts
+++ b/packages/core/core/src/factories/apiClientFactory.ts
@@ -2,17 +2,39 @@ import merge from 'lodash-es/merge';
 import { Logger } from './../utils';
 
 interface FactoryParams<T> {
+  tag: string;
   defaultSettings: any;
-  onSetup: (config: T) => void;
+  onSetup: (config: T) => T;
 }
 
+const isNuxt = () => {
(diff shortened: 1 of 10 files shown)


packages/core/core/src/factories/apiClientFactory.ts
```diff
@@ -2,17 +2,39 @@ import merge from 'lodash-es/merge';
 import { Logger } from './../utils';
 
 interface FactoryParams<T> {
+  tag: string;
   defaultSettings: any;
-  onSetup: (config: T) => void;
+  onSetup: (config: T) => T;
 }
 
+const isNuxt = () => {
```
filrak, line 10: this can be moved somewhere else imho

PR #5218, opened 2020-11-19
Context implementation

andrzejewsky:


packages/commercetools/composables/src/useCart/currentCart.ts
```diff
@@ -1,14 +1,14 @@
-import { createCart, getMe } from '@vue-storefront/commercetools-api';
+import { Context } from '@vue-storefront/core';
 
-const loadCurrentCart = async (customQueryFn = (user = null, cart = null) => ({ cart, user })) => {
+const loadCurrentCart = async (context: Context, customQueryFn = (user = null, cart = null) => ({ cart, user })) => {
```
filrak, line 3: we shouldn't have access to full `context` here, just `api`. Otherwise the same code will have different `context` on different enviroments and this could easily lead to issues

Not to mentions integrators who will most likely try to make use of nuxt ctx here

packages/commercetools/api-client/src/api/customerSignMeUp/index.ts
```diff
@@ -1,19 +1,18 @@
 import { CustomerSignMeUpDraft } from '../../types/GraphQL';
-import { getSettings } from '../../index';
 import CustomerSignMeUpMutation from './defaultMutation';
 import { SignInResponse } from '../../types/Api';
 import createAccessToken from '../../helpers/createAccessToken';
 
-const customerSignMeUp = async (draft: CustomerSignMeUpDraft): Promise<SignInResponse> => {
-  const { locale, acceptLanguage, currentToken, auth, client } = getSettings();
-  const registerResponse = await client.mutate({
+const customerSignMeUp = async (context, draft: CustomerSignMeUpDraft): Promise<SignInResponse> => {
```
filrak, line 6: why its `context` on some and `settings` on others?

filrak (commented):
We definitely need a type for IntegrationContext (then maybe we can call it `context` instead of `settings` because the type will differentiate it from nuxt context?

packages/commercetools/composables/nuxt/helpers/index.js
```diff
@@ -1,4 +1,6 @@
 import defaultConfig from '@vue-storefront/commercetools/nuxt/defaultConfig';
+import { createIntegrationPlugin } from '@vue-storefront/core';
```
filrak, line 2: 'core/nuxt` or `@vue-storefront/nuxt` ?

andrzejewsky: didn't work :/ but i need to work on this

PR #5257, opened 2020-11-27
Context standards

andrzejewsky:
### Related Issues
The standards of context API, naming, extending etc.

closes https://github.com/DivanteLtd/vue-storefront/issues/5249


packages/core/core/src/factories/apiClientFactory.ts
```diff
@@ -20,26 +19,22 @@ export interface BaseConfig {
...
-        tag: factoryParams.tag
-      };
-    }
+
+  const createApiClient = (config: ALL_SETTINGS, customApi: any = {}): ApiClientInstance => {
+    const settings = factoryParams.onSetup ? merge(config, factoryParams.onSetup(config)) as ALL_SETTINGS : { config, client: config.client };
+
+    Logger.debug('apiClientFactory.setup', settings);
+
+    const api = composeApiWithContext({ ...factoryParams.api, ...customApi }, settings);
```
filrak, line 28:
```suggestion
   // add context as 1st param
    const api = composeApiWithContext({ ...factoryParams.api, ...customApi }, settings);
```

packages/core/core/src/utils/nuxt/index.ts
```diff
@@ -1,18 +1,61 @@
-export const injectInContext = (inject, { tag, props }) => {
-  const key = 'vsf' + tag.toUpperCase();
-  inject(key, props);
+import { composeApiWithContext } from './../context';
+
+const createInjector = ({ tag, nuxtCtx, inject }) => {
```
filrak, line 3: how about `injecIntegrationToNuxtContext`, long but descriptive ;)

packages/core/core/src/utils/context/index.ts
```diff
@@ -1,29 +1,37 @@
...
+  return context.$vsf;
 };
 
 export {
   generateContext,
-  useContext,
-  configureContext
+  useVSFContext,
+  configureContext,
+  composeApiWithContext
```
filrak, line 36: how about `bind` instead of `compose`

filrak (approved):

PR #5043, opened 2020-10-20
refactor multistore

gibkigonzo:
### Related Issues

closes #4692 #3863 #3844 #3614

### Short Description and Why It's Useful

Added unit tests and e2e for multistore.
`storeCode` is now required for storeView
`url` is now responsible for creating internal prefix. So `appendStoreCode` is not needed anymore (It is left for compatibility)
For example:
```
{
  storeCode: 'de',
  appendStoreCode: true
}
```
Results in `/de` prefix for routes. Same thing can be done with:
```
{
  storeCode: 'de',
  url: '/de'
}
```
We can also create different url (https://github.com/DivanteLtd/vue-storefront/issues/3614) then `/${storeCode}`:
```
{
  storeCode: 'de',
  url: '/de_de'
}
```

Previously `mapStoreUrlsFor` needed to has same values as storeViews object props **and** storeCode. Now its changed and it's only related with storeViews object props.
Prev:
```
    "storeViews": {
      "mapStoreUrlsFor": ["de"],// 1. this need to match 2. and 3.
      "de": { // 2.
        "storeCode": "de" // 3.
      }
    }
```
Now:
```
    "storeViews": {
      "mapStoreUrlsFor": ["german"],// 1. this need to match 2.
      "german": { // 2.
        "storeCode": "de" // 3.
      }
    }
```
`mapStoreUrlsFor` is only used to disable storeView.

`defaultStoreCode` it is recommended to set some value for it. It is used only used when `multistore` is enabled. You can think about `defaultStoreCode` as fallback for multistore. For example: if you define German store on `"url": "/de"` and Italian store on `"url": "/it"`. Then you can decide which of those stores will be default one and will be reached on `"/"`. If you set `"defaultStoreCode": "de"` then on `"/"` you will have German store.

### Which Environment This Relates To
Check your case. In case of any doubts please read about [Release Cycle](https://docs.vuestorefront.io/guide/basics/release-cycle.html)

- [x] Test version (https://test.storefrontcloud.io) - this is a new feature or improvement for Vue Storefront. I've created branch from `develop` branch and want to merge it back to `develop`
- [ ] RC version (https://next.storefrontcloud.io) - this is a stabilisation fix for Release Candidate of Vue Storefront. I've created branch from `release` branch and want to merge it back to `release`
- [ ] Stable version (https://demo.storefrontcloud.io) - this is an important fix for current stable version. I've created branch from `hotfix` or `master` branch and want to merge it back to `hotfix`

### Upgrade Notes and Changelog
[...]

PR #5263, opened 2020-11-30
Use `queryString` library in `task/getUrl`

cewald:
### Related Issues

Related #5043

The new `adjustMultistoreApiUrl` method now uses url-encoding with `queryString.stringifyUrl()`. This leads into the problem that `getUrl()` method can't replace the tokens by simply using `replace`.

I refactored the `getUrl` method to use the `queryString` (`parseUrl`/`stringifyUrl`) library as it already is in us it for the `tokenInHeader` option.

### DIFF
--- a/core/lib/sync/task.ts
+++ b/core/lib/sync/task.ts
@@ -95,7 +95,6 @@ function _internalExecute (resolve, reject, task: Task, currentToken, currentCar
   }
   const url = getUrl(task, currentToken, currentCartId)
   const payload = getPayload(task, currentToken)
-  console.error('url', task, url, currentToken, currentCartId)
   let silentMode = false
   Logger.info('Executing sync task ' + url, 'sync', task)()
   return fetch(url, payload).then((response) => {
--- a/core/lib/sync/task.ts
+++ b/core/lib/sync/task.ts
@@ -31,19 +31,25 @@ function _sleep (time) {
 }
 
 function getUrl (task, currentToken, currentCartId) {
-  let url = task.url
-    .replace('{{token}}', (currentToken == null) ? '' : currentToken)
-    .replace('{{cartId}}', (currentCartId == null) ? '' : currentCartId)
+  const parsedUrl = queryString.parseUrl(task.url)
+
+  if (parsedUrl.query.token && parsedUrl.query.token === '{{token}}') {
+    parsedUrl.query.token = (currentToken == null) ? '' : currentToken
+  }
+
+  if (parsedUrl.query.cartId && parsedUrl.query.cartId === '{{cartId}}') {
+    parsedUrl.query.cartId = (currentCartId == null) ? '' : currentCartId
+  }
+
+  if (config.users.tokenInHeader) {
+    delete parsedUrl['query']['token']
+  }
+
+  let url = queryString.stringifyUrl(parsedUrl)
 
   url = processURLAddress(url); // use relative url paths
   if (config.storeViews.multistore) {
     url = adjustMultistoreApiUrl(url)
-  }
-
-  if (config.users.tokenInHeader) {
-    const parsedUrl = queryString.parseUrl(url)
-    delete parsedUrl['query']['token']
-    url = queryString.stringifyUrl(parsedUrl)
   }
 
   return url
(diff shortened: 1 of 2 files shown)


core/lib/sync/task.ts
```diff
@@ -31,19 +31,25 @@ function _sleep (time) {
...
 function getUrl (task, currentToken, currentCartId) {
-  let url = task.url
-    .replace('{{token}}', (currentToken == null) ? '' : currentToken)
-    .replace('{{cartId}}', (currentCartId == null) ? '' : currentCartId)
+  const parsedUrl = queryString.parseUrl(task.url)
 
-  url = processURLAddress(url); // use relative url paths
-  if (config.storeViews.multistore) {
-    url = adjustMultistoreApiUrl(url)
+  if (parsedUrl.query.token && parsedUrl.query.token === '{{token}}') {
```
Fifciu, line 36: I think you could use:
```js
if (parsedUrl.query && parsedUrl.query.token === '{{token}}') {
```

You could even go one step forward and use:
```ts
if (parsedUrl?.query?.token === '{{token}}') {
```

core/lib/sync/task.ts
```diff
@@ -31,19 +31,25 @@ function _sleep (time) {
...
+    parsedUrl.query.token = (currentToken == null) ? '' : currentToken
+  }
+
+  if (parsedUrl.query.cartId && parsedUrl.query.cartId === '{{cartId}}') {
```
Fifciu, line 40: Same as above

Fifciu (changes requested):

core/lib/sync/task.ts
cewald: Yes indeed, I've changed it.

Fifciu (approved):

gibkigonzo:
> This leads into the problem that getUrl() method can't replace the tokens by simply using replace.

What do you mean by that? 🤔
`adjustMultistoreApiUrl` should work in same way as before multistore refactor. As I understood it just replace storeCode in url and should return url. 
Token replacement in `getUrl` takes place before `adjustMultistoreApiUrl` is called so how it is related?

cewald:
`adjustMultistoreApiUrl` is working fine, except it makes `%7B%7Btoken%7D%7D` out of `{{token}}` – and that is why a simple string-replacement, as it was before, won‘t work.

gibkigonzo:
oww ok, I get it. But still `adjustMultistoreApiUrl` is called after `.replace`
replace => https://github.com/DivanteLtd/vue-storefront/blob/develop/core/lib/sync/task.ts#L34-L36
then `adjustMultistoreApiUrl` =>https://github.com/DivanteLtd/vue-storefront/blob/develop/core/lib/sync/task.ts#L40
So when `adjustMultistoreApiUrl` is called then `{{token}}` should be already replaced.

imo your refactor looks better then it was so it's good for me 👍

gibkigonzo (approved):

cewald:
Just for explanation: `adjustMultistoreApiUrl` is also used in `processLocalizedURLAddress` which again formats the endpoint-urls for the data-resolver-services. That's why it is called first and, if needed, afterwards in the `task/getUrl`.

gibkigonzo:
imo then we need to fix `adjustMultistoreApiUrl`, because `processLocalizedURLAddress` can be used in other places :thinking: Maybe there is some option in `queryString` to not transform `{` into `%7B%7B`

cewald:
Imo there is nothing to fix in `adjustMultistoreApiUrl` – it's actually good and probably more secure to url-encode the query-params using `queryString`. Actually, how it was done before was a bit crude – now it's consistent on both ends.

Anyway, as far as I can see `processLocalizedURLAddress` is currently only used in most data-resolver-services. And `adjustMultistoreApiUrl` only in the `processLocalizedURLAddress` method, the `getFallbackByUrl` action for url-mapping (actually same as data-resolver-service) and the task-lib itself. Should be allright.

Wdyt?

Fifciu:
@cewald please update the changelog and we could merge it

cewald:
@Fifciu done :)

Fifciu:
Great job, thanks!

PR #5361, opened 2020-12-30
Api Middleware

andrzejewsky:
Moves extensions and API clients to the server-side via Nuxt middleware.

# How it works?
Each integration will expose two bundles of API-client. First one to manage direct connection between VSF and commerce platform, and the second one to make proxy and redirection to our middleware.

When you call the proxy version of API-client, it will make a call to our middleware with the name of function you have called as a parameter, for example:

```js
context.$api.ct.getProduct() // calls /api/ct/getProduct
context.$api.ct.addToCart() // calls /api/ct/addToCart
context.$api.ct.removeFromCart() // calls /api/ct/removeFromCart
// etc.
```

Now, our middleware, exposed on `/api` will handle this by using the right API-client, shared from the server bundle - so the connection between commerce and VSF is made only on the server-side.

Changes:

## Integrators
Right now we have two factories instead of one.
- `apiClientFactory` - mostly is the same as the previous one, it will call a commerce integration directly, but it also allows you to add an extension
- `apiProxyFactory` - this one will redirect each API function to our middleware and it will be used only on the client side

### Exposing two entrypoints
We need to split API-client to the two bundles: first one that will be added to the client, and second one is being called on the server-side (middleware).

1. Create file `client.index.ts`
This file will expose a proxied version of your API-client that will redirect called function to our middleware

2. Create file `server.index.ts`
In general, this is the same API-client creation as it used to be so far. This one will be calling integration API (commerce) but this time it will be used in our middleware

3. Create file `index.ts`
This fill will be sharing only common things for both server and client bundle, such as typings, functions, helpers and so on.

## Usage
To use it, you have to add a module called `@vue-storefront/middleware/nuxt`  that you need to add under the `modules` section in `nuxt.config.js`. That module will install a middleware and will allow you to configure it (extensions, header redirections, extending endpoints, and more - TBD)


package.json
```diff
@@ -98,5 +98,8 @@
   },
   "lint-staged": {
     "*.{js,ts,vue}": "eslint --fix"
+  },
+  "dependencies": {
```
filrak, line 102: Avoid dependencies in root package json - they will not be included in the released packages

packages/commercetools/theme/nuxt.config.js
```diff
@@ -81,6 +84,7 @@ export default {
     ['@vue-storefront/nuxt-theme'],
     project-only-end */
     ['@vue-storefront/commercetools/nuxt', {
+      useMiddleware: true,
```
filrak, line 87: It should be true by default as disabling middleware is basically cutting you off from any extensibility and is just for hacking purposes

packages/core/core/src/factories/apiClientFactory/_utils.ts
```diff
@@ -0,0 +1,13 @@
+import { integrationPluginFactory } from './../../utils';
+
+export const compose = ({ createApiClient, createApiProxy, factoryParams }: any) => {
```
filrak, line 3: not so descriptive, compose what?

packages/core/middleware/module.js
```diff
@@ -0,0 +1,5 @@
+const createProxyMiddleware = require('./middleware');
```
filrak, line 1: should be `nuxt/middleware`

packages/core/middleware/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@vue-storefront/middleware",
```
filrak, line 2: nuxt-middleware

filrak (changes requested):

packages/commercetools/api-client/src/index.server.ts
```diff
@@ -64,12 +50,47 @@ const onSetup = (settings: Config): { config: Config; client: ClientInstance } =
...
-const { createApiClient } = apiClientFactory<Config, any>({
+const parseToken = (rawToken) => {
+  try {
+    return JSON.parse(rawToken);
+  } catch (e) {
+    return null;
+  }
+};
+
+const tokenExtension = (req, res) => {
```
filrak, line 61: typing

packages/commercetools/api-client/src/index.server.ts
```diff
@@ -64,12 +50,47 @@ const onSetup = (settings: Config): { config: Config; client: ClientInstance } =
...
+          delete req.cookies['vsf-commercetools-token'];
+        }
+      }
+    })
+  };
+};
+
+const { createApiClient, integrationPlugin } = apiClientFactory({
   tag: 'ct',
   onSetup,
```
filrak, line 88: `onCreate` makes much more sense now

packages/core/core/src/factories/apiFactory/types.ts
```diff
@@ -0,0 +1,51 @@
+import { IntegrationPlugin } from './../../utils/nuxt';
+export interface ApiExtensionInstance {
+  beforeSetup?: (config, headers?: Record<string, string>) => any;
```
filrak, line 3: why not object

packages/core/core/src/factories/apiFactory/types.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type ApiExtension = (req: any, res: any) => ApiExtensionInstance;
+
+export interface BaseApiFactoryParams<T, F = any> {
+  tag: string;
+  api: F;
+  isProxy?: boolean;
+}
+
+export interface ApiFactoryParams<T, F = any> extends BaseApiFactoryParams<T, F> {
```
filrak, line 17: isn't that only for ApiClientFactory (so no proxy)? I assume ApiFactoryParams means Api(Client|Proxy)FactoryParams

packages/core/core/src/factories/apiFactory/types.ts
```diff
@@ -0,0 +1,51 @@
+import { IntegrationPlugin } from './../../utils/nuxt';
```
filrak, line 1: just please check this https://github.com/vuestorefront/vue-storefront/pull/5361/files#diff-94593f73bba756dabad3c9eaca20619745bbae240f966b5fceb7802194df1dd0 and adjust the file ;)

packages/core/core/src/utils/context/index.ts
```diff
@@ -10,11 +10,23 @@ const configureContext = (config: ContextConfiguration) => {
   useVSFContext = config.useVSFContext || useVSFContext;
 };
 
-const applyContextForApi = (api, context) =>
+const applyContextForApi = (api, context, extensions = []) =>
```
filrak, line 13: https://github.com/kettanaito/naming-cheatsheet#naming-functions

are you applying context _for_ api or _to_ API?

packages/core/core/src/utils/nuxt/index.ts
```diff
@@ -1,9 +1,12 @@
 import { nuxtContextFactory } from './nuxtContextFactory';
```
filrak, line 1: why context is prefixed with `nuxt` and integration plugin isn't? both are for nuxt

packages/core/middleware/middleware.js
```diff
@@ -0,0 +1,54 @@
+const express = require('express');
```
filrak, line 1: shouldn't whole `middleware` dir be in `nuxt` dir?

packages/core/middleware/middleware.js
```diff
@@ -0,0 +1,54 @@
...
+
+function loadApiClient (apiClientPackageName, { req, res }) {
+  const apiClientPackage = require(apiClientPackageName);
+  const context = { middleware: { req, res } };
+  const createApiClient = apiClientPackage.createApiClient.bind(context);
+
+  return { ...apiClientPackage, createApiClient };
+}
+
+function createProxyMiddleware (moduleOptions, nuxtOptions) {
```
filrak, line 24: shouldn't this be just `createProxyEndpoint` ? middleware is already there

packages/core/middleware/middleware.js
```diff
@@ -0,0 +1,54 @@
...
+  app.post('/:integrationName/:functionName', async (req, res) => {
```
filrak, line 25: **minor** side effect

packages/core/docs/integrate/integration-guide.md
```diff
@@ -173,29 +196,16 @@ Each API function always contains `context` as a first parameter. This is the pl
...
-export const integrationPlugin = integrationPluginFactory(createApiClient);
-```
-
-That will expose the integration plugin wrapper, so the Nuxt will be able to create your integration within the context or extend already existing one, using a plugin
-
 ### 2.2 Creating nuxt module
 
 Inside of the composables packages you have to create another directory, next to `src` called `nuxt`. In that directory we need to place our Nuxt module. The Nuxt module is taking care of anything you want during the integration to be launched: adding plugins, injecting into the build process, creating some aliases, and more. The basic implementation of that module will add just a plugin that will configure our application (using a wrapper that you have already exposed)
 
+The plugin should also handle the option `useMiddleware` which is responsible for loading certain api-client. When `useMiddleware` is set to true, we need to load proxy version of api-client and start redirecting the traffic. Otherwise, we use direct connection.
```
filrak, line 203: if we're usnure about this feature for CT (if it isn't redundant) should we require this from others?

packages/core/docs/integrate/integration-guide.md
```diff
@@ -103,58 +103,81 @@ Once you copied and renamed the boilerplate run `yarn dev` in your `theme` folde
 
 Each integration starts with `api-client`. This is one of the packages which is responsible for communication between the Vue Storefront and external API. That's exactly the place where you have to configure your api connection, write your api functions, and expose generated api client to the users.
 
+Our api client has two parts: the proxy layer (that talks to our middleware) and a direct connection and as result, package of api-client always shares three entrypoints (treeshaking reasons):
+
+- `@vue-storefront/integration/client` - shares the `createProxyClient` and `integrationPlugin` for proxy
+- `@vue-storefront/integration/server` - shares the `createApiClient` and `integrationPlugin` for direct connection
+- `@vue-storefront/integration` - shares other library stuff, such a helpers, types etc.
```
filrak, line 110:
```suggestion
- `@vue-storefront/{INTEGRATION}/client` - shares the `createProxyClient` and `integrationPlugin` for proxy
- `@vue-storefront/{INTEGRATION}/server` - shares the `createApiClient` and `integrationPlugin` for direct connection
- `@vue-storefront/{INTEGRATION}` - shares other library stuff, such a helpers, types etc.
```

Let's stick to this all over our docs

packages/core/core/src/factories/apiFactory/types.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export interface FactoryInstance {
+  integrationPlugin: IntegrationPlugin;
+}
+
+export type CreateApiProxyFn = (givenConfig: any, customApi?: any) => ApiInstance
+export type CreateApiClientFn = (givenConfig: any, customApi?: any) => ApiInstance;
+
+export interface ProxyFactoryInstance extends FactoryInstance {
```
filrak, line 39: be consistant, I see here all variations of words "factory" "API" and "proxy" used in different order and in different combinations

Let's stick to one naming for
- apiClientFactory
- apiProxyFactory
- a unified type for both

packages/core/core/src/factories/apiFactory/apiProxyFactory.ts
```diff
@@ -0,0 +1,32 @@
+import axios from 'axios';
+import { applyContextForApi } from './../../utils/context';
+import { Logger } from './../../utils';
+import { createInstance } from './_utils';
+import { ApiProxyFactoryParams, ApiInstance, ProxyFactoryInstance } from './types';
+import { getConfig, createProxy } from './_proxyUtils';
+
+const apiProxyFactory = <ALL_SETTINGS, ALL_FUNCTIONS>(factoryParams: ApiProxyFactoryParams<ALL_SETTINGS, ALL_FUNCTIONS>): ProxyFactoryInstance => {
+  function createApiProxy (config: any, customApi: any = {}): ApiInstance {
+    const _config = getConfig({ context: this, factoryParams, givenConfig: config });
```
filrak, line 10:
```suggestion
    const _config = getConfig({ context: this, factoryParams, givenConfig: config });
```

get config of what?

Maybe `getIntegrationConfig` ?

packages/core/core/src/factories/apiFactory/_utils.ts
```diff
@@ -0,0 +1,21 @@
+import { integrationPluginFactory } from './../../utils';
+import { CreateApiProxyFn, CreateApiClientFn } from './types';
+
+interface CreatorFunctions {
+  createApiClient?: CreateApiClientFn;
+  createApiProxy?: CreateApiProxyFn;
+}
+
+export const createInstance = <T>(creatorFunctions: CreatorFunctions, factoryParams: any) => {
```
filrak, line 9: instance of what?

packages/core/core/src/factories/apiFactory/apiProxyFactory.ts
```diff
@@ -0,0 +1,32 @@
...
+    const client = axios.create(_config.axios);
+    const settings = { client, config: _config };
+
+    (settings as any).isProxy = true;
+
+    Logger.debug('apiProxyFactory.setup', settings);
+
+    const givenApi = applyContextForApi({ ...factoryParams.api, ...customApi }, settings);
+
+    const api = createProxy({ givenApi, client, factoryParams });
```
filrak, line 20: again - lets stick to one naming convention for api proxy => `createApiProxy` (?)

packages/commercetools/theme/nuxt.config.js
```diff
@@ -56,6 +56,9 @@ export default {
     '@nuxtjs/pwa',
     ['@vue-storefront/nuxt', {
       coreDevelopment: true,
+      logger: {
```
filrak, line 59: shouldn't we leave it like it is by default?

packages/core/core/src/factories/apiFactory/types.ts
andrzejewsky: extension always is depending on request so it's, more clean to have function that takes request and now you have access to it across the lifecycle functions instead of adding extra argument for each:

```js
const extension = (req) => ({
 beforeSetup: (config) => { have access to req }
 beforeCall: (params) => { have access to req }
 ...
})
```

instead of:


```js
const extension = {
 beforeSetup: (req, config) => { extra argument needed }
 beforeCall: (req, params) => {extra argument needed}
}
```

packages/core/core/src/utils/nuxt/index.ts
andrzejewsky: you import `integrationPlugin` only in the plugin file of nuxt.js, so that fact actually means that you are using plugin, so It doesn't need to be extra prefixed

packages/core/middleware/middleware.js
andrzejewsky: added module to `/nuxt` directory and middleware will stay in the root - that's for the future if we want to use middleware as separated app

packages/core/core/src/factories/apiFactory/types.ts
andrzejewsky: unified type? they return different things and have different parameters

filrak: then only 2 of those ;)

packages/about-you/api-client/src/index.ts
```diff
@@ -6,7 +6,7 @@ let apiClient: BapiClient | null = null;
 
 const { setup, update, getSettings } = apiClientFactory<any, any>({
   defaultSettings: {},
-  onSetup: (setupConfig: SetupConfig) => {
+  onCreate: (setupConfig: SetupConfig) => {
```
filrak, line 9: ayc is not aligned with this version anyway so I'd leave it on lower one and don't change anything

packages/core/core/src/factories/apiFactory/_proxyUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+      return Reflect.get(target, prop, receiver);
+    }
+
+    return async (...args) => client
+      .post(`/${factoryParams.tag}/${functionName}`, args)
+      .then(r => r.data);
+  }
+});
+
+export const getCookies = (context) => (context && context.req && context.req.headers.cookie) || '';
```
Fifciu, line 26: Optional chaining + null coalescing?

packages/core/core/src/factories/apiFactory/_proxyUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const getIntegrationConfig = ({ context, factoryParams, givenConfig }) => {
+  const cookie = getCookies(context);
+  const initialConfig = merge({
+    axios: {
+      baseURL: getBaseUrl(context && context.req),
```
Fifciu, line 32: Opt chaining

packages/core/core/src/factories/apiFactory/_proxyUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+      headers: {
+        ...(cookie ? { cookie } : {})
```
Fifciu, line 34: Not needed spread

packages/core/core/src/factories/apiFactory/apiClientFactory.ts
```diff
@@ -0,0 +1,37 @@
+import { ApiClientFactoryParams, ApiClientConfig, ApiInstance, ApiClientFactory } from './types';
+import { applyContextToApi } from './../../utils/context';
+import { createApiInstance } from './_utils';
+import { Logger } from './../../utils';
+
+const apiClientFactory = <ALL_SETTINGS extends ApiClientConfig, ALL_FUNCTIONS>(factoryParams: ApiClientFactoryParams<ALL_SETTINGS, ALL_FUNCTIONS>): ApiClientFactory => {
+  function createApiClient (config: any, customApi: any = {}): ApiInstance {
+    const extensions = factoryParams.extensions && this && this.middleware
+    // eslint-disable-next-line
+    ? Object.entries(factoryParams.extensions).map(([_, extensionFn]) => extensionFn(this.middleware.req, this.middleware.res))
```
Fifciu, line 10: Let's just use `Object.values`

packages/core/core/src/factories/apiFactory/types.ts
```diff
@@ -0,0 +1,51 @@
+import { IntegrationPlugin } from './../../utils/nuxt';
+export interface ApiClientExtensionLifecycle {
+  beforeCreate?: (config, headers?: Record<string, string>) => any;
```
Fifciu, line 3: I wonder if we should transform `config` to the `{ config }` to unify it with `afterCreate` and have easy possibility to extend it without BC in the future. Wdyt?

packages/core/docs/integrate/integration-guide.md
```diff
@@ -103,58 +103,81 @@ Once you copied and renamed the boilerplate run `yarn dev` in your `theme` folde
...
-- `tag` - that's the short name of your integration which will be used to distinguish it among of others
-- `onSetup` - an optional function that will be called during creating your API. In this place, you can call everything you need to create a connection to the API, such as creating SDK (eg. `axios` creation), merge given config with the defaults etc.
-- `api` - this is the section where you need to pass all of the API function you have created
+To create `api-client` instances you have to use correspoing factory, depending on what api you are creating: proxy or direct one. The creation in both cases in pretty similar, with small differences in the used fields:
+
+- `tag` - that's the short name of your integration which will be used to distinguish it among of others 
+- `onCreate` - a function that will be called during creating your API. In this place, you can call everything you need to create a connection to the API, such as creating SDK (eg. axios creation), merge given config with the defaults etc.. This function always returns `client` (connection you created) and `config` or (in case it's proxy) just `config`.
```
Fifciu, line 178: eg -> `e.g.`

packages/core/docs/integrate/integration-guide.md
```diff
@@ -103,58 +103,81 @@ Once you copied and renamed the boilerplate run `yarn dev` in your `theme` folde
...
+- `api` - this is the section where you need to pass all of the API function you have created (direct) and functions that you don't want to redirect to our middleware (proxy)
+- `extensions` - section available only in the direct connection api-client. It allows you to add an api backend extenstions for the api taht can add additiona features to the integrated platform
```
Fifciu, line 180: `taht` -> `that`,
`additiona` -> `additional`

Fifciu (changes requested):

packages/core/core/src/factories/apiFactory/_proxyUtils.ts
andrzejewsky: if the cookie is null I don't want to add a field cookie at all

packages/core/docs/integrate/integration-guide.md
andrzejewsky: docs will be updated in the separated pr

packages/core/core/src/factories/apiFactory/types.ts
andrzejewsky: sure, this is just the beginning of the extendibility, I'll create many tasks related to this.

packages/core/core/src/factories/apiFactory/_proxyUtils.ts
Fifciu: I meant: `headers: cookie ? { cookie } : {}`

packages/core/core/src/factories/apiFactory/apiProxyFactory.ts
```diff
@@ -0,0 +1,32 @@
...
+    const client = axios.create(_config.axios);
+    const settings = { client, config: _config };
+
+    (settings as any).isProxy = true;
+
+    Logger.debug('apiProxyFactory.create', settings);
+
+    const givenApi = applyContextToApi({ ...factoryParams.api, ...customApi }, settings);
+
+    const api = createProxiedApi({ givenApi, client, factoryParams });
```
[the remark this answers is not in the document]
andrzejewsky: create - this function returns a JS Proxy based on given params

Issue #5385, opened 2021-01-07
Can't go to the Payment checkout step


filrak:
![image](https://user-images.githubusercontent.com/15185752/103930248-bc014280-511e-11eb-9465-c1a0c4400485.png)
Click "select shipping method"
see error in console
![image](https://user-images.githubusercontent.com/15185752/103930305-cfaca900-511e-11eb-9608-6804d325b6d4.png)

Fifciu:
It is important to also check if we can go from Billing step to Order Review step. I am 99% sure - we can't. Problem appears due to changes in SFUI & bad handling it inside our app. 

As far as I remember, in country select (`pages/Checkout/Shipping.vue` line 94) we have to change @selected to @input, e.g:
```js
@input="setShippingDetailsAndUnpickAddress({ country: $event })"
```

Also as a workaround during work on CKO, in line 118 of `pages/Checkout/Payment.vue` I used:
```js
:value="billingDetails.phone || billingDetails.contactInfo.phone"
```

bloodf:
This issue has been closed due to inactivity. 

If you want to re-open the issue, please re-create the issue, and post newer information on this problem.

PR #5386, opened 2021-01-08
Fix select country on checkout

andrzejewsky:
### Related Issues
closes https://github.com/vuestorefront/vue-storefront/issues/5385


filrak:
@filipsobol most likely a change for BP as well

filrak (approved):

PR #5418, opened 2021-01-18
Audit docs: part 2

NataliaTepluhina:
## Short Description and Why It's Useful

This PR is a part of the audit of existing next docs before we start working on missing parts

### DIFF
--- a/packages/core/docs/advanced/logging.md
+++ b/packages/core/docs/advanced/logging.md
@@ -32,9 +32,9 @@ You can set the `verbosity` level which tells the app what do you want to log an
   '@vue-storefront/nuxt',
   {
     logger: {
-      verbosity: 'error',
-    },
-  },
+      verbosity: 'error'
+    }
+  }
 ];
 ```
 
@@ -49,9 +49,7 @@ If not explicitly changed, logging level depends on the current environment vari
 
 By default, we're printing all the events happening in the app in the console but you can easily write a new logger and use a third-party library (like [consola](https://github.com/nuxt-contrib/consola)) or pass the logs to the external service like [Sentry](https://sentry.io/welcome/)
 
-To override the default logger, pass a function to the `logger.customLogger` property of `@vue-storefront/nuxt` module. This function returns the logger object and as an argument, you have access to the `verbosity` level.
-
-If you are using our Nuxt module, setting up the logger is much simpler. All you have to do is just provide the configuration to the module:
+To override the default logger, pass a function to the `logger.customLogger` property of `@vue-storefront/nuxt` module. This function returns the logger object and as an argument, you have access to the `verbosity` level:
 
 ```js
 [
@@ -69,11 +67,11 @@ If you are using our Nuxt module, setting up the logger is much simpler. All you
           warn: (message: any, ...args) =>
             console.warn('[VSF][warn]', message, ...args),
           error: (message: any, ...args) =>
-            console.error('[VSF][error]', message, ...args),
+            console.error('[VSF][error]', message, ...args)
         };
-      },
-    },
-  },
+      }
+    }
+  }
 ];
 ```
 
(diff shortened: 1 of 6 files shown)


packages/core/docs/advanced/logging.md
```diff
@@ -44,44 +47,56 @@ If not explicitly changed, logging level depends on current environment variable
 
 ## Writing custom loggers
 
-By default we're printing all the events happening in the app in the console  but you can easily write a new logger and use a third-party library (like `consola`) or pass the logs to external service like `Sentry.io`
+By default, we're printing all the events happening in the app in the console but you can easily write a new logger and use a third-party library (like [consola](https://github.com/nuxt-contrib/consola)) or pass the logs to the external service like [Sentry](https://sentry.io/welcome/)
 
-To override the default logger simply pass a function to the `logger.customLogger` property of `@vue-storefront/nuxt` module. This function returns the logger object and as an argument, you have access to the `verbosity` level.
-
-If you are using our nuxt module, setting up the logger is much simpler. All you have to do is just provide the configuration to the module:
+To override the default logger, pass a function to the `logger.customLogger` property of `@vue-storefront/nuxt` module. This function returns the logger object and as an argument, you have access to the `verbosity` level.
```
NataliaTepluhina, line 52: @filrak we definitely need a code example here

filrak: this refers to line 57, perhaps this description is just obsolete (at least part of this that describes the example)

packages/core/docs/advanced/logging.md
```diff
@@ -44,44 +47,56 @@ If not explicitly changed, logging level depends on current environment variable
 
 ## Writing custom loggers
 
-By default we're printing all the events happening in the app in the console  but you can easily write a new logger and use a third-party library (like `consola`) or pass the logs to external service like `Sentry.io`
+By default, we're printing all the events happening in the app in the console but you can easily write a new logger and use a third-party library (like [consola](https://github.com/nuxt-contrib/consola)) or pass the logs to the external service like [Sentry](https://sentry.io/welcome/)
```
andrzejewsky, line 51: I think winston logger is a way better example here: https://github.com/winstonjs/winston

packages/core/docs/guide/error-handling.md
```diff
@@ -36,26 +37,23 @@ export interface UseCartErrors {
 }
 ```
 
-
 :::details Where does the error come from?
 
-Inside each async method we are clearing the current error before integration's method call and setting it in catch block.
+Inside each async method, we are clearing the current error before an integration's method call and setting it in the `catch` block.
+
```
andrzejewsky, line 43: I would delete the example below as it's a part of core, while core is something that end-user doesn't have to be aware of. For him, the knowledge about what he can gets from error fields is enough. I know it was merged as it is previously, but now when I'm reading this it sounds like we are explaining our implementation instead of how this API works.

I suggest:
--- START --- 
Inside each async method, we are catching errors when they occur and save them to the reactive property called `errors` under the key corresponding to the triggered method. 

Example:
```js
const { addToCart, errors } = useCart()

addToCart({ product: null }) // triggers an error
errors.addToCart // here you have error raised by addToCart function 
```
--- END --- 

and ofc delete the current example

andrzejewsky (approved):
two small comments from my side, not that important but would be nice to have them, besides looks better 👍

packages/core/docs/advanced/logging.md
NataliaTepluhina: I removed an unnecessary explanation here 👍🏻

packages/core/docs/advanced/logging.md
NataliaTepluhina: @andrzejewsky you mean instead of Sentry or consola?

packages/core/docs/guide/error-handling.md
NataliaTepluhina: @andrzejewsky thank you for the suggestion! Done 👍🏻

PR #5451, opened 2021-01-27
Update middleware feature

andrzejewsky:
- the client side of api-client was deleted
- API client is always the proxied one that points to our middleware
- the API client requires only two entrypoints: one for server and middleware purpose and the second one just for client-side as reference for typings or other helpers


packages/core/core/src/types.ts
```diff
@@ -582,3 +582,38 @@ export interface Context {
 export interface FactoryParams {
   provide?: (context: Context) => any;
 }
+
+export interface ApiClientExtensionLifecycle {
+  beforeCreate?: (config, headers?: Record<string, string>) => any;
+  afterCreate?: ({ config, client }, headers?: Record<string, string>) => { config; client };
```
filrak, line 588: why headers are optional? I guess they're always included, also - why we're not returning them? are they readonly? if so you can specify this in TS

also its better to have everything in 1 object for maintainability

packages/core/core/src/types.ts
```diff
@@ -582,3 +582,38 @@ export interface Context {
 export interface FactoryParams {
   provide?: (context: Context) => any;
 }
+
+export interface ApiClientExtensionLifecycle {
+  beforeCreate?: (config, headers?: Record<string, string>) => any;
```
filrak, line 587: same as below also can we make this just:
```suggestion
  beforeCreate?: ({ config, headers: Record<string, string> }) => any;
```
for maintainability?

filrak (changes requested):

packages/core/core/src/utils/context/index.ts
```diff
@@ -4,29 +4,27 @@ interface ContextConfiguration {
...
+}
+
 let useVSFContext = () => ({}) as Context;
 
 const configureContext = (config: ContextConfiguration) => {
   useVSFContext = config.useVSFContext || useVSFContext;
 };
 
-const applyContextToApi = (api, context, extensions = []) =>
+const NOP = (x) => x;
```
filrak, line 18: Its worth describing why we need no-op here and what this fn does as its hard to understand from code

PR #5500, opened 2021-02-06
Middleware update

andrzejewsky:
## An update and improvements for middleware.

Goals / issues
- implement extensions for API-middleware
- add the ability for extending API-client by additional functions
- add the ability for running middleware outside of nuxt.js

#### 1. Extensions
Each API-client extension has the following signature:

```ts
export interface ApiClientExtensionLifecycle {
  beforeCreate?: ({ config }) => any;
  afterCreate?: ({ config }) => any;
  beforeCall?: ({ config, callName, args }) => any;
  afterCall?: ({ config, callName, args }) => any;
}

export interface ApiClientExtension {
  extendApi?: Record<string, Function>;
  lifecycle?: (req: any, res: any) => ApiClientExtensionLifecycle;
}
```
_Where:_
- `ApiClientExtensionLifecycle` - defines the lifecycle of the incoming request in the middleware. We are able to inject within any stage of calling API-client functions or its configuration.
- `ApiClientExtension`- has two properties. The first one is dedicated to extending an existing set of API-functions - if you want to add an additional function that can be called on the middleware, this section is for this. Second, lifecycle that refers to the type above

##### Adding extension
We have a few ways for adding an extension:

_pre-defined extension:_
In some of the integration, we need to have some extension that is necessary to run the integration. An example is commerce tools token extension.

```ts
import { apiClientFactory } from '@vue-storefront/core';

const tokenExtension = { ... }

const { createApiClient } = apiClientFactory({
  onCreate,
  api,
  extensions: [tokenExtension]
});
```
The exposed API-client has already added an extension for handling token, without that we can't run commercetools integration.

_from nuxt config:_
If this is needed, you can add your own extension using nuxt.config.js:

```
    ['@vue-storefront/commercetools/nuxt', {
      middleware: {
        extensions: [
          {
            extendApi: { ... },
            lifecycle: () => { ... }
        ]
      },
    }]
```

_from nuxt module:_
Sometimes we have one package that extends an existing one. For example, we have commercetools API-client open source and we want to add a feature (from the different package) that uses and extends the current API-client - we need to add additional functions. 

To solve that problem, we can call `registerIntegration` inside of nuxt module, which adds additional extensions to the existing integration. **In this case, we are adding package names that expose `extensions` field - the middleware will be looking for that field in the packages you have added here**

```ts
  this.nuxt.registerIntegration('ct', {
    extensions: [
      '@vue-storefront/ct-shipping'
    ]
  });
```

_from manual config:_
this is for using middleware separately, jump to the next section.

#### 2. Registering integration module
The integration modules and the middleware don't know about each other. The middleware is now aware of any integration in the system and doesn't know what API-clients need to be loaded. To register integration we need to somehow tell the VSF about incoming integration. To do such a thing we have module function - `registerIntegration`

```ts
  this.nuxt.registerIntegration('ct', {
    apiClientPackage: '@vue-storefront/commercetools-api/server',
    modulePackage: '@vue-storefront/commercetools/nuxt',
    configuration: options,
    extensions: [
      '@vue-storefront/commercetools'
    ]
  });

```
That function takes the tag name of the given integration as the first argument and the configuration as the second.
When it comes to the configuration:

`apiClientPackage` - tells the VSF where is API-client that will be called on the middleware _(**mandatory**)_
`modulePackage` - tell the VSF what's the integration module that contains a configuration _(**mandatory**)_
`configuration` - an additional configuration that will be merged into the original one
`extensions` - an array wth extensions, strings that represents packages with extensions, **each must expose `extensions` field**

####  3.Running as separated app

```
import { createServer } from '@vue-storefront/middleware'

const config = {
 integrations: {
  ct: {
    apiClientPackage: '@vue-storefront/commercetools-api/server',
    modulePackage: '@vue-storefront/commercetools/nuxt',
    configuration: { apiKey: '...' },
    extensions: [...]
  }
 }
}

const app = createServer(config)

app.listen(8080)
```

NOTES:
- _We need to keep middleware as a separate module as there is no other way to add this automatically. All of our modules are the buildModules so they can be triggered only in either build time or development time. Even it we copy this module to the `.nuxt` filter (which keeps the build files) it needs to be added somehow in the production runtime._
- _We need to call `registerIntegration` all time in the new integrations, the VSF must somehow be aware of them, and know what to load_


packages/commercetools/composables/src/index.ts
```diff
@@ -15,7 +15,19 @@ import { track } from '@vue-storefront/core';
 
 track('VSFCommercetools');
 
+const extensions = [
```
andrzejewsky, line 18: to be deleted

packages/commercetools/theme/.gitignore
```diff
@@ -67,6 +67,7 @@ typings/
 
 # nuxt.js build output
 .nuxt
+.vsf
```
andrzejewsky, line 70: remove

packages/commercetools/theme/middleware.config.js
```diff
@@ -0,0 +1,39 @@
...
+        },
+        currency: 'USD',
+        country: 'US'
+      }
+    }
+  }
+};
+
+/**
+extensions -> function: (extensions) => extensions +
```
andrzejewsky, line 33: to be deleted

packages/core/core/package.json
```diff
@@ -18,7 +18,8 @@
     "vue": "^2.6.11",
     "lodash-es": "^4.17.15",
     "express": "^4.17.1",
-    "axios": "0.21.1"
+    "axios": "0.21.1",
+    "fs-extra": "^9.1.0"
```
andrzejewsky, line 22: to be removed

packages/core/core/src/server/memory/index.ts
```diff
@@ -0,0 +1,20 @@
+import fs from 'fs-extra';
```
andrzejewsky, line 1: to be deleted

packages/core/core/package.json
Fifciu: Why? You are using it in `core/core/src/server/memory/index.ts`

packages/core/core/src/utils/context/index.ts
```diff
@@ -28,9 +28,15 @@ const applyContextToApi = (
...
   Object.entries(api)
-    .reduce((prev, [key, fn]: any) => ({
+    .reduce((prev, [callName, fn]: any) => ({
       ...prev,
-      [key]: async (...args) => hooks.after(await fn(context, ...hooks.before(args)))
+      [callName]: async (...args) => {
+        const transformedArgs = hooks.before({ callName, args });
+        const response = await fn(context, ...transformedArgs);
+        const transformedResponse = hooks.after({ callName, args, response });
+
```
Fifciu, line 37: Wdyt about adding there Logger?

packages/core/middleware/src/createServer.ts
```diff
@@ -0,0 +1,44 @@
...
+import cors from 'cors';
+import consola from 'consola';
+import { registerIntegrations } from './integrations';
+
+const app = express();
+app.use(express.json());
+app.use(cookieParser());
+app.use(cors());
+
+const applyMiddlewreContext = (createApiClient, { req, res, extensions }) => {
```
Fifciu, line 12: Typo, same in line 31.

packages/core/middleware/nuxt/index.js
```diff
@@ -1,5 +1,9 @@
-const createProxyEndpoint = require('./../middleware');
+const { createServer } = require('@vue-storefront/middleware');
 
-module.exports = function VueStorefrontMiddleware (moduleOptions) {
-  this.addServerMiddleware(createProxyEndpoint(moduleOptions, this.options));
+module.exports = function VueStorefrontMiddleware () {
+  const { integrations } = require(this.nuxt.options.rootDir + '/middleware.config.js');
```
Fifciu, line 4: Wdyt about wrapping this line in try...catch and adding useful log for potential `middleware.config.js` without `.integrations`?

packages/core/middleware/nuxt/index.js
```diff
@@ -1,5 +1,9 @@
...
+  const handler = createServer({ integrations });
+  const serverMiddleware = {path: '/api', handler };
```
Fifciu, line 6: Remove space on right or add on left please :smile:

packages/core/middleware/package.json
```diff
@@ -1,17 +1,21 @@
...
+    "build": "rimraf lib && rollup -c"
   },
   "author": "Vue Storefront",
   "license": "MIT",
   "dependencies": {
+    "consola": "2.15.3",
     "express": "^4.17.1",
     "cookie-parser": "^1.4.5",
-    "is-https": "^3.0.2"
+    "is-https": "^3.0.2",
```
Fifciu, line 17: I see no usage of `is-https` in `core/middleware`

packages/core/middleware/src/createServer.ts
```diff
@@ -0,0 +1,44 @@
...
+  consola.success('Integrations loaded!');
+
+  app.post('/:integrationName/:functionName', async (req, res) => {
+    const { integrationName, functionName } = req.params;
+    const { apiClient, configuration, extensions } = integrations[integrationName];
+    const middlewareContext = { req, res, extensions };
+    const createApiClient = applyMiddlewreContext(apiClient.createApiClient, middlewareContext);
+    const apiClientInstance = createApiClient(configuration);
+    const apiFunction = apiClientInstance.api[functionName];
+    const platformResponse = await apiFunction(...req.body);
```
Fifciu, line 34: Have you tested it? I had problems with losing payload in proxy when tried to use it during you've been developing

packages/core/middleware/src/integrations.ts
```diff
@@ -0,0 +1,38 @@
+import consola from 'consola';
+
+const createRawExtensions = (apiClient, integrationConfig) => {
+  const extensionsCreateFn = integrationConfig.extensions;
+  const prederinedExtensions = apiClient.createApiClient._predefinedExtensions;
```
Fifciu, line 5: Typo in const's name, same x2 line below

Fifciu (changes requested):
Looks great, just found some cosmetic stuff and a few typos

packages/commercetools/theme/nuxt.config.js
```diff
@@ -81,25 +81,6 @@ export default {
     ['@vue-storefront/nuxt-theme'],
     project-only-end */
     ['@vue-storefront/commercetools/nuxt', {
-      api: {
-        uri: 'https://api.commercetools.com/vsf-ct-dev/graphql',
-        authHost: 'https://auth.sphere.io',
-        projectKey: 'vsf-ct-dev',
-        clientId: 'RT4iJGDbDzZe4b2E6RyeNe9s',
-        clientSecret: '5eBt3yfZJWw1j7V6kXjfKXpuFP-YQXpg',
```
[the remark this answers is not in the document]
andrzejewsky: magic my son

packages/core/core/src/factories/apiClientFactory.ts
```diff
@@ -1,37 +1,39 @@
...
-    settings.config = extensions
+    settings.config = lifecycles
       .filter(ext => ext.afterCreate)
-      .reduce((prev, curr) => curr.afterCreate(prev), settings.config);
+      .reduce((prev, curr) => curr.afterCreate({ configuration: prev }), settings.config);
 
     const extensionHooks = {
-      before: (args) => extensions
+      before: (params) => lifecycles
         .filter(e => e.beforeCall)
```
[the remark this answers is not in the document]
andrzejewsky: you don't have to, we force users to use functions here, it doesn't have to that bulletproof, also we provide config reference / TS types for this so...

packages/core/core/src/utils/context/index.ts
andrzejewsky: logging is something that I'll be considering in the new issue: what we want to log, what verbosity etc.

packages/core/middleware/nuxt/index.js
andrzejewsky: I have to write all handling errors, such as handling 404, 500 etc. but that's for another issue - don't want to make this PR bigger and bigger

packages/core/middleware/package.json
andrzejewsky: used in _proxyUtils

packages/core/middleware/src/createServer.ts
andrzejewsky: we'll be testing this together once is merged

packages/core/middleware/package.json
Fifciu: But `_proxyUtils` is inside `core/core` which has own `package.json`. Not inside `core/middleware`

filrak (approved):

Fifciu (approved):
Okejka

PR #5501, opened 2021-02-08
Update VSF1 SSR Cache docs

filipsobol:
[yarn cache](https://classic.yarnpkg.com/en/docs/cli/cache/) is a internal `yarn` command. To run script named `cache` it's required to use `yarn run` command.


Fifciu (approved):
Thanks for this PR

PR #5504, opened 2021-02-08
Docs: composables anatomy

NataliaTepluhina:
Added composables anatomy chapter to docs-next

* [x] I read and followed [contribution rules](https://github.com/DivanteLtd/vue-storefront/blob/master/CONTRIBUTING.md)

### DIFF
--- a/packages/core/docs/guide/composables.md
+++ b/packages/core/docs/guide/composables.md
@@ -226,7 +226,9 @@ However, the example above won't work without `onSSR` helper. Let's take a look 
 
 ### Using `onSSR` for server-side rendering
 
-By default, Vue Storefront supports SSR and shared-state using Nuxt.js features. Furthermore, we can't use asynchronous that depend on each other calls in the `setup` function(e.g. loading products by the id of category that you have to fetch first). To solve this issue, we provide a temporary solution - `onSSR`:
+By default, Vue Storefront supports conveying server-side data to the client with Nuxt.js 2 where `setup` function is synchronous. Because of that, we can't use asynchronous functions if their results depend on each other (e.g. by loading `products` using `id` of a category that you have to fetch first).
+
+To solve this issue, we provide a temporary solution - `onSSR`:
 
 ```js
 import { useProduct } from '{INTEGRATION}';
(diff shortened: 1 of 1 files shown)


packages/core/docs/guide/composables.md
```diff
@@ -203,3 +203,53 @@ export default {
 ```
 
 For some composables (like `useProduct` ) you will need to pass a unique ID as a parameter (it can be a product ID, category ID etc.). Others (like `useCart`) do not require an ID passed. You can always check a composable signature in the [API Reference](TODO)
+
+## Anatomy of a Composable
+
+Every Vue Storefront composable usually returns three main pieces:
+
+- **Main data object**. A read-only object that represents data returned by the composable. For example, in `useProduct` it's a `products` object, in `useCategory` it's `category` etc.
+- **Supportive data objects**. These properties depend directly on indirectly on the main data object. For example, `loading`, `error` or `isAuthenticated` from `useUser` depend on a `user`.
```
filrak, line 212:
```suggestion
- **Supportive data objects**. These properties depend directly or indirectly on the main data object. For example, `loading`, `error` or `isAuthenticated` from `useUser` depend on a `user`.
```

packages/core/docs/guide/composables.md
```diff
@@ -203,3 +203,53 @@ export default {
...
+- **Main function that interacts with data object**. This function usually calls the API and updates the main data object. For example in `useProduct` and `useCategory` it's a `search` method,in `useCart` it's a `load` method. The rule of thumb here is to use `search` when you need to pass some search parameters. `load` is usually called when you need to load some content based on cookies or `localStorage`
```
filrak, line 213:
```suggestion
- **Main function that interacts with data object**. This function usually calls the API and updates the main data object. For example in `useProduct` and `useCategory` it's a `search` method, in `useCart` it's a `load` method. The rule of thumb here is to use `search` when you need to pass some search parameters. `load` is usually called when you need to load some content based on cookies or `localStorage`
```

filrak (approved):

filrak:
Great job as always, thank you!

PR #5524, opened 2021-02-11
Update Cache library

filipsobol:
- [X] (Next only) I've followed this [instruction](https://docs-next.vuestorefront.io/contributing/creating-changelog.html) and I've created a `.js` file with information about my Pull Request


packages/core/cache/nuxt/index.js
```diff
@@ -1,93 +1,111 @@
...
+    return;
+  }
+
+  this.addServerMiddleware({
+    path: options.endpoint,
+    handler: async (request, response) => {
+      try {
+        // Remove leading slash and get URL params
+        const params = new URLSearchParams(request.url.replace(/^\//, ''));
+        const tags = params.get('tags').split(',');
```
Fifciu, line 19: I wonder if we maybe should support some special value like `*` in VSF1 which would purge whole cache (e.g. after deploy). Wdyt?

Fifciu (changes requested):

packages/core/cache/nuxt/index.js
```diff
@@ -1,93 +1,111 @@
...
+/**
+ * Adds endpoint to invalidate cache
+ */
+function createInvalidationEndpoint (driver, options) {
+  if (!options) {
+    return;
+  }
+
+  this.addServerMiddleware({
+    path: options.endpoint,
```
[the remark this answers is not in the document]
filipsobol: Fixed

filrak (approved):

packages/core/cache/nuxt/index.js
```diff
@@ -1,93 +1,117 @@
...
+  if (!options || !options.endpoint || !options.key) {
+    return;
+  }
+
+  this.addServerMiddleware({
+    path: options.endpoint,
+    handler: async (request, response) => {
+      try {
+        // Remove leading slash and get URL params
+        const params = new URLSearchParams(request.url.replace(/^\//, ''));
```
andrzejewsky: remove lines from 18 to 24 - you can't assume that someone as an invalidator will send you tags in the URL - imageina a webhooks from certain commerce platform or others - based on their payload you need to build tags

packages/core/cache/nuxt/index.js
```diff
@@ -1,93 +1,117 @@
...
+        ? Array.from(context.req.$vsfCache.tagsSet)
+        : [];
+      
+      return tags.map(({ prefix, value }) => `${prefix}${value}`);
     }
 
     try {
-      return await resolvedDriver.invoke({ route, context, getTags, render });
+      return await driver.invoke({
+        key: `page:${ route }`,
```
andrzejewsky, line 115: you already has access to route in the driver so that field is useless here - decide in the driver what and how you want to store pages

andrzejewsky (changes requested):

packages/core/cache/nuxt/index.js
filipsobol: Added

packages/core/cache/nuxt/index.js
```diff
@@ -1,93 +1,132 @@
...
+        return typeof invalidator === 'string'
+          ? requirePackage(invalidator)
+          : invalidator;
+      });
+
+      // Get tags from all invalidators
+      const tags = invalidators.reduce((tags, invalidator) => {
+        const newTags = invalidator({ request, response, options });
+        return tags.concat(newTags);
+      }, []);
```
Fifciu, line 37: Maybe it would be a good idea to use `Set` instead of `Array`? We would have only unique tags

packages/core/cache/nuxt/index.js
```diff
@@ -1,93 +1,132 @@
...
+    return;
+  }
 
-  return invalidatorsRaw.map(invalidator => {
-    if (typeof invalidator === 'string') {
-      return require(path.resolve(invalidator)).default;
+  const handler = async (request, response) => {
+    try {
+      // Resolve invalidator paths to packages
+      const invalidators = options.invalidators.map(invalidator => {
```
Fifciu, line 27: I think `driver.invalidate` is an invalidator. These methods just prepares list of tags to invalidate. Maybe we should use a different name to miss confusing it with `driver's invalidators`? Let me know wdyt

packages/core/cache/nuxt/index.js
```diff
@@ -1,93 +1,132 @@
...
+        render,
+        getTags
+      });
     } catch (err) {
-      Logger.error('Your cache driver thrown an error!')
-      Logger.error('Server is going to render fresh page (cacheless)')
-      Logger.error(err)
+      Logger.error('Cache driver thrown an error when fetching cache! Server will render fresh page.');
+      Logger.error(err);
       return render();
```
Fifciu, line 129: I imagine `render` is also async method. So I suggest keep the convention and either use `return await` or just `return` for both ones

Fifciu (changes requested):

packages/core/cache/nuxt/index.js
Fifciu: I see we are doing it like that in other places for tags

filipsobol: We use `Set` internally to collect tags on the page, but drivers and invalidators (renamed to `handlers`) should use arrays in my opinion. These functions are created by the developers, and `Set` is not so well known among junior developers compared to arrays.

However duplicate tags you mentioned might be a problem, so I added following to remove such elements:

```javascript
await driver.invalidate({
  request,
  response,
   tags: [...new Set(tags)] // Removes duplicates
});
```

packages/core/cache/nuxt/index.js
filipsobol: I renamed them to `handlers` so in config looks like this:

```javascript
['@vue-storefront/cache/nuxt', {
  invalidation: {
    endpoint: '/cache-invalidate',
    key: 'uniqueKey',
    handlers: []
  },
  driver: []
}]
```

packages/core/cache/nuxt/index.js
filipsobol: `await` before `driver.invoke` is required to catch errors. However, it's not required before `render`, because we can assume that it will work, and Promise returned from `async` function will still be awaited, so it doesn't give any benefit (WebStorm even show suggestions to remove `await`s from such places).

PR #5574, opened 2021-02-24
Update faceting docs

filipsobol:


andrzejewsky (approved):

andrzejewsky:
this is good but there is one thing that's missing - configuration. Please mention that what's the default configuration, what values you can set, because based on that the user can have different behavior in an app

packages/core/docs/commercetools/use-facet.md
```diff
@@ -167,31 +167,67 @@ type ProductVariant = {
...
+    { facet: 'variants.attributes.color.key', type: 'string', option: '', name: 'color' }
+  ],
+  sortingOptions: [
+    { id: 'latest', name: 'Latest', facet: 'createdAt', direction: 'desc' },
+    { id: 'price-up', name: 'Price from low to high', facet: 'price', direction: 'asc' },
+    { id: 'price-down', name: 'Price from high to low', facet: 'price', direction: 'desc' },
+    { id: 'relevance', name: 'Relevance', facet: 'score', direction: 'desc' },
+  ],
+  filteringStrategy: 'filter'
+}
```
andrzejewsky, line 200: this tells me nothing about those options, I don't know what they are built for. Please mention the following things:
- `pageOptions` - available options of amount of possible elements on the page
- `subcategoriesLimit` - don't remember that... :D 
- `availableFacets` - that are the available filters bu using CT facets, add ref to:https://docs.commercetools.com/api/projects/products-search#facets `option` is the option of aggregation (eg. subtree) and `name` means alias (please also point that name 'category' is constant, user shouldn't change it)
-  `sortingOptions` - similar to previous option: https://docs.commercetools.com/api/projects/products-search#sorting
- `filteringStrategy` - https://docs.commercetools.com/api/projects/products-search#filters possible options here: filter, query and facets (or they are named differently in the rest API sdk)

PR #5581, opened 2021-02-25
Fix "dev" mode in CT api-client

filipsobol:


Fifciu (approved):
It looks pretty fine for me. However, I wonder, why did you get rid of `pkg.server`? Wouldn't it be nice to have all knowledge about built files in package.json?

filrak (approved):

filipsobol:
> It looks pretty fine for me. However, I wonder, why did you get rid of `pkg.server`? Wouldn't it be nice to have all knowledge about built files in package.json?

This is a made-up field that doesn't normally exist in `package.json` unlike `main`, `module`, and `types`. It was only needed to provide an entry path for Rollup, so I decided to remove it. Let me know if you think it should stay.

packages/commercetools/api-client/rollup.config.js
```diff
@@ -1,17 +1,24 @@
...
+const main = generateBaseConfig(pkg);
+main.plugins.push(graphql());
 
-export default {
+const server = {
   input: 'src/index.server.ts',
   output: [
     {
-      file: pkg.server,
+      file: 'server/index.js',
```
andrzejewsky, line 13: why?

filipsobol: Reverted

andrzejewsky (approved):

PR #5600, opened 2021-03-02
Update useFacet docs

filipsobol:


packages/core/docs/commercetools/use-facet.md
```diff
@@ -186,7 +186,7 @@ If the explicit configuration is not provided, the following defaults will be us
   ],
   subcategoriesLimit: 100,
   availableFacets: [
-    { facet: 'categories.id', type: 'string', option: 'subtree("*")', name: 'category' },
+    { facet: 'categories.id', type: 'string', option: 'subtree("*")', name: 'category' }, // Don't change the "name" of this facet
```
Fifciu, line 189: Linter yields all the time on me to have comments - line before the line. Let's apply it there too

packages/core/docs/commercetools/use-facet.md
```diff
@@ -200,7 +200,21 @@ If the explicit configuration is not provided, the following defaults will be us
 }
 ```
 
-Configuration can be modified by passing identical configuration to:
+- `pageOptions` - array containing maximum number of elements displayed per page.
```
Fifciu, line 203: `an array`. In line `205` you used `an` - so let's unify it

Fifciu (changes requested):

packages/core/docs/commercetools/use-facet.md
filipsobol: This comment is used to grab users' attention, not to be kept in code. I think that having it on the same line as the code it refers to makes it slightly better in this case.

filrak (approved):

PR #5617, opened 2021-03-04
Update integration guide

filipsobol:


packages/core/docs/integrate/integration-guide.md
```diff
@@ -400,10 +395,14 @@ You need to create a couple of Vue components:
 |----------------------------------------------|-------------------------------------|-------------|
```
Fifciu, line 395: `You need to create a couple of Vue components` - not each one in this table is a Vue component. Let's rephrase this

Fifciu (changes requested):

PR #5653, opened 2021-03-12
Bug/fix register login problem

### LINKED ISSUE #5613: Register function from CT useUser composable allows user to log in.
## Current behavior
During the registration when the user provides an email address that already exists the API returns an error related to an already existing email and then tries to reach the token auth endpoint. As a result, a bad error message is returned and if the user provides a correct password for the existing email address the registration form allows them to log in.

## Expected behavior
After providing the already existing email address the composable should return the correct error and stop any further actions.

## Version of Vue Storefront

* [ ] Vue Storefront
* [x] Vue Storefront Next

* [ ] YES
* [x] NO

Baroshem:
### Related Issues

closes #5613 

Fix problem with register wrong error handling. Previously, after providing email that already exists in register form, there was no error indicating that. The only error we were getting was error from token generation service that credentials do not match. Now, if user provides email that is already registered during registration, the service will stop after validating email and will show error to user.

### Pull Request Checklist

### DIFF
--- a/packages/commercetools/api-client/src/helpers/commercetoolsLink/linkHandlers.ts
+++ b/packages/commercetools/api-client/src/helpers/commercetoolsLink/linkHandlers.ts
@@ -20,16 +20,20 @@ export const handleBeforeAuth = async ({ sdkAuth, tokenProvider, apolloReq, curr
   return tokenProvider.getTokenInfo();
 };
 
-export const handleAfterAuth = async ({ sdkAuth, tokenProvider, apolloReq, currentToken }) => {
+export const handleAfterAuth = async ({ sdkAuth, tokenProvider, apolloReq, currentToken, response }) => {
   if (!isUserSession(currentToken) && isUserOperation(apolloReq.operationName)) {
     const { email, password } = apolloReq.variables.draft;
     Logger.debug('Apollo authLinkAfter, customerPasswordFlow', apolloReq.operationName);
 
-    const token = await sdkAuth.customerPasswordFlow({ username: email, password });
-    tokenProvider.setTokenInfo(token);
-    Logger.debug('Apollo authLinkAfter, customerPasswordFlow, generated token: ', getAccessToken(token));
+    if (!response.errors.length) {
+      const token = await sdkAuth.customerPasswordFlow({ username: email, password });
+      tokenProvider.setTokenInfo(token);
+      Logger.debug('Apollo authLinkAfter, customerPasswordFlow, generated token: ', getAccessToken(token));
 
-    return token;
+      return token;
+    }
+
+    return currentToken;
   }
 
   return currentToken;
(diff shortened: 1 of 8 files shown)


packages/core/core/src/factories/useUserFactory.ts
```diff
@@ -44,13 +48,14 @@ export const useUserFactory = <USER, UPDATE_USER_PARAMS, REGISTER_USER_PARAMS ex
...
       Logger.debug('useUserFactory.register', providedUser);
+      resetErrorValue();
 
       try {
         loading.value = true;
         error.value.register = null;
         user.value = await factoryParams.register(context, providedUser);
       } catch (err) {
-        error.value.register = err;
+        error.value.register = err.response?.data || err;
```
andrzejewsky, line 58: please assign only `err` you can't assume there is always data field, each integration can be different and this is core (agnostic)

andrzejewsky (changes requested):

packages/core/core/src/factories/useUserFactory.ts
andrzejewsky: hmm use that mapping here: https://github.com/vuestorefront/vue-storefront/blob/release-2-2/packages/commercetools/composables/src/useUser/factoryParams.ts#L41

wrap authenticate in try-catch, and in the catch statement throw the error again but with mapping err.response?.data.errors (I guess errors are there)

Baroshem: I found out that `err.response?.data` wasn't needed.

andrzejewsky (approved):

filrak (approved):

packages/commercetools/api-client/src/helpers/commercetoolsLink/linkHandlers.ts
```diff
@@ -20,16 +20,20 @@ export const handleBeforeAuth = async ({ sdkAuth, tokenProvider, apolloReq, curr
...
-export const handleAfterAuth = async ({ sdkAuth, tokenProvider, apolloReq, currentToken }) => {
+export const handleAfterAuth = async ({ sdkAuth, tokenProvider, apolloReq, currentToken, response }) => {
   if (!isUserSession(currentToken) && isUserOperation(apolloReq.operationName)) {
     const { email, password } = apolloReq.variables.draft;
     Logger.debug('Apollo authLinkAfter, customerPasswordFlow', apolloReq.operationName);
 
-    const token = await sdkAuth.customerPasswordFlow({ username: email, password });
-    tokenProvider.setTokenInfo(token);
-    Logger.debug('Apollo authLinkAfter, customerPasswordFlow, generated token: ', getAccessToken(token));
+    if (!response.errors.length) {
```
krskibin, line 28: This line causing an error, and because of this go through the auth process is not possible.

krskibin:
Hey!
I added a comment to the line that causes an error. Because of that error, in version 1.1.6, there's no possibility to register a user.

PR #5661, opened 2021-03-16
Add Bazaarvoice docs

filipsobol:
Add Bazaarvoice docs

### DIFF
--- a/packages/core/docs/integrations/bazaarvoice.md
+++ b/packages/core/docs/integrations/bazaarvoice.md
@@ -1,0 +1,129 @@
+# Bazaarvoice <Badge text="Enterprise" type="info" />
+
+::: warning Paid feature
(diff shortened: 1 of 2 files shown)


filrak (approved):

packages/core/docs/integrations/bazaarvoice.md
```diff
@@ -0,0 +1,129 @@
...
+        clientName: '',
+        siteId: '',
+        environment: '',
+        locale: ''
+      }
+    ],
+  ],
+  build: {
+    transpile: [
+      '@vsf-enterprise/bazaarvoice'
```
andrzejewsky, line 42: why not in rawSources?

andrzejewsky (approved):

packages/core/docs/integrations/bazaarvoice.md
filipsobol: In this case, it doesn't make a difference, but I changed it to be consistent with other docs

PR #5705, opened 2021-03-24
Restore and update useFacet docs

filipsobol:
This PR restores and updates `useFacet` docs that were added 3 weeks ago, but disappeared from the docs https://github.com/vuestorefront/vue-storefront/pull/5600/files

### DIFF
--- a/packages/core/docs/commercetools/composables/use-facet.md
+++ b/packages/core/docs/commercetools/composables/use-facet.md
@@ -170,7 +170,7 @@ type ProductVariant = {
 ## Configuration
 
 ::: warning
-Configuration can be changed only for th Enterprise version of this package.
+Configuration can be changed only for the Enterprise version of this package.
 :::
 
 Faceting configuration can be modified to change available sorting options, filters, etc.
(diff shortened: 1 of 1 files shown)


packages/core/docs/commercetools/composables/use-facet.md
```diff
@@ -200,10 +196,48 @@ If the explicit configuration is not provided, the following defaults will be us
...
+  ]
+};
+
+// middleware.config.js
+module.exports = {
+  integrations: {
+    ctf: {
+      location: '@vsf-enterprise/ct-faceting/server',
+      configuration: {
+        // options
```
Fifciu, line 234: What options? Shouldn't we provide an example with mocked values? It would be much simpler for developers, even for as after a break from this part :D

Fifciu (changes requested):

packages/core/docs/commercetools/composables/use-facet.md
filipsobol: I changed the order of code examples in this section. I show places where the configuration should be put, but still without any examples because by default no additional configuration is required.
Right after that, I show the defaults used when no configuration is provided, so the developer can only take pieces they want to modify.

filrak (approved):

Fifciu (changes requested):
`configuration` part should not be empty

Fifciu (approved):

PR #5710, opened 2021-03-25
[WIP] Docs polishments

filrak:
Polishments for docs.

- Key Concepts: 
  - Added new sections (integrations, extensibility, backend-agnostic)
  - Updated configuration part
- Composables
  - Added information about correlation of composables and components for shipping/billing providers
- Enterprise
  -Gave more information about what it is and what are the differences between this and OS version

TBC

### DIFF
--- a/packages/core/docs/general/key-concepts.md
+++ b/packages/core/docs/general/key-concepts.md
@@ -4,13 +4,10 @@ This document will walk you through the most important concepts of Vue Storefron
 
 ## Configuration
 
-The first thing you usually want to do after setting up a new project is some configuration. No matter if you want to change your backend API credentials, change routes or add a custom logger, there is always a single place to do all these things - Vue Storefront Modules configuration in `nuxt.config.js`.
+There are two types of configuration in Vue Storefront:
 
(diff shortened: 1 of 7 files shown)


packages/core/docs/general/key-concepts.md
```diff
@@ -39,26 +36,86 @@ By default, we're using `nuxt-i18n` module for internationalization.
...
+You have access to all methods through `api` field of each integration
 
-## Integrations
+```js
+const products = $ct.api.getProducts(params)
+```
+
+## Backend-agnostic
+
+No matter what backend services you're using they are handled more or less the same way on the frontend. No matter what eCommerce platform, CMS or Search you're using you will always use the same getters and composables so it's easy to work with VSF projects on different tech stacks or try new services without making heavy investments.
```
NataliaTepluhina, line 103:
```suggestion
Whatever backend services you're using, they are handled more or less the same way on the frontend. No matter what eCommerce platform, CMS, or Search you're using you will always use the same getters and composables so it's easy to work with VSF projects on different tech stacks or try new services without making heavy investments.
```

NataliaTepluhina (commented):
@filrak one small note

PR #5714, opened 2021-03-29
Add extendApp for middleware extensions, add reloading functionality for cart

andrzejewsky:
adds: `extendApp` for middleware extensions

### DIFF
--- a/packages/core/core/src/factories/useCartFactory.ts
+++ b/packages/core/core/src/factories/useCartFactory.ts
@@ -39,10 +39,26 @@ export const useCartFactory = <CART, CART_ITEM, PRODUCT, COUPON>(
       removeItem: null,
       updateItemQty: null,
       load: null,
+      refresh: null,
       clear: null,
       applyCoupon: null,
       removeCoupon: null
     }, 'useCart-error');
+
+    const _refresh = (reportName) => async ({ customQuery } = { customQuery: undefined }) => {
+      Logger.debug(`useCart.${reportName}`);
+
+      try {
+        loading.value = true;
+        cart.value = await _factoryParams.load({ customQuery });
+        error.value[reportName] = null;
+      } catch (err) {
+        error.value[reportName] = err;
+        Logger.error(`useCart/${reportName}`, err);
+      } finally {
+        loading.value = false;
+      }
+    };
 
     const setCart = (newCart: CART) => {
       cart.value = newCart;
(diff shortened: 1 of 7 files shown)


packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -113,6 +113,21 @@ export const useCartFactory = <CART, CART_ITEM, PRODUCT, COUPON>(
...
 
+    const refresh = async ({ customQuery } = { customQuery: undefined }) => {
+      Logger.debug('useCart.refresh');
+
+      try {
+        loading.value = true;
+        cart.value = await _factoryParams.load({ customQuery });
+        error.value.load = null;
+      } catch (err) {
+        error.value.load = err;
```
Fifciu, line 124: `error.value.refresh` should be supported imo

Fifciu: If you decide to add it, do not forget to add it at the very top of the file inside ` const error: Ref<UseCartErrors> = sharedRef({` object

packages/core/docs/advanced/server-middleware.md
```diff
@@ -75,6 +75,7 @@ const extension = {
   extendApiMethods: {
     getProduct: async () => { /* ... */ }
   },
+  extendApp: (app) => {  /* ... */ },
```
Fifciu, line 78: Why extension can extend the app but integration cannot? I imagine we will end up having many 1+1 integrations (1 integration with 1 extension - because we want to extend express' app)

packages/core/core/src/factories/useCartFactory.ts
```diff
@@ -113,6 +113,21 @@ export const useCartFactory = <CART, CART_ITEM, PRODUCT, COUPON>(
...
+        Logger.error('useCart/load', err);
```
Fifciu, line 125: `useCart/load` or `refresh`?

Fifciu (changes requested):

packages/core/core/src/factories/useCartFactory.ts
andrzejewsky: please notice that load calls refresh - the difference is only in the side effects - load can load the cart only once, while refresh loads it always

packages/core/docs/advanced/server-middleware.md
andrzejewsky: _but integration cannot?_ - yes they can, extensions are defined in the middleware config.

packages/core/core/src/factories/useCartFactory.ts
Fifciu: So we want to have `error.value.refresh` not `.load`, wdyt?

filrak (approved):

PR #5774, opened 2021-04-16
Pass integration configuration to "extendApp" and fix CT cookies

filipsobol:
* [BREAKING] Pass object containing `app` and `configuration` to `extendApp` instead of `app` object only.
* Adds  `expires` to CT cookie to the same date as the token, so the cookie is stored for as long as the token is valid.

### Pull Request Checklist
- [X] I have tested my Pull Request on production build and (to my knowledge) it works without any issues


Baroshem (commented):
I tried different approach and it did not work for me. If that works for you, then you have my approval ;)

filipsobol:
> I tried different approach and it did not work for me. If that works for you, then you have my approval ;)

It does. The problem is that the cookie is set not only by `onTokenChange`, but also `onTokenRead`.

packages/commercetools/api-client/src/index.server.ts
```diff
@@ -70,13 +70,23 @@ const tokenExtension: ApiClientExtension = {
         auth: {
           onTokenChange: (newToken) => {
             if (!currentToken || currentToken.access_token !== newToken.access_token) {
-              res.cookie('vsf-commercetools-token', JSON.stringify(newToken));
+              res.cookie(
+                'vsf-commercetools-token',
+                JSON.stringify(newToken),
+                { expires: newToken?.expires_at ? new Date(newToken.expires_at) : new Date() }
```
Fifciu, line 76: Maybe?:
```ts
{ expires: new Date(newToken?.expires_at || null) }
```

filipsobol: I pushed a fix. Setting date to the past (which will be a case if we do `new Date(null)` removes a cookie, which is not intended behavior.

Fifciu (approved):

Baroshem (approved):
Well done!

PR #5791, opened 2021-04-21
EN-164: 2.3 docs update

vn-vlad:
Task: https://vsf.atlassian.net/browse/EN-164


packages/core/docs/integrate/integration-guide.md
```diff
@@ -189,6 +162,45 @@ const getProduct = async (context, params) => {
...
+          variables.locale = 'en'
+          return { query, variables }
+        }
+      }
+    }
+  }
+};
+```
+
+The custom query modifier function always has in the arguments the default query and default variables and must return the query and its variables as well. In the body you can do anything you want with those parameters - you can override them or even change to the new ones. After creating the modifier function, you can use `extendQuery` to change the default query from middleware api method by providing `customQuery` object parameter that contains the query name as key and the identifier of the modifier function as value form client-side api method call.
```
filrak, line 186: would be great to mention full flow (so also composables part)

filrak (approved):
Looks good to me, good job! Let's wait for @andrzejewsky to take a look ;)

rglepper:
Hi, from someone that's new to creating an integration and is reading this documentation. I'd like to say that it would really help if it wasn't written in isolation from what's in the boilerplate. Just a tiny example: if using the boilerplate, you don't need to _create_ index.server.ts because it's already there. It would be better to say that it's in the boilerplate and make sure the snippet in the example is the same code. Hope this makes sense.

Cheers

filrak:
Hey @rglepper, perhaps you'd like to improve it and add/change things that you should be done differently? I think your perspective will be super valuable here as you're the integrator ;)

filrak:
@vn-vlad can you propose this PR to `release/next`? this way it will get deployed immediately (and it should because we have outdated tutorial)

PR #5800, opened 2021-04-23
ci: add labeler config

Baroshem:
This PR adds labeler workflow for VSF 2 (Next) to automatically assing labels based on modified files.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,16 +1,12 @@
-# Add 'VSF1' label to any change to packages files
+# Add 'VSF1' label to any change to this branch
 VSF1:
 - true
 
-# Add 'core' label to any change to packages/core files within the source dir EXCEPT for the docs sub-folder
+# Add 'core' label to any change to core files
 core:
 - core/**/*
 
-# Add 'commercetools' label to any change to packages/commercetools files
-commercetools:
-- packages/commercetools/**/*
-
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,39 @@
+# Add 'VSF1' label to any change to packages files
+VSF1:
+- true
+
+# Add 'core' label to any change to packages/core files within the source dir EXCEPT for the docs sub-folder
```
Fifciu, line 5: `packages/core` -> `core`

.github/labeler.yml
```diff
@@ -0,0 +1,39 @@
...
+core:
+- core/**/*
+
+# Add 'commercetools' label to any change to packages/commercetools files
+commercetools:
```
Fifciu, line 10: It does not exist in vsf1

Fifciu (changes requested):
Awesome one. Just one thing to update. It would be awesome if you could also prepare similar PR for hotfix 1.12.3 and develop

.github/labeler.yml
Baroshem: You are right. Changed it. Thanks!

.github/labeler.yml
Baroshem: Removed

bloodf:

PR #5811, opened 2021-04-23
update docs

vn-vlad:
Task: https://vsf.atlassian.net/browse/EN-164

### DIFF
--- a/packages/core/docs/integrate/integration-guide.md
+++ b/packages/core/docs/integrate/integration-guide.md
@@ -101,7 +101,7 @@ Each integration starts with `api-client`. This is one of the packages which is 
 
 Our API client always shares two entry points:
 
-- `@vue-storefront/{INTEGRATION}/server` - shares the `createApiClient` and `integrationPlugin` for direct connection
+- `@vue-storefront/{INTEGRATION}/server` - shares the `createApiClient`.
 - `@vue-storefront/{INTEGRATION}` - shares other library code, such as helpers, types etc.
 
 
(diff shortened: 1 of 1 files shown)


packages/core/docs/integrate/integration-guide.md
```diff
@@ -151,7 +126,6 @@ const onCreate = (settings) => {
 
 
 const { createApiClient, integrationPlugin } = apiClientFactory({
```
filrak, line 128: https://github.com/vuestorefront/ecommerce-integration-boilerplate/blob/master/packages/api-client/src/index.server.ts#L15 no integration plugin, please check integration plugin's code as well

filrak (changes requested):

andrzejewsky (approved):

andrzejewsky:
A few more things that need to be updated

- example of plugin:

```js
import { integrationPlugin } from '@vue-storefront/core'

const moduleOptions = <%= serialize(options) %>;

export default integrationPlugin(({ app, integration }) => {
  const settings = { api: '/graphql', user: 'root' }

  integration.configure({ ...moduleOptions, ...settings })
});
```

- Extending an existing integration
just remove that section

- setup in factory params
please change it to `provide`

- Creating composables without VSF factories
- 
```js
import { vsfRef, useVSFContext } from '@vue-storefront/core';
const useCart = () => {
  const cart = vsfRef(null, 'my-own-cart')
  const context = useVSFContext();
  const addToCart = async ({ product }) => {
    return context.$ownAPI.updateCart(product)
  }
  return { addToCart, cart }
}
```

andrzejewsky (approved):

PR #5823, opened 2021-04-26
support for async hooks added

vn-vlad:
### Related Issues
https://vsf.atlassian.net/browse/OS-64

Hooks in the middleware are not async, which can be an issue when someone would like to trigger some async action there.
Added support for async call for `beforeCreate`, `afterCreate`, `beforeCall` and `afterCall` hooks.

### DIFF
--- a/packages/core/core/src/factories/apiClientFactory/index.ts
+++ b/packages/core/core/src/factories/apiClientFactory/index.ts
@@ -3,41 +3,68 @@ import {
   ApiClientConfig,
   ApiInstance,
   ApiClientFactory,
-  ApiClientExtension
+  ApiClientExtension,
+  ApiClientExtensionHooks
 } from './../../types';
 import { Logger } from './../../utils';
 import { applyContextToApi } from './context';
 
-const isFn = (x) => typeof x === 'function';
+type Fn<A extends any [] = any [], B = any> = (...xs: A) => B;
+type HooksNames = keyof ApiClientExtensionHooks;
+
+const isFn = (x): x is Fn => typeof x === 'function';
+
+const callWithProp = (prop: string, rest: any = {}) => (result: any, hook: any) =>
+  isFn(hook)
+    ? hook({...rest, [prop]: result})
+    : result;
+
+const callThenWithProp = (prop: string, rest: any = {}) => (result: any, hook: any) =>
+  isFn(hook)
+    ? result.then(result => hook({...rest, [prop]: result}))
+    : result;
+
+const handleHook = (hookName: HooksNames, call: Fn) => (result: any, next: ApiClientExtensionHooks) =>
+  call(result, next[hookName]);
 
 const apiClientFactory = <ALL_SETTINGS extends ApiClientConfig, ALL_FUNCTIONS>(factoryParams: ApiClientFactoryParams<ALL_SETTINGS, ALL_FUNCTIONS>): ApiClientFactory => {
   function createApiClient (config: any, customApi: any = {}): ApiInstance {
     const rawExtensions: ApiClientExtension[] = this?.middleware?.extensions || [];
+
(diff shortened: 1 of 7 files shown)


packages/core/core/src/factories/apiClientFactory/index.ts
```diff
@@ -11,39 +11,38 @@ import { applyContextToApi } from './context';
 const isFn = (x) => typeof x === 'function';
```
andrzejewsky, line 12: about this file:
- Please keep before/after create as it is (not async) - I know that is the opposite as it was in the AC but it doesn't make sense to have before/after configure async. In that way, createApiClient won't be async as well.
- beforeCall/After call just prefix async is enough:

```    const extensionHooks = {
      before: async (params) => lifecycles
        .filter(e => isFn(e.beforeCall))
        .reduce((args, e) => e.beforeCall({ ...params, configuration: settings.config, args}), params.args),
      after: async (params) => lifecycles
        .filter(e => isFn(e.afterCall))
        .reduce((response, e) => e.afterCall({ ...params, configuration: settings.config, response }), params.response)
    };
```

packages/core/core/src/types.ts
```diff
@@ -652,10 +652,10 @@ export interface AfterCallParams<C> extends CallHookParams<C> {
...
 
 export interface ApiClientExtensionHooks<C = any> {
-  beforeCreate?: (params: HookParams<C>) => C;
-  afterCreate?: (params: HookParams<C>) => C;
-  beforeCall?: (params: BeforeCallParams<C>) => BeforeCallArgs;
-  afterCall?: (params: AfterCallParams<C>) => AfterCallArgs;
+  beforeCreate?: (params: HookParams<C>) => Promise<C> | C;
+  afterCreate?: (params: HookParams<C>) => Promise<C> | C;
+  beforeCall?: (params: BeforeCallParams<C>) => Promise<BeforeCallArgs> | BeforeCallArgs;
+  afterCall?: (params: AfterCallParams<C>) => Promise<AfterCallArgs> | AfterCallArgs;
```
andrzejewsky, line 658: always Promise

andrzejewsky (changes requested):

packages/core/core/src/factories/apiClientFactory/index.ts
```diff
@@ -21,29 +21,28 @@ const apiClientFactory = <ALL_SETTINGS extends ApiClientConfig, ALL_FUNCTIONS>(f
...
-    };
+      .reduce((configuration, curr) => curr.afterCreate({ configuration }), settings.config);
+
+    const before = async (params) => lifecycles
+      .filter(e => isFn(e.beforeCall))
+      .reduce((args, e) => args.then(args => e.beforeCall({ ...params, configuration: settings.config, args})), Promise.resolve(params.args));
+
+    const after = async (params) => lifecycles
+      .filter(e => isFn(e.afterCall))
+      .reduce((response, e) => response.then(response => e.afterCall({ ...params, configuration: settings.config, response })), Promise.resolve(params.response));
```
andrzejewsky, line 40: please create a functions that will wrap the repetitive things:

```js
const handlePromise = (params, hookName, fieldName) => result =>
   e.[hookName]({ ...params, configuration: settings.config, [fieldName]: result  })

const createPromises =  (params, hookName, fieldName) =>  (hook, e) =>
   hook.then(handlePromise(params, hookName, fieldName)), Promise.resolve(params[fieldName]));

```

then use as:

```js
.reduce(createPromises('beforeCall', 'args'));

.reduce(createPromises('afterCall', 'response'));
```

isn't simpler?

andrzejewsky (changes requested):

andrzejewsky (approved):

andrzejewsky (approved):

PR #5854, opened 2021-04-28
Add args parameter to custom queries

andrzejewsky:
closes: https://github.com/vuestorefront/vue-storefront/issues/5855

### DIFF
--- a/packages/core/docs/guide/composables.md
+++ b/packages/core/docs/guide/composables.md
@@ -430,7 +430,7 @@ module.exports = {
         'my-products-query': ({ query, variables, metadata }) => {
 
           variables.locale = 'en'
-          variables.where = `masterData(variants(size === ${metadata.size}))`
+          variables.size = metadata.size
 
           return { query, variables }
         }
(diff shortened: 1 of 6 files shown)


packages/core/docs/guide/composables.md
```diff
@@ -434,4 +434,4 @@ module.exports = {
 };
 ```
 
-The custom query function always has in the arguments the default query and default variables and must return the query and its variables as well. In the body you can do anything you want with those parameters - you can override them or even change to the new ones.
+The custom query function always has in the arguments the default query (`query`), default variables (`variables`) and additional parameters (`args`) sent from the front-end. This function always must return the query and its variables as well, while in the body you can do anything you want with those parameters - you can override them or even change to the new ones.
```
filrak, line 437: can you add some example with a frontend call and then receiving this in query?

filrak (changes requested):

packages/core/docs/guide/composables.md
```diff
@@ -422,9 +427,10 @@ module.exports = {
       location: '@vue-storefront/commercetools-api/server',
       configuration: { /* ... */ },
       customQueries: {
-        'my-products-query': ({ query, variables }) => {
+        'my-products-query': ({ query, variables, metadata }) => {
```
filrak, line 430: make something more generic

filrak: `metaData` ?

filrak (approved):

filipsobol:
Docs look fine :thumbsup:

PR #5860, opened 2021-04-30
fix: update installation guide

lukaszjedrasik:

Updated installation guide


Fifciu (changes requested):
Looks great. However, we need it also in `production-setup.md`

Fifciu (approved):

Issue #5814, opened 2021-04-23
[RFC] Accessing the platform API from the composables


andrzejewsky:
# Summary
[summary]: #summary

Accessing the platform API from the composables

# Motivation
[motivation]: #motivation

We are providing composables for each integration (that are created from core factories), but only for common scenarios that can be mostly the same for each eCommerce platform. However, there are always something very specific for certain platform that cannot be implemented as agnostic feature very easily, without any lose in usability (weird naming or behavior) or performance (many calls at the same time because dictated flow from core).

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

For such a feature we always say that you can access api-client directly and use original commerce API right away - but it means that you need to resign from composables and that is not the best DX.

We'd like to extend the creation of composables by additional features - platform APIs. In this concept, each composable, besides implementing the common (agnostic) flows, will also share possible functions that are specific for a given integration.

Let's consider a possible use-case: The commercetools has a feature called custom line item, that allows you to add any item in the cart that's not "tracked" by the system. As you can see this feature is quite specific to the platform and it's hard to implement that in an agnostic way (for instance other platforms will not have that feature). So how we can access that feature from composable?

Here is the example:

```js
const { addToCart, cart, api } = useCart()

addToCart() // this is common scenario, agnostic api

api.addLineItem() // this is platform-specific scenario, you can add line item, commercetools feature

cart.value // both can influence the result of cart
```

How it works:

* composables besides the common scenarios and core-defined fields will share an additional field called `api`
* the field `api` contains functions that are platform-specific
* those functions can change the response fields (cart in this case) as the other functions
* the integrator can implement as many functions like this as he want

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

### Integrators API:
```js
const factoryParams = {
  addToCart: async (context, params) { /* ... */ }
  // other, common functions as always
  api: {
     // platform specific ones
     addLineCart: async (context, params) { 
        const lineItems = await context.$ct.api.addLineItem(params.item)

        return { ...params.currentCart, lineItems }
      }
  }
}

const useCart = useCartFactory(factoryParams)
```

_Note_:  _those functions are automatically wrapped by loading and error state (try-catch)_

# API-client vs platform API from composables

- reactivity
- functions are related to the composable purpose
- all of the function are available on-demand without any additional imports
...

# Drawbacks
[drawbacks]: #drawbacks

[...]

PR #5901, opened 2021-05-12
Platofrm api access

andrzejewsky:
closes: https://github.com/vuestorefront/vue-storefront/issues/5814


packages/core/docs/guide/composables.md
```diff
@@ -218,13 +218,16 @@ Every Vue Storefront composable usually returns three main pieces:
 - **Main data object**. A read-only object that represents data returned by the composable. For example, in `useProduct` it's a `products` object, in `useCategory` it's `category` etc.
 - **Supportive data objects**. These properties depend directly on indirectly on the main data object. For example, `loading`, `error` or `isAuthenticated` from `useUser` depend on a `user`.
 - **Main function that interacts with data object**. This function usually calls the API and updates the main data object. For example in `useProduct` and `useCategory` it's a `search` method,in `useCart` it's a `load` method. The rule of thumb here is to use `search` when you need to pass some search parameters. `load` is usually called when you need to load some content based on cookies or `localStorage`
+- **platform-specyfic API access**. This is the section where you can reach out to the API functions that are specific to the platform you are using. By default we provide an agnostic approach of using the same API and function calls for each platform, however sometimes there is a need to use something very specific to the certain service, thus you can easily access it over the `api` object.
```
filrak, line 221:
```suggestion
- **platform-specific API access**. This is the section where you can reach out to the API functions that are specific to the platform you are using. By default we provide an agnostic approach of using the same API and function calls for each platform, however sometimes there is a need to use something very specific to the certain service, thus you can easily access it over the `api` object.
```

filrak (commented):
search & repalce `specyfic` to `specific`

filrak (approved):

PR #5904, opened 2021-05-13
Fix reactivity of "error" objects in composables

filipsobol:
This PR fixes the reactivity of "error" objects. Currently, when an error occurs we only update the key inside of the "error" object. This doesn't work well with Vue.js devtools and the `watch` method.

To see the updated state in devtools, it's required to select another component and select the initial one again.

`watch` only works in one specific case and `watchEffect` doesn't work at all:

```javascript
    watch(() => ({ ...error.value }), value=> console.log(value)); // Works
    watch(() => error.value, value=> console.log(value));  // Doesn't work
    watchEffect(() => console.log(error.value)); // Doesn't work
```

When this PR is merged, devtools and all 3 watchers will work as expected

### Pull Request Checklist
- [X] I have tested my Pull Request on production build and (to my knowledge) it works without any issues


filipsobol:
Closed in favor of https://github.com/vuestorefront/vue-storefront/pull/5923

PR #5914, opened 2021-05-14
support for async hooks

vn-vlad:
### Related Issues
https://vsf.atlassian.net/browse/OS-64

Hooks in the middleware are not async, which can be an issue when someone would like to trigger some async action there.
Added support for async call for `beforeCreate`, `afterCreate`, `beforeCall` and `afterCall` hooks.


packages/core/core/src/types.ts
```diff
@@ -654,8 +654,8 @@ export interface AfterCallParams<C> extends CallHookParams<C> {
 export interface ApiClientExtensionHooks<C = any> {
   beforeCreate?: (params: HookParams<C>) => C;
   afterCreate?: (params: HookParams<C>) => C;
```
filipsobol, line 656: I think it would make sense to make these two `async` as well

filipsobol (changes requested):

bloodf:

PR #5916, opened 2021-05-14
Cart subtotal

vn-vlad:
https://vsf.atlassian.net/browse/OS-101

Subtotal only displays sum of single line items
Subtotal should be calculated based both on line items prices and their quantity


filipsobol:
Duplicate of https://github.com/vuestorefront/vue-storefront/pull/5907

PR #5918, opened 2021-05-17
Update Product Catalog guide

filipsobol:


packages/core/docs/guide/product-catalog.md
```diff
@@ -1,112 +1,206 @@
...
+Use the `getFiltered` getter without the second parameter to get the list of all products.
+
+```vue
+<script>
 import { productGetters } from '{INTEGRATION}'
 
 export default {
   setup () {
-    // ...
+    // Load products as shown in the first example above
```
igorwojciechowski, line 57: Maybe we should just add this code here so that reader does not have to scroll to different sections?
Also `first example above` - kinda risky if we add some section before this one :)

packages/core/docs/guide/product-catalog.md
```diff
@@ -1,112 +1,206 @@
...
 
-  onSSR(async () => {
-    // triggering a search based on criteria available in url query eg. ?colo=red&sortBy=latest
-    await search(searchParams);
-  });
+export default {
+  setup () {
+    // Load products as shown in the first example above
+
+    const breadcrumbs = computed(() => facetGetters.getBreadcrumbs(result.value));
```
igorwojciechowski, line 182: But what is `result` here? Cuz there is no `result` in the first example.

packages/core/docs/guide/product-catalog.md
```diff
@@ -1,112 +1,206 @@
...
+    const sortBy = computed(() => facetGetters.getSortOptions(result.value));
+    const categoryTree = computed(() => facetGetters.getCategoryTree(result.value));
+    const facets = computed(() => facetGetters.getGrouped(result.value));
+    const products = computed(() => facetGetters.getProducts(result.value));
```
igorwojciechowski, line 186: If we load `products` as in the first example above, then `products` is already declared.

NataliaTepluhina (approved):

packages/core/docs/guide/product-catalog.md
filipsobol: Do you have any suggestions on how to fix it? I don't want to copy so many lines from the first example, especially with the comment about `searchParams`.

igorwojciechowski: how about...
```js{1-2,7-9,14}
import { computed } from '@vue/composition-api';
import { productGetters } from '{INTEGRATION}';

export default {
  setup () {
    const { products, search } = useProduct();
    const productList = compute(() => {
      return productGetters.getFiltered(products.value);
    }
    
    search(/*  product search params  */);

    return {
      productList
    };
  };
};
```

with `js{1-2,7-9,14}`.  It doesn't seem to be that long to me and might be more precise. I dunno...

Also, in example above `import { computed }` might be worth adding :)

packages/core/docs/guide/product-catalog.md
```diff
@@ -1,112 +1,245 @@
...
-You can extract certain information from `result` object using `facetGetters`:
+    const attributes = {
+      size: '38',
+      color: 'gray'
+    };
 
-```ts
-setup (props, context) {
-  const { result, search, loading } = useFacet();
+    const selectedVariant = computed(() => {
```
igorwojciechowski, line 145: `computed` import is missing :)

packages/core/docs/guide/product-catalog.md
```diff
@@ -1,112 +1,245 @@
...
+    onSSR(async () => {
+      /**
+       * "searchParams" may vary depending on the integration used.
+       * Please refer to its documentation for more details.
+       */
+      await search(searchParams);
+    })
 
-The same way we've used `getFiltered` to select the Master Variant we can use it to select other variants.
+    const masterVariant = computed(() => {
```
igorwojciechowski, line 103: `computed` import is missing :)

packages/core/docs/guide/product-catalog.md
```diff
@@ -1,112 +1,245 @@
...
-    const list = computed(() => productGetters.getFiltered(products.value));
+    onSSR(async () => {
+      /**
+       * "searchParams" may vary depending on the integration used.
+       * Please refer to its documentation for more details.
+       */
+      await search(searchParams);
+    })
+
+    const productList = computed(() => {
```
igorwojciechowski, line 68: `computed` import is missing :)

igorwojciechowski (changes requested):

igorwojciechowski (approved):

PR #5919, opened 2021-05-17
Read i18n from cookies on midddleware

andrzejewsky:
Redding locales, currencies and country from incoming cookies instead of config file in middleware.

### DIFF
--- a/packages/core/docs/commercetools/configuration.md
+++ b/packages/core/docs/commercetools/configuration.md
@@ -57,7 +57,7 @@ module.exports = {
 - `clientSecret` - Commercetools secret API key. Visit [Commercetools documentation](https://docs.commercetools.com/tutorials/getting-started#creating-an-api-client) for more details about creating an API Client
 - `scopes` - The scope constrains the endpoints to which a client has access, and whether a client has read or write access to an endpoint. Visit [Commercetools documentation](https://docs.commercetools.com/api/scopes#top) for more details about Scopes.
 
-By default the i18n settings, such as `currency`, `locale` and `country` are loaded from cookies. To force and override that behavior you can also set those properties under the `api` section.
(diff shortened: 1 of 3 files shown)


filrak (changes requested):
Shouldn't we remove `country` and `currency` from middleware config then?

Also can you update https://docs.vuestorefront.io/v2/commercetools/configuration.html#middleware-commercetools-configuration accordingly?

filipsobol:
@andrzejewsky Please update [Intternationalization doc](https://docs.vuestorefront.io/v2/advanced/internationalization.html) to reflect the changes. I can review and, if necessary, edit it.

andrzejewsky:
@filipsobol @filrak the old properties are still supported, we can't just remove them - for some users the old behavior can be the one they use. I added support i18n by cookies, but the old way still work

packages/core/docs/commercetools/configuration.md
```diff
@@ -59,6 +57,8 @@ module.exports = {
 - `clientSecret` - Commercetools secret API key. Visit [Commercetools documentation](https://docs.commercetools.com/tutorials/getting-started#creating-an-api-client) for more details about creating an API Client
 - `scopes` - The scope constrains the endpoints to which a client has access, and whether a client has read or write access to an endpoint. Visit [Commercetools documentation](https://docs.commercetools.com/api/scopes#top) for more details about Scopes.
 
+By default the i18n settings, such as `currency`, `locale` and `country` are loaded from cookies. To force and override that behavior you can also set those properties under the `api` section.
```
filipsobol, line 60: Please add an example

filipsobol, line 60:
```suggestion
By default, the internationalization settings, such as `currency`, `locale`, and `country` are loaded from cookies. To override this behavior you can set those properties inside the `api` configuration.
```

filipsobol (changes requested):

PR #5945, opened 2021-05-26
useStore

vn-vlad:
https://vsf.atlassian.net/browse/EN-258
https://vsf.atlassian.net/browse/EN-260


packages/commercetools/composables/src/getters/storeGetters.ts
```diff
@@ -0,0 +1,172 @@
...
+
+interface AgnosticStore {
+  _storeID: string;
+  _channelID: string;
+  name: string;
+  id: string;
+  description: string;
+  locales: string[];
+  address: Maybe<Address>;
+  geoLocation: Maybe<Geometry>;
```
andrzejewsky, line 39: both address and geoLocation need to be defined as Agnostic ones as well. Agnostic field means our format that represents something across all platforms, so all of data format will be transformed to that. The store, along with its data can be implemented in agnostic way as it's not really complex and can be easily mapped

packages/commercetools/composables/src/getters/storeGetters.ts
```diff
@@ -0,0 +1,172 @@
...
+
+const storeGetters: UseStoreGetters<StoreQueryResult, AgnosticStore, Maybe<Address>, Maybe<Geometry>> = {
+  getItems,
+  getName,
+  getID,
+  getDescription,
+  getLocales,
+  getAddress,
+  getLocation
+};
```
andrzejewsky, line 170: please think about getters - if in getItems you returns AgnosticStore[] - do you still need rest of them?

Getters have two purposes:
- select a subset of response
- transform platform-related data format to Agnostic* types

andrzejewsky (changes requested):
Looks really good! Two comments and also thing about:

- What about displaying a list of stores?  How do I know which one is selected?
- Perhaps I need a getter for getting a single store that is already selected? eg. `getSelected(response)`  (in load/change factory param, you will return a response as well as additional field eg. `_selected`, which will contain a selected element, so you can easily use this field in a getter)
- please look at CT documentation and adjust all queries to use selected channel (but this one can be done as next task, so please change at least one query, for instance for getting product to make sure that it works)

packages/commercetools/composables/src/getters/storeGetters.ts
andrzejewsky: please consider Address as something simple, don't implement any complex interface that handles properties like streetName etc. perhaps addresLine1 and addressLine2 will be fine? and you can add private (_address) field that keeps original format (for customizations)

filrak (commented):
Looks awesome. You figured that out very well, congrats ;)
@filipsobol will now haunt you for docs :)

packages/commercetools/composables/src/getters/storeGetters.ts
```diff
@@ -0,0 +1,172 @@
...
+
+type FilterCriteriaRecordValue <T extends Record<string, any>> =
+  FilterCriteriaRecord<T>[keyof T]
+
+type FilterCriteriaRecordField <T extends Record<string, any>> = [
+  keyof T, FilterCriteriaRecordValue<T>
+];
+
+interface AgnosticStore {
+  _storeID: string;
```
filrak, line 32: multiple channels can be associated to a single store

packages/core/core/__tests__/factories/useStoreFactory.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+describe('[CORE - factories] useStoreFactory', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('initial setup', () => {
+
+    it('should have proper initial properties', () => {
+      const useUser = useStoreFactory(factoryParams);
```
filrak, line 19: are you sure its useUser? ;)

packages/commercetools/api-client/src/fragments/line-item.ts
```diff
@@ -0,0 +1,67 @@
...
+            name(acceptLanguage: $acceptLanguage)
+            isActive
+          }
+        }
+      }
+    }
+    variant {
+      id
+      sku
+      price(currency: "USD") {
```
Fifciu, line 29: Shouldn't it be dynamic?

packages/commercetools/api-client/src/types/Api.ts
```diff
@@ -122,6 +125,7 @@ interface ApiMethods {
   updateCartQuantity ({ id, version }: CartDetails, product: LineItem): Promise<CartResponse>;
   updateShippingDetails (cart: Cart, shippingDetails: Address): Promise<CartResponse>;
   isGuest: () => boolean;
+  getStores(params: CustomQuery): Promise<StoreQueryResult>;
```
Fifciu, line 128: Maybe we should nest it in case we want to add additional params in the future?
```ts
  getStores(params: { customQuery: CustomQuery }): Promise<StoreQueryResult>;
```

packages/commercetools/api-client/src/types/setup.ts
```diff
@@ -53,8 +58,10 @@ export interface SetupConfig<TCacheShape> {
   locales?: LocaleItem[];
   languageMap?: Record<string, any>;
   acceptLanguage?: string[];
+  store: string;
   cookies?: CookiesConfig;
   auth?: Auth;
+  stores?: Stores;
```
Fifciu, line 64: It sounds a bit like `Array<Store>` for me. Maybe it would be worth refactoring this name to make sure developers won't be confused? Wdyt?

packages/commercetools/composables/nuxt/plugin.js
```diff
@@ -24,6 +25,11 @@ export default integrationPlugin(({ app, integration }) => {
     return app.$cookies.get(CT_TOKEN_COOKIE_NAME);
   };
 
+  // Stores
+  const changeCurrentStore = (id) => {
```
Fifciu, line 29: What is the type of `id`?

packages/commercetools/composables/src/getters/storeGetters.ts
```diff
@@ -0,0 +1,101 @@
...
+    addressLine1: `${address?.country ?? ''} ${address?.city ?? ''} ${address?.postalCode ?? ''}`.trim(),
+    addressLine2: `${address?.streetName ?? ''} ${address?.streetNumber ?? ''}`.trim(),
+    _address: address ?? null
+  };
+}
+
+function mapToLocale <T extends Localized>(localized: T): AgnosticLocale {
+  return {
+    code: localized.locale,
+    label: ''
```
Fifciu, line 34: Should it be an empty string?

Fifciu (changes requested):
Great work @vn-vlad 

I've put a few "for-sure questions" in the CR.

packages/core/core/src/types.ts
```diff
@@ -718,3 +741,29 @@ export type ApiClientMethods<T> = {
...
+  error: ComputedProperty<UseStoreErrors>;
+}
+
+export interface UseStore<STORES, CHANGE_PARAMS> {
+  (): UseStoreInterface<STORES, CHANGE_PARAMS>;
+}
+
+export interface UseStoreGetters<STORES, STORE_ITEM> {
+  getItems(stores: STORES, ...xs: any[]): STORE_ITEM[];
+  getSelected(stores: STORES): STORE_ITEM
```
andrzejewsky, line 768: I would suggest using AgnosticStore[] and AgnosticStore here, without generics - I know you typed also private fields (prefixed by `_`) but... they don't have to be typed, it's something for our internal usage and end developer doesn't have to be aware of it

andrzejewsky (changes requested):
Few comments, please also update changelog.

packages/commercetools/composables/src/getters/storeGetters.ts
vn-vlad: CT GraphQL `LocalizedString` type does not provide locale label value.
The default value is empty string to match the `AgnosticLocale` type.
Maybe `AgnosticLocale["label"]` property should be optional?

andrzejewsky (approved):

PR #5993, opened 2021-06-14
Update Adyen Doc

Fifciu:
Updated Adyen Doc

### DIFF
--- a/packages/core/docs/integrations/adyen.md
+++ b/packages/core/docs/integrations/adyen.md
@@ -227,7 +227,7 @@ Adyen's module isn't compatible with [Checkout.com's module](https://github.com/
 
 ### How to debug data flow?
 
-Open the `Network` tab in the browser's devtools. Each payment request will have commercetools [Payment object](https://docs.commercetools.com/api/projects/payments#payment) in the response. You can check `custom.fields` to see what data was sent to Adyen and what was the response (or error). A list of defined custom fields is [here](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/resources/web-components-payment-type.json).
+Open the `Network` tab in the browser's devtools. Each payment request will have commercetools [Payment object](https://docs.commercetools.com/api/projects/payments#payment) in the response. You can check `custom.fields` to see what data was sent to Adyen and what was the response (or error). Available custom fields are listed [here](https://github.com/commercetools/commercetools-adyen-integration/blob/master/extension/resources/web-components-payment-type.json).
 
 ### Error: NotFound: URI not found: /<project_name>/carts/<cart_id>
 `ctApi.apiHost` property inside your `middleware.config.js` contains wrong path. It should be `https://<SHOP_DOMAIN>.com/` instead of `https://<SHOP_DOMAIN>.com/<project_name>/graphql`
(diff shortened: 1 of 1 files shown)


Baroshem (approved):
I wasn't added as a reviewer but I really like content. When implementing Adyen I might probably run into similar problems so having straight answer to the question is a big 👍

Baroshem (commented):
I wasn't added as a reviewer but I really like content. When implementing Adyen I might probably run into similar problems so having straight answer to the question is a big 👍

packages/core/docs/integrations/adyen.md
```diff
@@ -224,6 +224,8 @@ If the transaction is authorized, the server's controller for `payAndOrder`/`sub
 Adyen's module isn't compatible with [Checkout.com's module](https://github.com/vuestorefront/checkout-com).
 
 ## FAQ
+### How to debug?
+Open browser's `Network` tab. Each request related to the payment will have Commercetools Payment object in the response. You could look at `custom.fields.SOME_FIELD` to discover what has been sent to the Adyen and what was the response (or error!).
```
filipsobol, line 228:
```suggestion

### How to debug data flow?

Open the` Network` tab in the browser's devtools. Each payment request will have commercetools [Payment object](https://docs.commercetools.com/api/projects/payments#payment) in the response. You can check `custom.fields.SOME_FIELD` to see what has data was sent to Adyen and what was the response (or error).

```

filipsobol: Please specify what `custom.fields.SOME_FIELD` is. This name is not used anywhere else in this document.

packages/core/docs/integrations/adyen.md
```diff
@@ -236,3 +238,11 @@ For more information, see the 6th step of the [Adyen's installation guide](./ady
 
 ### Klarna Pay Later does not work for United States
 Klarna Pay Later is not supported in the United States. However, sometimes it is added when you enable Klarna in Adyen's dashboard. If you have this problem, contact Adyen's support to remove it.
+
+### Nothing happens after 3DS2 Auth in one environment but works in a different environment
+There might be a situation when you can finish 3DS2 Auth in the local environment. But it is impossible on staging. It simply means you've provided `http://localhost:3000` as an `origin` in the `middleware.config.js` instead of your staging's URL. 
+
+It can work properly only on provided `origin`.
```
filipsobol, line 245:
```suggestion
### 3DS2 Auth doesn't work in one environment

There might be a situation when you can finish 3DS2 Auth in the local environment but not in the other, like staging. When this happens, make sure to change `origin` in the `middleware.config.js` from `http://localhost:3000` to the URL of your staging environment. 
```

packages/core/docs/integrations/adyen.md
```diff
@@ -236,3 +238,11 @@ For more information, see the 6th step of the [Adyen's installation guide](./ady
...
+
+### Structure of DetailsRequest contains the following unknown fields...
+Update version of the both extension module and notification module. Just modify tag to the newest version in `extension.Dockerfile` and `notification.Dockerfile`.
```
filipsobol, line 248:
```suggestion

Update extension and notification modules to the newest version by updating the tag in `extension.Dockerfile` and `notification.Dockerfile`.

```

filipsobol: I suggest being more precise with the work `newest` since it might be different in the future. If you think it's okay, we can leave it.

filipsobol (changes requested):

filipsobol (changes requested):

filipsobol (approved):

PR #5996, opened 2021-06-15
Docs refactor composables page

igorwojciechowski:
### Related Issues

closes #

### Pull Request Checklist

- [ ] Default Theme
- [ ] Capybara Theme


packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+    return {
+      addItem,
+      error
+    };
+  }
+};
+</script>
+```
+
+There is a dedicated TypeScript interface for every composable. Take a look at this one from `useCart`:
```
filipsobol, line 27:
```suggestion
The `@vue-storefront/core` package exports interfaces for `error` objects for every composable. Let's take a look at the one from the `useCart` composable:
```

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+  load: Error;
+  clear: Error;
+  applyCoupon: Error;
+  removeCoupon: Error;
+}
+```
+
+:::details Where does the error come from?
+
+Inside each async method, we are catching errors when they occur and save them to the reactive property called `error` under the key corresponding to the triggered method:
```
filipsobol, line 43:
```suggestion
Inside each async method, we catch thrown errors and save them in the `error` object under the key corresponding to the called method.
```

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+
+```ts
+const { addItem, error } = useCart();
+
+addItem({ product: null }); // triggers an error
```
filipsobol, line 48:
```suggestion
addItem({ product: null }); // throws an error
```

filipsobol: I would avoid using work `trigger` unless there is no alternative

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+error.value.addItem; // here you have error raised by addItem function
+```
+
+:::
+
+:::details Where can I find an interface of the error property from a certain composable?
```
filipsobol, line 54:
```suggestion
:::details Where can I find an interface of the error property for a specific composable?
```

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+
+When you are writing a code inside a script part of the Vue component, your IDE should give you hints on every different type of the composable. That's why you probably do not need to check these interfaces in the core's code.
+
+However, if you still want to check interfaces, you could find them inside [`packages/core/core/src/types.ts`](https://github.com/vuestorefront/vue-storefront/blob/next/packages/core/core/src/types.ts).
+
+[//]: # 'TODO: This should be added to API reference'
+
```
filipsobol, line 61:
```suggestion
```

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+:::
+
+### How to listen for errors?
+
+Let's imagine you have some global component for error notifications. You want to send information about every new error to this component. But how to know when a new error appears? You can observe an error object with a watcher:
```
filipsobol, line 66:
```suggestion
Let's imagine you have some global component for error notifications. You want to send information about every new error to this component. But how do you know when a new error appears?

You can observe an error object with a watcher:
```

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+    error.removeItem &&
+    error.removeItem !== prevError.removeItem
+  )
+    send({ type: 'danger', message: error.removeItem.message });
+});
+```
+
+In this example, we are using `useUiNotification` - a composable that handles notifications state. You can read more about it in the API reference.
+
+[//]: # 'TODO: This should be added to API reference'
```
filipsobol, line 87:
```suggestion
```

filipsobol (changes requested):

packages/core/docs/guide/composables.md
```diff
@@ -1,200 +1,20 @@
...
 Composable is a function that uses [Composition API](#composition-api) under the hood. Composables are the main public API of Vue Storefront and, in most cases, the only API except configuration you will work with.
 
 You can treat composables as independent micro-applications. They manage their own state, handle server-side rendering, and rarely interact with each other (except [useUser](/composables/use-user.html) and [useCart](/composables/use-cart.html)). No matter what integration you are using, your application will always have the same set of composables with the same interfaces.
 
-To use a composable, you need to import it from an integration you are using, and call it on the component `setup` option:
+To use a composable, you need to import it from an integration you use, and call it on the component `setup` option:
 
 ```js
-import { useProduct } from '{INTEGRATION}';
+import { useProduct } from '@vue-storefront/commercetools'; // or any other integration package
```
filipsobol, line 12: I suggest sticking with `{INTEGRATION}` to be consistent with the rest of the pages. `@vue-storefront/commercetools` might be harder to catch for people using different integrations.

packages/core/docs/guide/composition-api.md
```diff
@@ -0,0 +1,189 @@
...
+        console.log(this.username)
+      }
+    }
+  };
+</script>
+```
+
+Methods can be also created inside the `setup` option:
+
+```js{5-11}
```
filipsobol, line 99: I don't know if my math is right, but shouldn't it be `{6-11}`?

packages/core/docs/guide/composition-api.md
```diff
@@ -0,0 +1,189 @@
...
+    };
+
+    return { username, password, submitForm };
+  }
+};
+```
+
+Let's disable a button when username or password is empty. In order to do this, we can create a computed property - and this can be also done with the Composition API inside the `setup`:
+
+```html{5,15-17}
```
filipsobol, line 117:
```suggestion
```html{5,15-19}
```

packages/core/docs/guide/composition-api.md
igorwojciechowski: Right, thx!

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,81 @@
...
+
+:::details Where does the error come from?
+
+Inside each async method, we catch thrown errors and save them in the `error` object under the key corresponding to the called method.
+
+```ts
+const { addItem, error } = useCart();
+
+addItem({ product: null }); // throws an error
+error.value.addItem; // here you have error raised by addItem function
```
filipsobol, line 46:
```suggestion
error.value.addItem; // here you access error thrown by addItem function
```

packages/core/docs/guide/error-handling.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+:::
+
+:::details Where can I find an interface of the error property for a specific composable?
+
+When you are writing a code inside a script part of the Vue component, your IDE should give you hints on every different type of the composable. That's why you probably do not need to check these interfaces in the core's code.
+
+However, if you still want to check interfaces, you could find them inside [`packages/core/core/src/types.ts`](https://github.com/vuestorefront/vue-storefront/blob/next/packages/core/core/src/types.ts).
```
filipsobol, line 55:
```suggestion
However, if you still want to check interfaces, you could find them in [core API reference](../core/api-reference/core).
```

Please test this link

filipsobol (approved):

PR #6021, opened 2021-06-24
Reading token fix

andrzejewsky:


igorwojciechowski:
Tested locally: seems to work fine!

packages/commercetools/composables/src/useCart/index.ts
```diff
@@ -17,6 +17,7 @@ const getCurrentCartDetails = async (context: Context, currentCart): Promise<Car
 const useCartFactoryParams: UseCartFactoryParams<CartDetails, LineItem, ProductVariant, AgnosticCoupon> = {
   load: async (context: Context, { customQuery }) => {
     const { $ct } = context;
+    if (!$ct.config.auth.onTokenRead()) return null;
```
Mateuszpietrusinski, line 20: Is that object keys are always defined? What I mean, should we add `?` changing to check if the object key is available?

andrzejewsky: yes, they are provided by our nuxt plugin for comemrcetools

Mateuszpietrusinski (approved):

PR #6049, opened 2021-07-05
New `eCommerce integration`, `Creating custom composable` and `Adding customQuery support` guides

filipsobol:


packages/core/docs/integrate/integration-guide.md
```diff
@@ -5,319 +5,234 @@ The application does not reload automatically after saving the changes in Server
...
+      }
+    }
+  }
+};
+```
 
-The `factoryParams` are including functions that you have to implement to provide the functionality of implementing composable, such a fetching something from the API, or triggering other actions.
+## Implement `useProduct` functionality
 
+It's impossible to write a tutorial explaining how to implement every composable because some might differ wildly between the platforms. For this reason, we will explain how to implement `useProduct` composable, `getProduct` API endpoint, and `productGetters` and leave the rest to you.
```
igorwojciechowski, line 197:
```suggestion
It would be impossible to write a tutorial explaining how to implement every composable because some might differ wildly between the platforms. For this reason, we will explain how to implement `useProduct` composable, `getProduct` API endpoint, and `productGetters` to give as much overview as we can.
```
I would avoid saying something like "It's too hard, so do it yourself" :) At least that was my impression/interpretation.

packages/core/docs/integrate/integration-guide.md
```diff
@@ -5,319 +5,234 @@ The application does not reload automatically after saving the changes in Server
...
 
-```ts
-// composables/src/useCart/index.js
+### Understand composables
 
-import { useCartFactory, UseCartFactoryParams, Context } from '@vue-storefront/core';
+Before implementing any composable, you should get familiar with its TypeScript interfaces.
```
igorwojciechowski, line 201: AFAIK, the TypeScript knowledge requirement for creating an integration is optional. Just wanted to point that out.

packages/core/docs/integrate/integration-guide.md
```diff
@@ -340,78 +255,108 @@ const factoryParams: UseUserFactoryParams = {
...
+:::
 
-### Creating composables without VSF factories
+### Understand `api-client`
 
-Sometimes you need to omit using VSF factories. You can have a complex integration or decide to follow your way of implementing it. In this case, you can use Vue Storefront architecture and create your own composable, by creating a custom factory, or not using one at all.
+In the previous section, we added a call to the `getProduct` endpoint. Before we implement this endpoint, you should understand why we need `api-client` in the first place.
 
-However, you have to keep in mind that you need to handle context and reactive properties as well - of course, there are helpers for that.
+As mentioned in [Project structure](#project-structure) section, `api-client` is a server that acts as a proxy. All requests to and from various APIs pass through it. You might be wondering why we choose this architecture. Theoretically, calling APIs directly from the browser would result in a better performance.
```
igorwojciechowski, line 264:
```suggestion
As mentioned in [Project Structure](#project-structure) section, `api-client` is a server that acts as a proxy. All requests to and from various APIs pass through it. You might be wondering why we choose this architecture. Theoretically, calling APIs directly from the browser would result in a better performance.
```
I think that titles should be capitalized. Not 100% sure tho :)

packages/core/docs/integrate/integration-guide.md
```diff
@@ -340,78 +255,108 @@ const factoryParams: UseUserFactoryParams = {
...
+  return product?.name || '';
+};
+```
+:::
 
-In the default theme, components used to display or modify integration-specific data, like forms or checkout information are blank. This is because each integration might use different data formats and properties.
+## Create a theme
 
-You need to create few Vue components and JavaScript files:
+Some form or checkout components are blank in the default theme because they display or modify integration-specific data. For this reason, you need to create few Vue components and JavaScript files:
```
igorwojciechowski, line 359:
```suggestion
Some forms or checkout components are blank in the default theme because they display or modify integration-specific data. For this reason, you need to create few Vue components and JavaScript files:
```
Plural, right?

igorwojciechowski (changes requested):

packages/core/docs/integrate/integration-guide.md
filipsobol: Nice catch. I removed `(optionally)` before Typescript - integrators should know it. I think there is no way around it.

filipsobol:
Related update in the Boilerplate repo - https://github.com/vuestorefront/ecommerce-integration-boilerplate/pull/11

packages/core/docs/advanced/extending-graphql-queries.md
```diff
@@ -1,4 +1,4 @@
-# Extending GraphQL Queries
+# Extending GraphQL queries
```
igorwojciechowski, line 1: Actually, that might not necessarilly be a bad thing to write titiles using capital letters. In many places, e.g. newspaper articles, etc. or even documentation sites, titles/headers are being capitalized. Still seems more like a choice than only good way to do it.

packages/core/docs/integrate/integration-guide.md
```diff
@@ -5,319 +5,232 @@ The application does not reload automatically after saving the changes in Server
...
 
 ::: warning Want to build an integration?
-If you want to integrate with Vue Storefront, please **contact the core team on our [Discord](https://discord.vuestorefront.io) server** first. We are eager to help you with building it and ensuring its high quality! Building the integration with the help of the core team is the best way to keep its quality high and making it officially recommended once completed.
+If you want to integrate with Vue Storefront, please **contact the core team on our [Discord](https://discord.vuestorefront.io) server**. We are eager to help you to ensure its high quality and maybe even officially recommend it 😉
 :::
 
 ## Introduction
 
-Integrating any eCommerce platform with Vue Storefront is a relatively simple process. The only requirement that the eCommerce platform needs to fulfill to be integrated is having a fully functional REST/GraphQL API.
+Integrating an eCommerce platform with Vue Storefront sounds scary. Luckily, some of our partners and community members with different seniority levels have successfully done it. We are sure that even without prior experience with Vue Storefront, you can too.
```
igorwojciechowski, line 13:
```suggestion
Integrating an eCommerce platform with Vue Storefront sounds scary. Luckily, many of our partners and community members with different seniority levels have successfully done it. We are sure that even without prior experience with Vue Storefront, you can too.
```
So that it does not imply that some have failed as well :D

packages/core/docs/integrate/integration-guide.md
```diff
@@ -5,319 +5,232 @@ The application does not reload automatically after saving the changes in Server
...
 
-This documentation will guide you through integrating your eCommerce platform with Vue Storefront.
+This tutorial will guide you through the process of creating integration and explain the concepts behind Vue Storefront.
```
igorwojciechowski, line 15: As it is not a tutorial per se, maybe we could look for an another word? "guide" would be an reoccurance but it might be worth looking for some synonym, I dunno.

packages/core/docs/integrate/integration-guide.md
```diff
@@ -5,319 +5,232 @@ The application does not reload automatically after saving the changes in Server
...
-- set the default billing address
-- see past orders
+### Theme
 
-**Wishlist**
+This project is a template for generating new Vue Storefront shops. It's a [Nuxt.js](https://nuxtjs.org/) application that contains pages, Vue components, and assets. It uses `composables` to interact with the platform and `getters` to display the data to the user. 
 
-- add/remove the product
-- associate with the logged-in user
+Out of the box, the `theme` directory doesn't contain much - just a few configuration files and empty directories. However, this doesn't mean that you have to create the whole theme from scratch. When your integration is ready, you can use our CLI to combine this project with our base theme to generate a new Nuxt.js application with all necessary pieces inside.
```
igorwojciechowski, line 56: Actually, one of these `directories` could stay a `folder` :) The last comment intention was to avoid using same word in the same sentence:

```suggestion
Out of the box, the `theme` directory doesn't contain much - just a few configuration files and empty folders. However, ...
```

igorwojciechowski (approved):

PR #6063, opened 2021-07-08
[EN-365]: Constant 500 error on middleware even if integration api returns different one

lsliwaradioluz:
### Related Issues

closes [EN-365](https://vsf.atlassian.net/browse/EN-365)

A simple change removing the hard-coded `500` status code set for unsuccessful API requests. Now the one coming from the integration API is taken instead. 

### Pull Request Checklist

- [ ] Default Theme
- [ ] Capybara Theme


packages/core/middleware/src/createServer.ts
```diff
@@ -42,8 +42,7 @@ function createServer (config: MiddlewareConfig): Express {
 
       res.send(platformResponse);
     } catch (error) {
-      res.status(500);
-
+      res.status(error.networkError.statusCode);
```
Fifciu, line 45:
```suggestion
      res.status(error.networkError.statusCode || 500);
```
Maybe like that?

Fifciu (changes requested):

packages/core/middleware/src/createServer.ts
filipsobol: This will work only with Apollo GraphQL client. Other GraphQL or REST clients will throw errors with different structure.

lsliwaradioluz: @Fifciu @filipsobol Guys, I've created a little helper with a purpose to agnosticize getting the status from various error objects provided by different integrations. Could this be a more versatile solution?

Razz21: @lsliwaradioluz quick reminder - `null` also has type `object`

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "build:docs": "cd packages/core/docs && yarn build",
     "dev:docs": "cd packages/core/docs && yarn dev",
-    "build:core": "cd packages/core/core && yarn build && cd - && yarn build:middleware",
```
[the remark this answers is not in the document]
lsliwaradioluz: @filipsobol [changed](https://github.com/vuestorefront/vue-storefront/pull/6063/commits/ae642cdfcdcc49a715e6c2365c24f016b2b96390) as requested in private

packages/core/middleware/src/createServer.ts
lsliwaradioluz: Thank you @Razz21

filipsobol (approved):

packages/core/middleware/src/helpers.ts
```diff
@@ -0,0 +1,18 @@
...
+      if (getAgnosticStatusCode(o, ...keys)) {
+        return getAgnosticStatusCode(o, ...keys);
+      }
+    }
+  } else if (obj && typeof obj === 'object') {
+    for (const key of Object.keys(obj)) {
+      if (keys.includes(key)) {
+        return obj[key];
+      } else if (getAgnosticStatusCode(obj[key], ...keys)) {
+        return getAgnosticStatusCode(obj[key], ...keys);
```
Fifciu, line 13: I don't like it is called twice. Could you assign it to some variable and call only once?

packages/core/middleware/src/helpers.ts
```diff
@@ -0,0 +1,18 @@
+export const getAgnosticStatusCode = (obj: unknown, ...keys: string[]) => {
+  if (Array.isArray(obj)) {
+    for (const o of obj) {
```
Fifciu, line 3: Maybe change `obj` to `objOrArray` and then here use `const obj of objOrArray`?

packages/core/middleware/src/helpers.ts
```diff
@@ -0,0 +1,18 @@
...
+      if (getAgnosticStatusCode(o, ...keys)) {
+        return getAgnosticStatusCode(o, ...keys);
+      }
+    }
+  } else if (obj && typeof obj === 'object') {
+    for (const key of Object.keys(obj)) {
+      if (keys.includes(key)) {
```
Fifciu, line 10: Would it be possible to do smth like?:
```ts
const variableNameExplainingUsWhatsItChecking = keys.includes(key);
```
Maybe it would make core more readable - currently I do not have idea why there is a condition like that and maybe with suggested approach I would

packages/core/middleware/src/helpers.ts
```diff
@@ -0,0 +1,18 @@
...
+        return obj[key];
+      } else if (getAgnosticStatusCode(obj[key], ...keys)) {
+        return getAgnosticStatusCode(obj[key], ...keys);
+      }
+    }
+  }
+  return null;
```
Fifciu, line 17: If you remove this line it will be equal `return undefined` - isn't it ok?

Fifciu (changes requested):
A few suggestions from me.

Please also write unit tests for `getAgnosticStatusCode` - you won't regret it :D

packages/core/middleware/src/helpers.ts
```diff
@@ -0,0 +1,18 @@
+export const getAgnosticStatusCode = (obj: unknown, ...keys: string[]) => {
+  if (Array.isArray(obj)) {
+    for (const o of obj) {
+      if (getAgnosticStatusCode(o, ...keys)) {
```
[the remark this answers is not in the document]
Fifciu: or above from this perspective

Fifciu (approved):

Baroshem:
@lsliwaradioluz @filipsobol @Fifciu I am not entirely sure whether this is the correct approach for that. In my opinion we should create an agnostic 'error thrower' in our integration rather than in core itself. Something like an error mapper that would map the error thrown by the integration and conert it to the form that is accepted and passed further by middleware. In the proposed approach we are just saying: "Search recurly through properties of the error object to find status and code". Shouldn't we just pass the error in generic form from middleware back to the frontend? Maybe this is just my opinion but I would like you to make comments so we are aligned in this topic :)

filipsobol:
> @lsliwaradioluz @filipsobol @Fifciu I am not entirely sure whether this is the correct approach for that. In my opinion we should create an agnostic 'error thrower' in our integration rather than in core itself. Something like an error mapper that would map the error thrown by the integration and conert it to the form that is accepted and passed further by middleware. In the proposed approach we are just saying: "Search recurly through properties of the error object to find status and code". Shouldn't we just pass the error in generic form from middleware back to the frontend? Maybe this is just my opinion but I would like you to make comments so we are aligned in this topic :)

I absolutely agree and this was my initial suggestion. However, @andrzejewsky said that we can't expect integrations to throw Error objects in specific format.

Agnostic Error object could contain specific properties like `message`, `httpCode` and whatever else we could need. If properties will be known, the Error object can easily be converted into HTTP response (with body containing all the keys and proper status code) and even be recreated on the frontend. This way we could have errors in the same format in the middleware and frontend.

Fifciu:
> > @lsliwaradioluz @filipsobol @Fifciu I am not entirely sure whether this is the correct approach for that. In my opinion we should create an agnostic 'error thrower' in our integration rather than in core itself. Something like an error mapper that would map the error thrown by the integration and conert it to the form that is accepted and passed further by middleware. In the proposed approach we are just saying: "Search recurly through properties of the error object to find status and code". Shouldn't we just pass the error in generic form from middleware back to the frontend? Maybe this is just my opinion but I would like you to make comments so we are aligned in this topic :)
> 
> I absolutely agree and this was my initial suggestion. However, @andrzejewsky said that we can't expect integrations to throw Error objects in specific format.
> 
> Agnostic Error object could contain specific properties like `message`, `httpCode` and whatever else we could need. If properties will be known, the Error object can easily be converted into HTTP response (with body containing all the keys and proper status code) and even be recreated on the frontend. This way we could have errors in the same format in the middleware and frontend.

Totally agree

Baroshem (approved):
@lsliwaradioluz 
I approve this pull request but also created a task in our Jira to develop a generic error mapper solution later

Fifciu:
> @lsliwaradioluz
> I approve this pull request but also created a task in our Jira to develop a generic error mapper solution later

Yup, let's do this

Baroshem:
@lsliwaradioluz 

Please fix the issues before merging, you have errors in tests

lsliwaradioluz:
@Baroshem done

PR #6067, opened 2021-07-13
Feat: PLP simple caching

lsliwaradioluz:

Hi guys, here's just a quick idea of mine.

#### The problem
SPAs are obviously great for performance, allowing us to dynamically render content onto a single HTML document through the power of Javascript. However, since the routes are changing but there's no true page reload, the out-of-the-box browser cache is not working the way it usually does in several cases. That way we're sending a lot of redundant requests for data we had already fetched. 

The best example here would be our PLPs. While looking for products, we usually leave and re-enter them many times in a short period of time. Unfortunately, every time we hit the browser's `back` button to navigate back, our `Category.vue` page component sends requests for categories and product projections. It happens over and over again. 

https://www.loom.com/share/473eecc3b15f4b43a552717cb87caee7

#### The solution
How about introducing a simple caching of the PLP for our boilerplate? It could be of great use in projects without [more sophisticated mechanism](https://docs.vuestorefront.io/v2/advanced/ssr-cache.html#installation) applied. 

It could be achieved through the `keep-alive` option passed to the `<nuxt />` component in the default layout. 
image

- we are caching the Category component here and we cache it only once - hence we do not overburden our app
- we key each component rendered by the `<nuxt />` component individually - hence we allow for a full reload in case of changing sorting params or filters

Navigating back to the PLP is way less painful and sends 0 requests instead of 2-3: 

https://www.loom.com/share/6c520d6b09f241139a56e8a811757126

This is just a random thought that popped into my head. We could implement that in our starting vsf next project. We should also mention in the SSR caching docs that this feature should be removed if one decides to use redis or any other driver.


Baroshem (approved):
Looks good to me!

filipsobol (approved):

dawid-ziobro:
We've introduced other way of caching, also it's an old PR. I'm closing it.

PR #6070, opened 2021-07-13
fix: links for banner buttons

justyna-13:
Waiting on a new(?) issue for changelog

### Related Issues

closes #

### Pull Request Checklist

- [ ] Default Theme
- [ ] Capybara Theme

### DIFF
--- a/packages/commercetools/theme/nuxt.config.js
+++ b/packages/commercetools/theme/nuxt.config.js
@@ -148,5 +148,11 @@ export default {
         })
       })
     ]
+  },
+  publicRuntimeConfig: {
+    homeBannerA: process.env.HOME_BANNER_A || '/',
+    homeBannerB: process.env.HOME_BANNER_B || '/',
+    homeBannerC: process.env.HOME_BANNER_C || '/',
+    homeBannerD: process.env.HOME_BANNER_D || '/'
   }
 };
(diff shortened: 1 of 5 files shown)


packages/commercetools/theme/nuxt.config.js
```diff
@@ -148,5 +148,11 @@ export default {
         })
       })
     ]
+  },
+  publicRuntimeConfig: {
+    homeBannerA: process.env.HOME_BANNER_A || '/',
+    homeBannerB: process.env.HOME_BANNER_B || '/',
+    homeBannerC: process.env.HOME_BANNER_C || '/',
+    homeBannerD: process.env.HOME_BANNER_D || '/'
```
Mateuszpietrusinski, line 156: Why are you setting these values from `process.env`?

justyna-13: we've already talked about it but I'll leave an explanation for others - it's used because we don't want to have any hardcoded links to our categories that are useful for demo purposes

filipsobol: Using `procss.env` isn't any better - I would say that it's barely different from hardcoding the URL in the template.

IMHO, **it's a very poor practice to add confusing code like this** (link loaded from a variable loaded from configuration, that is loaded from the environment variables) **to ALL commercetools projects just for our demo**.

If I saw `this.$config` in the template without any context, I would have to google it. I would probably guess that it comes from Nuxt.js, and I would include it in my Google search term. But most developers would google "vsf this.$congfig" and find nothing (or docs for VSF1).

I know that this probably isn't your decision, but it's a NO from me.

@Baroshem

andrzejewsky (approved):

PR #6093, opened 2021-07-20
Cache control headers for home, cateogry, product

andrzejewsky:


Baroshem (approved):

packages/core/nuxt-theme-module/theme/helpers/cacheControl.js
```diff
@@ -0,0 +1,11 @@
+const cacheControl = (values) => ({ res }) => {
```
Fifciu, line 1: Maybe we should put it in `middleware` dir? I know it is a function that returns middleware, that's why I am not 100% sure. Wdyt?

Fifciu (approved):

packages/core/nuxt-theme-module/theme/helpers/cacheControl.js
andrzejewsky: I want this middleware jus for specific pages, and also I want to see in the page components that they are cached

PR #6096, opened 2021-07-20
feat(en-274): wishlist slider modification (add to cart)

lsliwaradioluz:
### Related Issues

closes [EN-274](https://vsf.atlassian.net/browse/EN-274)

This PR extends/modifies the Wishlist Sidebar functionality so that it:
- has the `Add all to cart` button if there are items on the wishlist. Using it adds all wishlist items to cart and clears the wishlist
- has a working `Add to cart` button on every product tile instead of the inactive `Save for later` button
- navigates to the PDP and toggles the sidebar if a user clicks on the product name

### Screenshots of Visual Changes before/after (if There Are Any)

image

### Pull Request Checklist

- [ ] Default Theme
- [ ] Capybara Theme


packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
```diff
@@ -28,16 +28,25 @@
...
                   <div class="collected-product__properties">
                     <SfProperty v-for="(attribute, key) in wishlistGetters.getItemAttributes(product, ['color', 'size'])" :key="key" :name="key" :value="attribute"/>
                   </div>
                 </template>
+                <template #actions>
+                  <SfButton
+                    class="sf-button--text desktop-only"
+                    @click.native="addToCart(product)"
+                  >
+                    Add to cart
```
Baroshem, line 47: i18n where possible. It will allow customer to change string values globally and allow for translations :)

packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
```diff
@@ -100,25 +120,50 @@ export default {
...
     const totalItems = computed(() => wishlistGetters.getTotalItems(wishlist.value));
 
+    watch(() => root.$route, () => {
+      if (isWishlistSidebarOpen.value) toggleWishlistSidebar();
+    });
+
+    const addToCart = async (product) => {
+      await addItem({
+        product: { sku: product.variant.sku },
+        quantity: 1
```
Baroshem, line 139: Should this be hardcoded? I don't remember exactly whether we have a qty ref in wishlist but we could introduce it here to add item with desired quantity, not just one.

packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
```diff
@@ -100,25 +120,50 @@ export default {
...
+      });
+      removeFromWishlist({ product });
+    };
+
+    const addAllToCart = async () => {
+      await Promise.all(
```
Baroshem, line 145: This is not the best approach for adding bulk products to cart because if we have like a 10 products in our wishlist and then press this button we would fire 10 async requests to the backend. Could you please check if we have a more efficient way of adding to cart (something like a bulk add to cart)?

Baroshem (changes requested):
Some smaller and bigger comments. Please, take a look

packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
lsliwaradioluz: @Baroshem I don't think we have too many existing options here. However, I could easily create such an api method from scratch which would accept multiple products and create multiple `addLineItem` actions from them. That would allow us to add the new functionality in a neat, clean and performance-saving way, without any breaking changes.
```ts
interface ProductPayload {
  product: ProductVariant,
  quantity: Number
}

addMultipleToCart(
  settings,
  { id, version }: CartDetails,
  products: ProductPayload[],
  customQuery?: CustomQuery
 ) => Promise<CartResponse>
 ```

packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
lsliwaradioluz: @Baroshem Unfortunately not, there is only a button `Add to cart`, without the qty selector
image

Baroshem: Ok, so let's leave it as it is :)

packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
Baroshem: I see. This would require an RFC I believe because it is a completely new feature. @andrzejewsky what do you think about it?

lsliwaradioluz: Same for `clearCart` which is mocked at the moment. It would require a similar approach. I could implement both.

andrzejewsky: such logic shouldn't be a part of the front-end... you are implemented to some extent business logic here. That's supposed to be implemented on the backend-side only and as front-end dev, you would need to know that when you call a specific endpoint, it will transfer your product from the current wishlist to cart items. 

Additionally, you are implementing the custom feature, that's not a part of comemrcetools at all, while VSF is the only client of that certain platform. If you want to add something like this, you can create an api-client function that will do than transition.

What's more, please next time do a research of the API you integrate with, because quick check and you can add all of the items within one request, instead of iterating and making separate calls

lsliwaradioluz: > Additionally, you are implementing the custom feature, that's not a part of comemrcetools at all, while VSF is the only client of that certain platform. If you want to add something like this, you can create an api-client function that will do than transition.

> What's more, please next time do a research of the API you integrate with, because quick check and you can add all of the items within one request, instead of iterating and making separate calls

@andrzejewsky Correct me if I'm wrong, but isn't that what I had proposed here?

> @Baroshem I don't think we have too many existing options here. However, I could easily create such an api method from scratch which would accept multiple products and create multiple addLineItem actions from them. That would allow us to add the new functionality in a neat, clean and performance-saving way, without any breaking changes.

andrzejewsky: @lsliwaradioluz yup, just posted in general what we can proceed with scenarios like this one

lsliwaradioluz: @andrzejewsky Ok, so basically doing it like I did above (with the Promise.all() thing) is definitely a no-no - agreed. However, is there a green light for implementing those new api-client methods, namely `addMultipleToCart` and `clearCart` ? I just want to make sure.

andrzejewsky: you can create just `addLineItems` and accept array, so you can reuse that in adding item in current composable, and also developers will be able to use that function later in the code if there is a need

andrzejewsky: and ofc i would suggest to create another api-client function that will transfer wishlist items to cart items

packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
dawid-ziobro: Added

packages/core/nuxt-theme-module/theme/components/WishlistSidebar.vue
dawid-ziobro: Guys take a look at it, it's changed now.

dawid-ziobro:
Closing due to inactivity.

PR #6097, opened 2021-07-20
Remove old commands and fix core tests

filipsobol:
This PR removes commands from projects that we removed from this repository and fixes tests for core packages.


Fifciu (changes requested):

Baroshem (commented):
I havent found anything else than what Fifciu reviewed. For now I am adding comment but after the final fixes I will approve it

PR #6100, opened 2021-07-20
Bug/6099 missing dependency

jaydubb12:
### Related Issues
#6098 

closes #
6099

Appends missing dependency to package.json.
NOTE, I used version 2.15.7 of the dependency which requires use of node 12

### Pull Request Checklist

- [ ] Default Theme
- [ ] Capybara Theme

### DIFF
--- a/packages/core/docs/changelog/6089.js
+++ b/packages/core/docs/changelog/6089.js
@@ -1,0 +1,8 @@
+module.exports = {
+  description: 'Fixed incorrect method`',
+  link: 'https://github.com/vuestorefront/vue-storefront/issues/6089',
+  isBreaking: false,
+  breakingChanges: [],
+  author: 'jaydubb12',
+  linkToGitHubAccount: 'https://github.com/jaydubb12/vue-storefront'
+};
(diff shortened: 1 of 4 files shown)


packages/core/docs/changelog/6089.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  description: 'Fixed incorrect method`',
+  link: 'https://github.com/vuestorefront/vue-storefront/issues/6089',
+  isBreaking: false,
+  breakingChanges: [],
+  author: 'jaydubb12',
+  linkToGitHubAccount: 'https://github.com/jaydubb12/vue-storefront'
+};
```
filipsobol, line 8: This file is unnecessary since this change only affects an internal logger, and `log.warning` is not even used.

We don't include internal changes in changelogs if they don't affect end developers. This makes it easier for them to upgrade after we release a new version.

filipsobol (changes requested):

packages/core/docs/changelog/6089.js
andrzejewsky: yeah  @jaydubb12 can you remove it?

bloodf (approved):

PR #6126, opened 2021-07-27
refactor: updated the applyCoupon interface to match the removeCoupon

bloodf:
### Related Issues

closes [EN-267](https://vsf.atlassian.net/browse/EN-267)

change the applyCoupon composable interface so it can match the removeCoupon and be used with the AgnosticCoupon as well.

This is a breaking change on the integrators, and need to be informed.

The docs for the update were added

### Pull Request Checklist

- [ ] Default Theme
- [ ] Capybara Theme


packages/commercetools/composables/src/useCart/index.ts
```diff
@@ -50,11 +50,11 @@ const useCartFactoryParams: UseCartFactoryParams<CartDetails, LineItem, ProductV
...
-  applyCoupon: async (context: Context, { currentCart, couponCode, customQuery }) => {
+  applyCoupon: async (context: Context, { currentCart, coupon, customQuery }) => {
     const cartDetails = await getCurrentCartDetails(context, currentCart);
 
-    const { data } = await context.$ct.api.applyCartCoupon(cartDetails, couponCode, customQuery);
-    return { updatedCart: data.cart, updatedCoupon: couponCode };
+    const { data } = await context.$ct.api.applyCartCoupon(cartDetails, coupon, customQuery);
+    return { updatedCart: data.cart, updatedCoupon: coupon };
   },
   removeCoupon: async (context: Context, { currentCart, coupon, customQuery }) => {
```
filipsobol, line 59: The argument in both the `applyCoupon` and the `removeCoupon` method should be named `couponCode`, so it's obvious that we expect a code, not a Coupon object.

**The `applyCoupon` method doesn't change.** The `removeCoupon`, however, should accept `couponCode` instead of `coupon`. You can use the `currenctCart` object to extract the Coupon object based on the code and then get the `coupon.id` used in the method.

filipsobol (changes requested):

packages/commercetools/composables/src/useCart/index.ts
bloodf: fixed this problem, and changed the migrate to 2.4

packages/commercetools/composables/src/useCart/index.ts
```diff
@@ -56,18 +57,20 @@ const useCartFactoryParams: UseCartFactoryParams<CartDetails, LineItem, ProductV
     const { data } = await context.$ct.api.applyCartCoupon(cartDetails, couponCode, customQuery);
     return { updatedCart: data.cart, updatedCoupon: couponCode };
   },
-  removeCoupon: async (context: Context, { currentCart, coupon, customQuery }) => {
-    const cartDetails = await getCurrentCartDetails(context, currentCart);
+  removeCoupon: async (context: Context, { currentCart, couponCode, customQuery }) => {
+    const cartDetails = await getCurrentCartDetails<true>(context, currentCart, true);
+
+    const couponId = cartDetails.discountCodes.find((d) => d.discountCode.code === couponCode)?.discountCode?.id;
```
filipsobol, line 63:
```typescript
if (!couponId) {
    return { updatedCart: currentCart };
}
```

packages/boilerplate/composables/src/useCart/index.ts
```diff
@@ -5,9 +5,9 @@ import {
   useCartFactory,
   UseCartFactoryParams
 } from '@vue-storefront/core';
-import { Cart, CartItem, Coupon, Product } from '../types';
```
filipsobol, line 8: Is the `Coupon` object used anywhere else? If not, it can be removed.

bloodf: Its removed already :)

packages/commercetools/composables/src/useCart/index.ts
bloodf: added :)

PR #6149, opened 2021-08-03
[EN-178]: CartGetters.getDiscounts is mocked

lsliwaradioluz:
## Related Issues
closes [EN-178](https://vsf.atlassian.net/browse/EN-178)

## Short Description of the PR

The goal was to replace mocked `getDiscounts` cart getter with a fully functional implementation.

## Background
There are 5 kinds of discounts in Commercetools:
1. Product discounts: applied to products before they are even added to the cart, determine the special price (visible on PLP/PDP)
2. Absolute cart discounts: fixed amount subtracted from the cart total
3. Relative cart discounts: % of total subtracted from the cart total
4. Fixed cart discounts: it sets all (or chosen) lineItems' prices to a fixed value 
5. Gift line items: adding regular line items to cart but with 0$ and not included in total

The new `getDiscounts` getter does not take the first kind of discounts into consideration. When it comes to the remaining four, there are were a few things to consider:
- numeric cart discounts are applied to all products or a subset of products in a targeted cart, lowering their prices. It is added as a [Cart Discount object](https://docs.commercetools.com/api/projects/cartDiscounts#cartdiscount) to the lineItem's `discountedPricePerQuantity[0].discountedPrice.includedDiscounts` array. It can be either: 

a) absolute
```
{
  "discount": {
    "name": "Cart winter discount",
    "isActive": true,
    "value": {
      "type": "absolute"
    }
  },
  "discountedAmount": {
    "centAmount": 300 // absolute discount value
  }
}
```
b) relative
```
{
  "discount": {
    "name": "Cart summer discount",
    "isActive": true,
    "value": {
      "type": "relative"
    }
  },
  "discountedAmount": {
    "centAmount": 70 // absolute discount value
  }
}
```
c) fixed
```
{
  "discount": {
    "name": "Fixed price discount",
    "isActive": true,
    "value": {
      "type": "fixed"
    }
  },
  "discountedAmount": {
    "centAmount": 801 // absolute discount value
  }
}
```
- The final price of a lineItem is represented by `discountedPricePerQuantity[0].discountedPrice.value.centAmount` property (similarly to the `totalPrice.centAmount` property).
- The `discountedAmount.centAmount` refers to a single unit of a given lineItem (quantity = 1). It means that if we have added a product with `quantity = 2` the `discountedAmount.centAmount` has to be multiplied by 2 to obtain the total discount amount for a given lineItem
- gift line item discount adds a separate line item to all carts or carts meeting discount conditions if these had been specified. It can be recognized by:
a) the `lineItemMode` set to `GiftLineItem` (the default is `Standard`)
b) an object inside the `discountedPricePerQuantity` array containing the `discountedPrice.value` property equal to `0`
c) the `totalPrice.centAmount` equal to `0`
d) `discountedPricePerQuantity[0].discountedPrice.includedDiscounts` contains a similar object:
```
{
  "discount": {
    "name": "Gift line item",
    "isActive": true,
    "value": {
      "type": "giftLineItem" // crucial
    }
  },
  "discountedAmount": {
    "centAmount": 3600 // in fact, it equals lineItem's regular price
  }
}
```
Considering the information above, the `getDiscounts` getter should simply return an array of `AgnosticDiscount` objects:

```ts
interface AgnosticDiscount {
  id: string;
  name: string;
  description: string;
  value: number; // the total amount of money we saved
  valueType: 'relative' | 'absolute' | 'fixed' | 'giftLineItem'; // self-explanatory
  typeId: string; // always cart-discount
  requiresCoupon: boolean; // if `false` it means the discount is added automatically on cart creation
}
```

To generate such an array, the getter method must:
- iterate through all lineItems to check their discounts
- if it comes across a discount that hasn't been recognized before => add it to the discounts array it will return in the end
- if it comes across a discount that has already been recognized => update it's value

## The Cart Sidebar discounts problem
Do you think we should display cart discounts in Cart Sidebar? At the moment we do, which results in lineItems' prices being different from the ones we have on PLP/PDP:

cart_sidebar_plp_price_mismatch

I changed that so that the prices on PLP/PDP and CartSidebar match even if a discount is applied. However, in such scenario the discount info will only be displayed during the checkout and the client won't be aware of it before he actually reaches that stage. Do you think it is ok? **Or should we include the information on discounts in Cart Sidebar as well?** 

The price flow after the changes would look somewhat like this:
price_flow_1

price_flow_2

### Prices
Currently when a gift line item is added to the cart, we display it in CartSidebar with a regular price. Its price is also calculated into our cart's `subtotal`. This can be problematic and even make customers feel outraged ("how come this product has a normal price when it was supposed to be a gift?!"). To fix that, I had to exclude gift line items' prices from cart subtotal calculation and return

```js
{ regular: null, special: null }
```

from `getCartItemPrice` getter in case it is passed a gift line item as an argument.

The challenge here is the recognition of gift line items among other lineItems. The most straightforward way (and probably suggested by Commercetools) would be to add the `lineItemMode` field to the query and perform checks against it. It also required adding `lineItemMode` property to the `ProductVariant` interface in `/commercetools/composables/src/types/GraphQL.ts`.

Alternatively, instead of performing the check against the `lineItemMode` property, we could write a different check method which would verify if any of the following three conditions is true:
- lineItem has an object inside the `discountedPricePerQuantity` array which contains the `discountedPrice.value` property equal to `0`
- lineItem has the `totalPrice.centAmount` equal to `0`
- lineItem's `discountedPricePerQuantity[0].discountedPrice.includedDiscounts` contains a similar object:

```json
{
  "discount": {
    "name": "Gift line item",
    "isActive": true,
    "value": {
      "type": "giftLineItem" // crucial
    }
  },
  "discountedAmount": {
    "centAmount": 3600
  }
}
```
**I'm open to discussion on which approach would be better here.**

Regardless of the approach, the end result is as follows:
image

### Displaying in a cart preview
GiftLineItems discounts are returned along with all other discounts by the new `getDiscounts` cart getter. However, in my opinion, they should not be displayed the same way other discounts are in Cart Preview since we're getting a free item here and not a lower cart price (even though commercetools treats it as such). I've filtered out discounts with `valueType` equal to `giftLineItem` in our demo but, still, **I'm looking for a better way to signal presence of a gift in the cart to the customers in Cart Preview**.

## Handling coupon addition/removal
I've added logic responsible for displaying and removing coupons from the cart in checkout (the coupon apply logic was already there).
image


packages/commercetools/composables/__tests__/getters/_createPrice.spec.ts
```diff
@@ -72,26 +72,26 @@ describe('[commercetools-getters] utils/createPrice', () => {
...
-        value: { centAmount: 250 },
-        discounted: {
-          discount: {
-            isActive: true
-          },
-          value: { centAmount: 350 }
-        }
-      }
-    } as any);
+  // it('returns regular and special price for line item', () => {
```
Baroshem, line 75: Why is it left commented? ;)

packages/commercetools/composables/src/getters/_utils.ts
```diff
@@ -67,14 +67,14 @@ const getDiscount = (product: ProductVariant | LineItem) => product.price?.disco
 const getSpecialPrice = (product: ProductVariant | LineItem) => {
   const discount = getDiscount(product);
 
-  if (isLineItem(product)) {
-    const { discountedPricePerQuantity } = product;
-    const discountsLength = discountedPricePerQuantity.length;
+  // if (isLineItem(product)) {
```
Baroshem, line 70: Again comments

packages/commercetools/composables/src/getters/cartGetters.ts
```diff
@@ -72,13 +74,37 @@ export const getCartTotalItems = (cart: Cart): number => {
 
 export const getFormattedPrice = (price: number) => price as any as string;
 
-export const getCoupons = (cart: Cart): AgnosticCoupon[] => {
+export const getCartCoupons = (cart: Cart): AgnosticCoupon[] => {
```
Baroshem, line 77: I wouldn't change the name of the existing getter (even if we are not using it) because it will result in a breaking change.

packages/commercetools/composables/src/getters/cartGetters.ts
```diff
@@ -72,13 +74,37 @@ export const getCartTotalItems = (cart: Cart): number => {
...
 };
 
-// eslint-disable-next-line
-export const getDiscounts = (cart: Cart): AgnosticDiscount[] => {
-  return [];
+export const getCartDiscounts = (cart: Cart): AgnosticDiscount[] => {
+  const lineItems = getCartItems(cart);
+  return lineItems.reduce((reducedDiscounts, lineItem) => {
+    const discounts = lineItem.discountedPricePerQuantity;
+    const discountsLength = discounts.length;
```
Baroshem, line 85: You can instead store a last discount item in this variable
```suggestion
    const lastDiscountItemIndex = discounts.length - 1;
    
    discounts[lastDiscountItemIndex]?.discountedPrice.includedDiscounts ....
```

packages/commercetools/theme/components/Checkout/CartPreview.vue
```diff
@@ -42,14 +44,37 @@
...
+          v-for="coupon in coupons"
+          :key="coupon.id"
+          :name="coupon.name"
+          class="sf-property--full-width sf-property--small coupon"
+        >
+          <template #value>
+            <SfButton
+              class="sf-button--text desktop-only"
+              style="margin: 0 0 1rem auto; display: block;"
+              @click="removeCoupon({ coupon: coupon })"
```
Baroshem, line 59:
```suggestion
              @click="removeCoupon({ coupon })"
```

packages/core/core/src/types.ts
```diff
@@ -522,9 +522,11 @@ export interface AgnosticCoupon {
 export interface AgnosticDiscount {
   id: string;
   name: string;
-  description: string;
   value: number;
-  code?: string;
+  valueType: 'relative' | 'absolute' | 'fixed' | 'giftLineItem';
```
Baroshem, line 526: I dont think these valueTypes are agnostic. Maybe we could add just a string here and use enum in the integration (CT)?

packages/core/nuxt-theme-module/theme/lang/en.js
```diff
@@ -120,6 +120,8 @@ export default {
   'Product suggestions': 'Product suggestions',
   'Search for items': 'Search for items',
   'Enter promo code': 'Enter promo code',
+  'This promo code is invalid': 'This promo code is invalid',
```
Baroshem, line 123: You are missing translation in german. Please add it as well :)

Baroshem (changes requested):

packages/commercetools/composables/__tests__/getters/_createPrice.spec.ts
lsliwaradioluz: Can I remove it altogether?

packages/commercetools/composables/src/getters/cartGetters.ts
lsliwaradioluz: I think it doesn't really matter here since in the end, the function is still exported as `getCoupons` inside the `cartGetters` object. What changed is only the naming of the function itself. I just thought it might be good to keep it consistend with the remaining function names in the file but its purely optional here and the change could safely be reverted.

```
const cartGetters: CartGetters<Cart, LineItem> = {
  getTotals: getCartTotals,
  getShippingPrice: getCartShippingPrice,
  getItems: getCartItems,
  getItemName: getCartItemName,
  getItemImage: getCartItemImage,
  getItemPrice: getCartItemPrice,
  getItemQty: getCartItemQty,
  getItemAttributes: getCartItemAttributes,
  getItemSku: getCartItemSku,
  getTotalItems: getCartTotalItems,
  getFormattedPrice,
  getCoupons: getCartCoupons, // still getCoupons
  getDiscounts: getCartDiscounts
};
```

packages/core/nuxt-theme-module/theme/lang/en.js
lsliwaradioluz: [Added](https://github.com/vuestorefront/vue-storefront/pull/6149/commits/f73929a75dc726802ba4c51122a1920e8b2959fa) ✔️

packages/core/core/src/types.ts
lsliwaradioluz: Like [this](https://github.com/vuestorefront/vue-storefront/pull/6149/commits/67a560c93562346fb4c54cf89349528338747bff)? :)

packages/commercetools/composables/src/getters/_utils.ts
```diff
@@ -84,7 +84,7 @@ const getSpecialPrice = (product: ProductVariant | LineItem) => {
 };
 
 export const createPrice = (product: ProductVariant | LineItem): AgnosticPrice => {
-  if (!product) {
+  if (!product || product.lineItemMode === 'GiftLineItem') {
```
fhaertig, line 87: I think you could use the isLineItem(product) function instead of extending the ProductVariant with lineItemMode (this might be confusing as the commercetools ProductVariant does not have the attribute)

fhaertig (commented):
Nice work, I really appreciate the effort you put in this request! Just added some small ideas. Additionally I encoutered a problem with GiftLineItems: If you remove the line item directly instead of removing the discount, you are not able to add the gift line item by code again even when explicitly removing the discount before.
I solved it by disabling the ability to remove gift line items by UI

PR #6168, opened 2021-08-06
Feature: Custom `where` in product, category and order search + `not` option in product filters

lsliwaradioluz:
#### Adding the `not` option to product filters
Yesterday I encountered a situation where I wanted to fetch a product using the `getProduct` method but only if it **didn't** have an enum attribute set to a particular value. That would require creating a query predicate similar to this:

```
masterData(current(masterVariant(not(attributes(name="${attribute_name}" and value(key="${attribute_value}""))))))
```

Such a query predicate can be created by passing an appropriate `filters` property to the search params. Unfortunately, our current `buildProductWhere` method does not support the negations. Hence I've updated the `Filter` interface to:

```
interface Filter {
  type: AttributeType;
  name: string;
  value: any;
  not?: boolean; // <= new property
}
```
and added handling of the new `not` property by the `mapFilterToPredicate` method:

```
let predicate = `attributes(name = "${filter.name}" and ${valuePredicate})`;
if (filter.not) {
  predicate = `not(${predicate})`;
}

return `masterData(current(masterVariant(${predicate})))`;
```
#### Adding the custom where option to product, category and order search
The aforementioned problem with passing `not` is only one of many edge cases. In my opinion, sooner or later, users will encounter situations where the VSF layer will come up short and the full control over our query predicates will be necessary. To solve that problem, we could allow passing a custom `where` among other search params for products, categories and orders.

Custom `where` would not overwrite the one created by the `build[entity-name]Where` helper. It would be mixed-in instead (after the `and` connective). It could serve as a supplement to the generated query predicate or replace it altogether, ensuring full flexibility:

```
const buildProductWhere = (settings, search) => {
  //...
  if (search?.where) {
    predicates.push(search.where);
  }
  
  return predicates.join(' and ');
}
```

### Pull Request Checklist

- [ ] Default Theme
- [ ] Capybara Theme


Baroshem (approved):
Looks good to me, but let's wait for Fifciu :)

Fifciu (changes requested):
The feature looks great for me. Really useful one and the code is of good quality.

One request from me: please add unit tests for new `where` parameter.
One suggestion from me: are you sure `not` as a property name is ok? Maybe `negate` instead? I am not sure. It is worth to ask @filrak about that too

lsliwaradioluz:
@Fifciu Alright, I'll add him as a reviewer here.

lsliwaradioluz:
@Fifciu I've added the required tests. Let's wait for @filrak insight on the `not` naming. He's on PTO today but should be back on Monday.

Razz21:
@lsliwaradioluz please add TS definitions to composables and update documentation.

dawid-ziobro:
Closing due to possible other solution found and inactivity.

PR #6192, opened 2021-08-16
Fix api ref + missing typings

lukaszjedrasik:
closes #6192

Fixes for broken API references after release.


packages/commercetools/api-client/package.json
```diff
@@ -5,7 +5,7 @@
   "server": "server/index.js",
   "main": "lib/index.cjs.js",
   "module": "lib/index.es.js",
-  "types": "lib/index.d.ts",
+  "types": "lib/*/*/src/index.d.ts",
```
filipsobol, line 8: I'm afraid that's not right. After the package is released, this path will be incorrect, and typings might stop working.

packages/commercetools/api-client/src/index.ts
```diff
@@ -1,7 +1,9 @@
+import * as cartActions from './helpers/cart/actions';
+
 export * from './fragments';
 export * from './types/Api';
 export * from './types/GraphQL';
 export * from './types/setup';
 export { isAnonymousSession, isUserSession } from './helpers/utils';
 export { createErrorHandler } from './helpers/commercetoolsLink/graphqlError';
-export * as cartActions from './helpers/cart/actions';
+export { cartActions };
```
filipsobol, line 9: Is there a reason for this change? The previous implementation is cleaner and does the same thing.

packages/core/docs/Dockerfile
```diff
@@ -4,9 +4,15 @@ WORKDIR /var/www
 
 COPY . .
 
-RUN npm install \
-   && npm run build
+RUN yarn install --network-concurrency 1 \
+    && yarn build:core \
+    && yarn build:ct:api-client \
+    && yarn cache clean --all
```
filipsobol, line 10: Can you explain `--network-concurrency 1` and `yarn cache clean --all`?

packages/commercetools/api-client/src/index.ts
lukaszjedrasik: It is workaround for `api-extractor`
https://github.com/microsoft/rushstack/issues/2780

packages/commercetools/api-client/package.json
lukaszjedrasik: Actually, we don't have `index.d.ts` inside `lib` folder

packages/core/docs/Dockerfile
lukaszjedrasik: I think that was unnecessary. Removed it.

packages/commercetools/api-client/package.json
```diff
@@ -5,7 +5,7 @@
   "server": "server/index.js",
   "main": "lib/index.cjs.js",
   "module": "lib/index.es.js",
-  "types": "lib/index.d.ts",
+  "types": "lib/commercetools/api-client/src/index.d.ts",
```
filipsobol, line 8: This still is incorrect. Once this package is built, the types are available under the `lib/index.d.ts` path

filipsobol (changes requested):

packages/commercetools/api-client/package.json
filipsobol: We do when you build the package using `yarn build:ct:tools`

packages/commercetools/api-client/package.json
lukaszjedrasik: I'm not sure 
Zrzut ekranu 2021-08-16 o 11 29 44

lukaszjedrasik: Updated:

Zrzut ekranu 2021-08-16 o 13 57 39

filipsobol (approved):

PR #6193, opened 2021-08-16
[EN-404] FIX: Products cannot be added to wishlist for logged-in customer

lsliwaradioluz:
### Related Issues
closes [EN-404](https://vsf.atlassian.net/browse/EN-404)

We're dealing with these problems: 
- if I have a guest wishlist and then log in, I can not add anything to the wishlist after that (GraphQL error: Version mismatch. Concurrent modification)
- user wishlist is not loaded right after log in. It is only loaded after adding another product to it or reloading the page
- user wishlist is not cleared after log out

The key here is to implement wishlist `loading` on login and `clearing` on logout. Adding these to the authentication flow gets rid of the aforementioned problems. The solution I provide is a quick, 5-minute one. Works fine but I don't believe it is optimal. As suggested by @Baroshem, we should create a separate `useUser` composable in our enterprise-packages which handles it instead.


lsliwaradioluz:
I close this PR since the issue has been solved in the enterprise-packages.

PR #6243, opened 2021-08-27
[Docs] Reorganize docs into smaller categories

filipsobol:


packages/core/docs/.vuepress/config.js
```diff
@@ -183,141 +132,173 @@ module.exports = {
...
+            ['/general/key-concepts', 'Key concepts'],
             ['/guide/theme', 'Theme'],
             ['/guide/configuration', 'Configuration'],
+            ['/advanced/internationalization', 'Internationalization'],
+            ['/advanced/logging', 'Logging'],
+            // ['/', 'Glossary']
+          ]
+        },
+        {
+          title: 'Composition',
```
AdamPawlinski, line 236: How about changing this title to "Essential elements" or "Essential ingredients", would it be better describing name for this section?

AdamPawlinski (commented):
What do you think aboout moving up "How-to guide" in commercetools integration docs above "composables" section and leave it open by default? I think it's more important when someone begins developing his project. Also it is crucial for me to have this landing page (that we spoke about), because it will be very helpful for new users to navigate throughout the docs. Also I wondered if the integration section is not hided too much.  Probably many users will be interested for the specific integration. Maybe thi landing page could be solution for that. Wdyt?

filipsobol:
> What do you think aboout moving up "How-to guide" in commercetools integration docs above "composables" section and leave it open by default? I think it's more important when someone begins developing his project. Also it is crucial for me to have this landing page (that we spoke about), because it will be very helpful for new users to navigate throughout the docs. Also I wondered if the integration section is not hided too much. Probably many users will be interested for the specific integration. Maybe thi landing page could be solution for that. Wdyt?

Good suggestions 👍 I agree about moving "How-to guides" up. I updated the PR.

I think that this PR makes the "Integrations" link more visible than before, but I also agree that we could emphasize it more, especially with the landing page. I will talk with our designer about this.

packages/core/docs/.vuepress/config.js
filipsobol: I think `Composition` better explains what's inside of this category. `Essentials` or `Essential elements` is too generic and similar to `Getting started` since it also contains things essential to working with VSF projects.

WDYT @igorwojciechowski and @lukaszjedrasik?

lukaszjedrasik: For me `Composition` is cleaner and sounds cool 😄. Even in the vue3 docs, we have a Reusability & `Composition` section.

AdamPawlinski: that's okay let's leave 'Composition' :)

AdamPawlinski (approved):

lukaszjedrasik (approved):

PR #6261, opened 2021-09-02
Downgrade GH Action to fix deployment

filipsobol:
Caused by https://github.com/elgohr/Publish-Docker-Github-Action/issues/134

### DIFF
--- a/.github/workflows/deploy-storefrontcloud.yml
+++ b/.github/workflows/deploy-storefrontcloud.yml
@@ -14,9 +14,9 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: "12.22.4"
+          node-version: "12.x"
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-storefrontcloud.yml
```diff
@@ -16,7 +16,8 @@ jobs:
         with:
           node-version: "12.x"
       - name: Build and publish docker image
-        uses: elgohr/Publish-Docker-Github-Action@master
+        # 3.04 is hardcoded as a workaround for https://github.com/elgohr/Publish-Docker-Github-Action/issues/134
```
mkoszut, line 19: I think you can keep it as a comment
`# uses: elgohr/Publish-Docker-Github-Action@master`

mkoszut (changes requested):

PR #6353, opened 2021-09-27
fix(cache): cache package invalidation bug

lsliwaradioluz:
## Description

Our cache package is behaving unexpectedly while hitting the `/cache-invalidate` endpoint. Instead of doing its job, it is resulting in the browser downloading an empty file named `cache-invalidate`.

It turns out the bug is occurring only if the `@nuxt/pwa` module is enabled. The service worker registered by it is unhappy with the `/cache-invalidate` endpoint handler using the `response.end()` method and is causing some internal redirect. The bug can be reproduced only in production. By production I mean an application deployed to the server since the bug cannot be reproduced after building our app locally (even though the service worker is registered there too).

To fix the bug, I've changed the previous handler logic so that it uses `response.status().send()` instead of `response.writeHead().end()`.

### DIFF
--- a/packages/core/cache/nuxt/index.js
+++ b/packages/core/cache/nuxt/index.js
@@ -28,17 +28,16 @@ function createInvalidationEndpoint (driver, options) {
         tags: Array.from(new Set(tags))
       });
 
-      response.status(200).send('Cache invalidated successfully!');
+      response
+        .status(200)
+        .send('Cache invalidated successfully!');
     } catch (error) {
       Logger.error('Cache driver thrown an error when invalidating cache! Operation skipped.');
       Logger.error(error);
 
       response
         .status(500)
-        .send(
-          'Cache driver thrown an error when invalidating cache! Operation skipped. ' +
-          error
-        );
+        .send(`Cache driver thrown an error when invalidating cache! Operation skipped. ${error}`);
     }
   };
 
--- a/packages/core/cache/nuxt/index.js
+++ b/packages/core/cache/nuxt/index.js
@@ -28,15 +28,18 @@ function createInvalidationEndpoint (driver, options) {
         tags: Array.from(new Set(tags))
       });
 
-      response.writeHead(200);
+      response.status(200).send('Cache invalidated successfully!');
     } catch (error) {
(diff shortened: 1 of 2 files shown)


packages/core/cache/nuxt/index.js
```diff
@@ -28,15 +28,18 @@ function createInvalidationEndpoint (driver, options) {
         tags: Array.from(new Set(tags))
       });
 
-      response.writeHead(200);
+      response.status(200).send('Cache invalidated successfully!');
```
filipsobol, line 31:
```suggestion
      response
        .status(200)
        .send('Cache invalidated successfully!');
```

Just to be consistent

packages/core/cache/nuxt/index.js
```diff
@@ -28,15 +28,18 @@ function createInvalidationEndpoint (driver, options) {
...
     } catch (error) {
       Logger.error('Cache driver thrown an error when invalidating cache! Operation skipped.');
       Logger.error(error);
 
-      response.writeHead(500);
+      response
+        .status(500)
+        .send(
+          'Cache driver thrown an error when invalidating cache! Operation skipped. ' +
+          error
```
filipsobol, line 40:
```suggestion
          error.toString()
```

WDYT?

packages/core/cache/nuxt/index.js
lsliwaradioluz: Corrected in [41c9831](https://github.com/vuestorefront/vue-storefront/pull/6353/commits/41c9831c9ec2be830f7db5fbd44acd84c959f8b8)

packages/core/cache/nuxt/index.js
lsliwaradioluz: Corrected in [41c9831](https://github.com/vuestorefront/vue-storefront/pull/6353/commits/41c9831c9ec2be830f7db5fbd44acd84c959f8b8)

filipsobol (approved):

lsliwaradioluz:
This PR's branch has been created from the old & unused `next` branch. Here's the updated PR: https://github.com/vuestorefront/vue-storefront/pull/6355, this one can be closed.

PR #6404, opened 2021-10-01
chore(commercetools): update GraphQL types

filipsobol:
## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] I have read the **CONTRIBUTING** document.

#### Changelog

### DIFF
--- a/packages/commercetools/api-client/src/helpers/cart/actions.ts
+++ b/packages/commercetools/api-client/src/helpers/cart/actions.ts
@@ -110,9 +110,17 @@ export const setBillingAddressAction = (billingDetails: Address): { setBillingAd
         building: billingDetails.building,
         apartment: billingDetails.apartment,
         pOBox: billingDetails.pOBox,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
         phone: billingDetails.phone || billingDetails.contactInfo?.phone,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
         mobile: billingDetails.mobile || billingDetails.contactInfo?.mobile,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
         email: billingDetails.email || billingDetails.contactInfo?.email,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
         fax: billingDetails.fax || billingDetails.contactInfo?.fax,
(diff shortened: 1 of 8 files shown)


packages/commercetools/api-client/src/helpers/cart/actions.ts
```diff
@@ -58,9 +58,17 @@ export const setShippingAddressAction = (shippingDetails: Address): { setShippin
         building: shippingDetails.building,
         apartment: shippingDetails.apartment,
         pOBox: shippingDetails.pOBox,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
andrzejewsky, line 62: ignores? why?

andrzejewsky (changes requested):

packages/commercetools/api-client/src/helpers/cart/actions.ts
filipsobol: `contactInfo` was removed from `Address`. We already logged a warning if somebody used it in projects (lines 38-40).

However, I want this PR to be released as a patch, and removing `|| shippingDetails.contactInfo?.X` would be a breaking change. That's why in here, I added `@ts-ignore`, but in separate PR for `develop` (next minor release), I will remove this completely, together with a deprecation warning.

andrzejewsky:
```ts
// this you import
type AddressBase = {
  something: string
}

// comment here as legacy
type Address = AddressBase & {
  contactInfo: {
    street: string
  }
}


const testFunction = (shippingDetails: Address) => {

  return {
    test: shippingDetails.something,
    street: shippingDetails?.contactInfo.street || 'test'
  }
}
```

andrzejewsky: or edit generated file, and just add those fields in the interface/type but with the comment that's deprecated

andrzejewsky: however, this is BC break that already happened - if someone will access contactInfo field, it will raise an error it get undefined data even if you still saying (from the TS perspective that still exist)

andrzejewsky: what's more... you are saing that ts-ignores are for preventing from BC's while when you generated a new types, they don't have contactInfo so... means BC

filipsobol:
Will be implemented in https://github.com/vuestorefront/vue-storefront/pull/6410

PR #6420, opened 2021-10-05
feat(perf): runtime caching using composables' shared refs

lsliwaradioluz:
## Motivation

_This feature has been requested by one of our clients._

In our core, there are **5** composable factories that use the `sharedRef` util:
- useContentFactory
- useProductFactory
- useCategoryFactory
- useReviewFactory
- useSearchFactory

The content we fetch using their `search` method persists in `$sharedRefsMap`. However, with the current implementation, we rarely make use of it. You can observe that in projects where we have dynamic pages built from data coming from some CMS:

https://www.loom.com/share/9e5b84c1417645b28fac5ef3abfbdb51

In the example above we had requests for content and products (a couple of them). Every time we go back to the first page, the app re-fetches the data we already have in our "store". By utilizing the new runtime cache mechanisms, we can save bandwidth and speed up the process of navigating back to certain pages (or navigating to pages for the same time if they rely on data we've already cached!):

https://www.loom.com/share/276a72cfca39415b9b98483bd94563eb

## Description
Runtime caching is now possible thanks to the new logic inside the aforementioned composables' search methods. Whether the data will be re-fetched or not, depends on a couple of factors:
```
if (!force && isCacheValid(products, cacheTimestamp, cacheTimeToLive)) return;
```
`isCacheValid` is a utility function that will return `true` when the data we want to fetch already exists in the store or the time that has elapsed since the timestamp had been created does not exceed the `cacheTimeToLive` value. It takes in three parameters.
```ts
export const isCacheValid = (
  content: any,
  timestamp: Ref<number>,
  cacheTimeToLive?: number
) => {
```
`cacheTimeToLive` value can be passed to the composable as the second argument:
```
const { search, products } = useProduct('universal-id', 300);
```
`cacheTimestamp` is a `sharedRef` created by the second utility function this PR adds: `setCacheTimestamp`.
```
const cacheTimestamp: Ref<number> = setCacheTimestamp(`useProduct-cache-${id}`);
```
If re-fetching the data regardless of the `isCacheValid` return value is what we need, we can always use the `force` param that gives us full control over that. It is disabled by default:
```
const search = async (searchParams, force = false)
```

Long story short, by passing `cacheTimeToLive` and `force` params in certain ways, we can achieve three runtime caching strategies in VSF:
- If you want the fetched content to be cached forever and never be re-fetched (that is, as long as you stay in the SPA mode), do not pass the `cacheTimeToLive` argument to the composable.
- If you want the fetched content to be cached only for a given time, pass the `cacheTimeToLive` argument to the composable. Important! The unit of `cacheTimeToLive` is **seconds**. I.e. if you pass value `300`, your content will be cached for 300 seconds which equals 5 minutes. If this time elapses, your composable will re-fetch the data.
- If you don't want the fetched content to be cached at all, set the the `force` parameter of the composable's search method to `true` (or any other "truthy" value). It should be passed as the last argument.

## Types of changes

## Checklist:

#### Changelog

#### Tests
- [x] I have written test cases for my code
- [x] I have added tests to cover my changes.

#### Code standards

#### Docs
- [x] My change requires a change to the documentation.


packages/core/core/src/utils/runtimeCache/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  const existingTimestamp = $sharedRefsMap.get(key);
+  if (existingTimestamp) return existingTimestamp;
+  const timestamp = sharedRef(Date.now(), key);
+  return timestamp;
+};
+
+export const isCacheValid = (
+  content: any,
+  timestamp: Ref<number>,
+  cacheTimeToLive?: number
```
Fifciu, line 22: I would keep it in MS, not seconds

packages/core/core/src/types.ts
```diff
@@ -378,7 +380,7 @@ REVIEWS_SEARCH_PARAMS,
 REVIEW_ADD_PARAMS,
 API extends PlatformApi = any
 > extends Composable<API> {
-  search(params: ComposableFunctionArgs<REVIEWS_SEARCH_PARAMS>): Promise<void>;
+  search(params: ComposableFunctionArgs<REVIEWS_SEARCH_PARAMS>, force?: boolean): Promise<void>;
```
Fifciu, line 383: Again some of them has cacheTimestamp and some does not. Why?

packages/core/core/src/utils/runtimeCache/index.ts
lsliwaradioluz: Updated in [c054d61](https://github.com/vuestorefront/vue-storefront/pull/6420/commits/c054d61b40e7f812f9d6d5c4ab3cc7ce4be7ec8c).

lsliwaradioluz: But isn't re-calculating & passing such large numbers as arguments a little bit too cumbersome?

packages/core/core/src/types.ts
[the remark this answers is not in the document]
lsliwaradioluz: Added in [d65b0a7](https://github.com/vuestorefront/vue-storefront/pull/6420/commits/d65b0a7f532d7c4e40cdb945f7a43a82c568b2e7)

packages/core/core/src/types.ts
```diff
@@ -78,7 +78,7 @@ export interface UseProduct<
   products: ComputedProperty<PRODUCTS>;
   loading: ComputedProperty<boolean>;
   error: ComputedProperty<UseProductErrors>;
-  search(params: ComposableFunctionArgs<PRODUCT_SEARCH_PARAMS>): Promise<void>;
+  search(params: ComposableFunctionArgs<PRODUCT_SEARCH_PARAMS>, force?: boolean): Promise<void>;
```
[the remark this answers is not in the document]
lsliwaradioluz: Added in [d65b0a7](https://github.com/vuestorefront/vue-storefront/pull/6420/commits/d65b0a7f532d7c4e40cdb945f7a43a82c568b2e7)

packages/core/core/src/types.ts
lsliwaradioluz: Added in [d65b0a7](https://github.com/vuestorefront/vue-storefront/pull/6420/commits/d65b0a7f532d7c4e40cdb945f7a43a82c568b2e7). By the way, some of them had and some didn't because for example `useProduct` and `useReview` interfaces contained an index signature `[x: string]: any;` and were more flexible in that regard.

packages/core/core/src/types.ts
```diff
@@ -471,10 +473,11 @@ export interface UseContent<
   CONTENT_SEARCH_PARAMS,
   API extends PlatformApi = any
 > extends Composable<API> {
-  search: (params: CONTENT_SEARCH_PARAMS) => Promise<void>;
+  search: (params: CONTENT_SEARCH_PARAMS, force?: boolean) => Promise<void>;
```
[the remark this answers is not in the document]
lsliwaradioluz: That was a really good point Filip. Implementing the `force` param as part of the `params` object made things even easier and - as you'd said - more flexible. Tweaked in [fac902f](https://github.com/vuestorefront/vue-storefront/pull/6420/commits/fac902ffb4d07e3c993403c0d2b9b79fb7ee6e45)

packages/core/core/src/factories/useSearchFactory.ts
```diff
@@ -9,21 +9,23 @@ export interface UseSearchFactoryParams<RESULT, SEARCH_PARAMS> extends FactoryPa
...
 ) {
-  return function useSearch(id: string): UseSearch<RESULT, SEARCH_PARAMS> {
+  return function useSearch(id: string, cacheTimeToLive?: number): UseSearch<RESULT, SEARCH_PARAMS> {
     const result: Ref<RESULT> = sharedRef([], `useSearch-products-${id}`);
     const loading = sharedRef(false, `useSearch-loading-${id}`);
     const _factoryParams = configureFactoryParams(factoryParams);
     const error: Ref<UseSearchErrors> = sharedRef({
       search: null
     }, `useSearch-error-${id}`);
+    const cacheTimestamp: Ref<number> = setCacheTimestamp(`useSearch-cache-${id}`);
```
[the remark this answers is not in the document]
lsliwaradioluz: The `setCacheTimestamp` util prevents overwriting the existing timestamp when you navigate back to the page you'd visited before. It think it could even be abbreviated from this:
```
export const setCacheTimestamp = (key: string) => {
  const { $sharedRefsMap } = useVSFContext();
  const existingTimestamp = $sharedRefsMap.get(key);
  if (existingTimestamp) return existingTimestamp;
  const timestamp = sharedRef(Date.now(), key);
  return timestamp;
};
```

to that:

```
export const setCacheTimestamp = (key: string) => {
  const { $sharedRefsMap } = useVSFContext();
  return $sharedRefsMap.get(key) || sharedRef(Date.now(), key);
};
```

packages/core/core/__tests__/factories/useContentFactory.spec.ts
```diff
@@ -25,24 +26,43 @@ describe('[CORE - factories] useContentFactory', () => {
...
   beforeEach(() => {
     jest.clearAllMocks();
+    (isCacheValid as any).mockReturnValue(false);
     createContentFactoryMock();
   });
 
   it('returns content initial values', () => {
-    const { loading, content, error } = useContent('test-id');
-
+    jest.spyOn(Date, 'now').mockImplementationOnce(() => 1);
```
Fifciu, line 34: Nice one. I didn't know its possible to do this way

packages/core/core/src/utils/runtimeCache/index.ts
```diff
@@ -0,0 +1,23 @@
...
+) => {
+  if (_isContentCached(content.value)) {
+    if (!cacheTimeToLive) return true;
+    const now = Date.now();
+    const cacheLife = (now - timestamp.value) / 1000;
+    return cacheLife < cacheTimeToLive;
+  } else {
+    return false;
+  }
```
Fifciu, line 22:
```suggestion
  if (_isContentCached(content.value)) {
    if (!cacheTimeToLive) return true;
    const cacheLife = (Date.now() - timestamp.value) / 1000;
    return cacheLife < cacheTimeToLive;
  }
  return false;
```

packages/core/core/__tests__/setup.ts
```diff
@@ -22,7 +22,8 @@ jest.mock('../src/utils', () => ({
   vsfRef: jest.fn(ref),
   generateContext: jest.fn(() => ({ context: null })),
   configureFactoryParams: jest.fn((fParams) => fParams),
-  useVSFContext: jest.fn()
+  useVSFContext: jest.fn(),
+  isCacheValid: jest.fn()
```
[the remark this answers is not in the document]
lsliwaradioluz: I have a problem with this one.

I can set the default return value of this mock to `false` - that's correct. I can also overwrite it using `(isCacheValid as jest.Mock).mockReturnValueOnce(true)` in chosen tests. However, for some reason, the return value does not get re-set to `false` after the mock has been called by the factory function. It does if I call it directly in the test body though. 🤷

packages/core/core/__tests__/factories/useCategoryFactory.spec.ts
```diff
@@ -35,6 +36,9 @@ describe('[CORE - factories] useCategoryFactory', () => {
   });
 
   describe('methods', () => {
+    beforeEach(() => {
+      (isCacheValid as any).mockReturnValue(false);
+    });
     describe('search', () => {
       it('should invoke search', async () => {
```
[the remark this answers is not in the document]
lsliwaradioluz: [Added](https://github.com/vuestorefront/vue-storefront/pull/6420/commits/07c37903fe03d75ab48073e775c5f4bc122ae065)

Fifciu (approved):

lsliwaradioluz:
❗❗❗ THIS PR IS ON HOLD - PLEASE, DO NOT MERGE

dawid-ziobro:
This feature will not be implemented.

PR #6504, opened 2021-11-04
chore: sync `release/next` with `develop` and update Rollup

filipsobol:
This PR syncs `release/next` with `develop` and updates Rollup and its plugins, because of the error fixes in the newer versions that prevented the application from building.


packages/commercetools/api-client/src/api/applyCartCoupon/index.ts
```diff
@@ -3,6 +3,10 @@ import updateCart from '../updateCart';
 import { CartDetails, CartResponse } from '../../types/Api';
 import { addDiscountCodeAction } from '../../helpers/actions/cart';
 
+/**
+ * @remarks References:
+ * {@link CartDetails}, {@link CartResponse}
```
alefbarbeli, line 8: Why this kind of markdown is just on this branch???

lukaszjedrasik (approved):

alefbarbeli (approved):

packages/commercetools/api-client/src/api/applyCartCoupon/index.ts
alefbarbeli: update: I've looked into the `release/next` branch and I got it

PR #6531, opened 2021-11-22
feat(en-426): add keepAlive option to middleware

PiotrGrzywa:
## Description

It's adding possibility to enable keepAlive on axios.

## Related Issue

https://vsf.atlassian.net/browse/EN-426

## Motivation and Context

It helps to reduce latency between Middleware API and integration

## How Has This Been Tested?

Checking if request has a keepAlive header present.

To do it I was using interceptors like:
```
client.interceptors.request?.use(request => {
   console.log('>>> Axios: Starting Request', request);
   return request;
});
```

## Types of changes

## Checklist:

#### Changelog

#### Tests

- [ ] Default Theme
- [ ] Capybara Theme

#### Code standards

#### Docs

### DIFF
--- a/packages/core/core/src/utils/nuxt/index.ts
+++ b/packages/core/core/src/utils/nuxt/index.ts
@@ -40,8 +40,8 @@ export const integrationPlugin = (pluginFn: NuxtPlugin) => (nuxtCtx: NuxtContext
     let axiosConfig;
 
     if (process.server) {
-      const http = await import('http');
-      const https = await import('https');
+      const http = require('http');
+      const https = require('https');
 
       axiosConfig = {
         ...config.axios,
--- a/packages/core/core/src/utils/nuxt/index.ts
+++ b/packages/core/core/src/utils/nuxt/index.ts
@@ -2,8 +2,6 @@ import { createExtendIntegrationInCtx, createAddIntegrationToCtx } from './conte
 import { getIntegrationConfig, createProxiedApi } from './_proxyUtils';
 import { Context as NuxtContext, Plugin as NuxtPlugin } from '@nuxt/types';
 import axios from 'axios';
-import http from 'http';
-import https from 'https';
 
 type InjectFn = (key: string, value: any) => void;
 export type IntegrationPlugin = (pluginFn: NuxtPlugin) => NuxtPlugin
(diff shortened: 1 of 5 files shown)


packages/core/core/src/utils/nuxt/index.ts
```diff
@@ -31,7 +33,18 @@ export const integrationPlugin = (pluginFn: NuxtPlugin) => (nuxtCtx: NuxtContext
...
-    const client = axios.create(config.axios);
+    const keepAliveConfig = {
+      keepAlive: process.env.AXIOS_KEEP_ALIVE_ENABLED === 'true',
+      keepAliveMsecs: parseInt(process.env.AXIOS_KEEP_ALIVE_MSECS) || 1000,
+      maxFreeSockets: parseInt(process.env.AXIOS_MAX_FREE_SOCKETS) || 256
+    };
+
+    const client = axios.create({
+      ...config.axios,
+      httpAgent: new http.Agent(keepAliveConfig),
```
kamnog, line 44: I'm just thinking if there is a point to manage `httpAgent` and `httpsAgent` based on `process.env.NODE_ENV`. So if we are inside `DEVELOPMENT` we should care about `httpAgent` and in other case scenarios let's config `httpsAgent`. Wdyt?

kamnog (approved):
besides some minor things LGTM 👍

lukaszjedrasik (approved):

packages/core/core/src/utils/nuxt/index.ts
PiotrGrzywa: After consulting this with @filipsobol I get info that there is possibility to that production build is run in local environment, so it will be better to leave both.

packages/core/core/src/utils/nuxt/index.ts
```diff
@@ -31,7 +31,29 @@ export const integrationPlugin = (pluginFn: NuxtPlugin) => (nuxtCtx: NuxtContext
       config.axios.baseURL = process.server ? ssrMiddlewareUrl || middlewareUrl : middlewareUrl;
     }
 
-    const client = axios.create(config.axios);
+    const keepAliveConfig = {
+      keepAlive: process.env.AXIOS_KEEP_ALIVE_ENABLED === 'true',
+      keepAliveMsecs: parseInt(process.env.AXIOS_KEEP_ALIVE_MSECS) || 1000,
```
dawid-ziobro, line 36: Did you consider setting those consts in config instead of using env vars?

dawid-ziobro (approved):

PiotrGrzywa:
This had no impact on connections done to commercetools API. We should find another solution for this problem.

Issue #6536, opened 2021-11-26
[Bug]: commercetools useMakeOrder.make Cannot read property 'order' of undefined


CheStas:
### Describe the Bug

If commercetools api returns an error after  `createMyOrderFromCart` request, the request is successful but the response doesn't have  `order` field and has `graphQLErrors` field. Instead of graphQLErrors message, we observe `useMakeOrder.make Cannot read property 'order' of undefined` error message

### Current behavior

`useMakeOrder` composable doesn't have correct error in error.value.make if any occurs after request to commercetool api

### Expected behavior

check graphQLErrors field from a response and throw an error in any. graphQLErrors should be assigned to error.value.make and provide a useful error message.

### Steps to reproduce

Place an order with useMakeOrder commercetool composable that returns an error. for example, try to place an order with an item in the cart that is out of stock.

### What version of Vue Storefront are you using?

2.4.5

### What version of Node.js are you using?

14.17.2

### What browser (and version) are you using?

Chrome

### What operating system (and version) are you using?

macOS

### Relevant log output

```shell
{"message":"[GraphQL error]: Message: Some line items are out of stock at the time of placing the order: 8505., Location: , Path: order","level":"error","service":"makers","env":"dev"}

ERROR  [VSF][error]:  useMakeOrder.make Cannot read property 'order' of undefined
```

### Able to fix / change the documentation?

- [X] Yes
- [ ] No

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

CheStas:
the issue affects not only useMakeOrder composable but every composable.
look please at the example with apply coupon code.
![apply coupon 200 response](https://user-images.githubusercontent.com/17459834/148377250-f4a06ca0-4f68-489d-bccd-814ca94fe40f.png)
![apply coupon response body](https://user-images.githubusercontent.com/17459834/148377360-441963c9-902f-410b-acaf-1881cccd1f1a.png)
error log in the console
`useCart/applyCoupon TypeError: Cannot read properties of undefined (reading 'cart')`

and the example with place order
try to place an order with wrong version
![image](https://user-images.githubusercontent.com/17459834/148378027-fd77b70a-608d-47d3-a270-7fff0b71b7ef.png)
but in the console
`useMakeOrder.make TypeError: Cannot read properties of undefined (reading 'order')`

filipsobol:
The open-source version of commercetools integration is no longer supported.

PR #6538, opened 2021-11-29
fix(ct): handle error from ct create order from cart request

CheStas:
handle error after createMyOrderFromCart request.

## Description
If commercetools api returns an error after  `createMyOrderFromCart` request, the request is successful but the response doesn't have  `order` field and has `errors` field. Instead of graphQLErrors message, we observe `useMakeOrder.make Cannot read property 'order' of undefined` error message

## Related Issue
[https://github.com/vuestorefront/vue-storefront/issues/6536](https://github.com/vuestorefront/vue-storefront/issues/6536)

## Motivation and Context
instead of useful error message we observe  `useMakeOrder.make Cannot read property 'order' of undefined`

## How Has This Been Tested?
need help to test and verify it works ok.

## Types of changes

## Checklist:

#### Changelog

#### Tests

- [ ] Default Theme
- [ ] Capybara Theme

#### Code standards

#### Docs


CheStas:
that's how the issue looks on the demo page. also, look at the response. the response type doesn't match with what we expect 

image

packages/commercetools/composables/src/useMakeOrder/index.ts
```diff
@@ -15,7 +15,14 @@ const useMakeOrderFactoryParams = {
 
   make: async (context: Context, { customQuery }): Promise<Order> => {
     const { id, version } = context.cart.cart.value;
-    const response = await context.$ct.api.createMyOrderFromCart({ id, version }, customQuery);
+    // TODO fix type in @vue-storefront/core after update apollo-client
+    const response: any = await context.$ct.api.createMyOrderFromCart({ id, version }, customQuery);
+    if (response.graphQLErrors) {
+      throw response.graphQLErrors[0];
```
dawid-ziobro, line 21: I believe you want to see a message so it would be better to throw `response.graphQLErrors[0].message`

CheStas: I'm not sure about it. we then log the error and assign it in composable to `error.value.make` to have access to it in components, we want to have access not only to the message or error, but to the error code for example.

also there is a similar example in `packages/commercetools/composables/src/useForgotPassword/index.ts`

dawid-ziobro: Ah ok, i get it now

dawid-ziobro (approved):

PiotrGrzywa (approved):

PR #6545, opened 2021-12-01
fix: `Product` and `OrderHistory` components

filipsobol:


packages/core/nuxt-theme-module/theme/pages/MyAccount/OrderHistory.vue
```diff
@@ -137,7 +137,7 @@ export default {
 
     return {
       tableHeaders,
-      orders: computed(() => orders ? orders.value.results : []),
+      orders,
```
kamnog, line 140: you can get rid of `orders` as you are using `totalOrders` inside template

dawid-ziobro (approved):

packages/core/nuxt-theme-module/theme/pages/MyAccount/OrderHistory.vue
filipsobol: `orders` is used in the `v-for` loop in line 61.

PR #6551, opened 2021-12-05
fix(core): allow disabling locale redirects

lsliwaradioluz:
## Description
I think it would be a good idea to leave the door open for preventing locale redirects in vsf applications.

Last month we overwrote the default `i18n` locale redirects logic in the `i18nCookiesPlugin`. It detects the browser locale and redirects to it unconditionally (provided that it's one of the locales available in the VSF application). It happens even if the `detectBrowserLanguage` option is set to `false`.

My proposition here is to give our customers the ability to prevent redirects by setting the `detectBrowserLanguage` to `false`. And - to prevent generating the cookies by the `i18n` library - use the following configuration:

```
i18n: {
    detectBrowserLanguage: {
      useCookie: false
    }
}
```

It's the approach suggested by the nuxt-i18n [documentation](https://i18n.nuxtjs.org/browser-language-detection/):
image

## Types of changes

## Checklist:

#### Changelog
- [z] I have updated the Changelog ([V1](https://github.com/DivanteLtd/vue-storefront/blob/develop/CHANGELOG.md)) [v2](https://docs-next.vuestorefront.io/contributing/creating-changelog.html) and mentioned all breaking changes in the public API.

#### Tests

- [ ] Default Theme
- [ ] Capybara Theme

#### Code standards

#### Docs

### DIFF
--- a/packages/boilerplate/theme/nuxt.config.js
+++ b/packages/boilerplate/theme/nuxt.config.js
@@ -121,7 +121,9 @@ export default {
         }
       }
     },
-    detectBrowserLanguage: false
+    detectBrowserLanguage: {
+      useCookie: false
+    }
   },
 
   styleResources: {
--- a/packages/commercetools/theme/nuxt.config.js
+++ b/packages/commercetools/theme/nuxt.config.js
@@ -116,7 +116,9 @@ export default {
         }
       }
     },
-    detectBrowserLanguage: false
+    detectBrowserLanguage: {
+      useCookie: false
+    }
   },
   styleResources: {
     scss: [require.resolve('@storefront-ui/shared/styles/_helpers.scss', { paths: [process.cwd()] })]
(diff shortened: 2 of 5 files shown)


bloodf:
I would like to add the possibility to not update the currency also because the current version is also updating the currency cookie based on the browser locale.

dawid-ziobro (approved):

dawid-ziobro:
> I would like to add the possibility to not update the currency also because the current version is also updating the currency cookie based on the browser locale.

It will be resolved in another task

SullyAnn:
Hello, I tried to disable the accepted language of the browser by adding this :

vtoc:
The proposed fix seems incomplete to me: 
Even if no redirect is sent,```app.i18n.cookieValues``` will still store the ```targetLocale```. I think a better approach would be to set ```targetLocale``` to ```i18nOptions.defaultLocale``` if ```i18nOptions.detectBrowserLanguage``` is false and have a different option to disable redirects to i18n targets

Also ```if (redirectPath && i18nOptions.detectBrowserLanguage)``` does work as proposed, if it is set to false as it was before

filrak:
Closed due to inactivity

PR #6562, opened 2021-12-13
chore: update caching in GitHub Actions

filipsobol:


.github/workflows/test.yml
```diff
@@ -13,138 +13,36 @@ on:
...
       - name: Test CLI
         run: yarn test:cli --coverage
 
-      - name: Test core
-        run: yarn test:core --coverage
-
       - name: Test middleware
         run: yarn test:middleware --coverage
-
-  validate_integrations:
```
dawid-ziobro, line 93: What is the reason for removing it?

filipsobol: We no longer maintain boilerplate, and commercetools on this repo, and both will be deleted soon. There is no point in running tests for them here.

dawid-ziobro: Ok, that's what i've thought

dawid-ziobro (approved):

PR #6631, opened 2022-02-11
docs: refactor integrations page

filipsobol:
We need to restyle the `Integrations` page to a list because some integrations have links to a few documentation pages, which doesn't look good in a grid. I also fixed the e-commerce platform selector and added a badge for Enterprise packages because it was easy to miss before.

localhost_8080_v2_integrations_


bloodf (approved):
Just the change on the Prestashop integration

PR #6657, opened 2022-03-02
fix: disable cache when not used

RakowskiPrzemyslaw:
This ensures that cache is enabled before going to cache.

## Description
with this, app doesn't break when there are uses useCache in theme, because it will return early


kamnog (approved):

RakowskiPrzemyslaw:
closing this, wrong branch. here is to develop https://github.com/vuestorefront/vue-storefront/pull/6657

PR #6672, opened 2022-03-14
fix: get api host from env variables

dawid-ziobro:
## Description

Allows for setting api host from environmental variables.

## Related Issue

CT-208

## Types of changes

## Checklist:

#### Changelog

#### Tests

#### Code standards

#### Docs

### DIFF
--- a/packages/core/core/src/utils/nuxt/_proxyUtils.ts
+++ b/packages/core/core/src/utils/nuxt/_proxyUtils.ts
@@ -9,7 +9,7 @@ interface CreateProxiedApiParams {
   tag: string;
 }
 
-export const getBaseUrl = (req: IncomingMessage, basePath: string | undefined = '/', secureApiHost: string | undefined): string => {
+export const getBaseUrl = (req: IncomingMessage, basePath: string | undefined = '/', secureApiHost?: string): string => {
   if (!req) return `${basePath}api/`;
   const { headers } = req;
   const isHttps = require('is-https')(req);
@@ -39,7 +39,7 @@ export const getIntegrationConfig = (context: NuxtContext, configuration: any) =
   const cookie = getCookies(context);
   const initialConfig = merge({
     axios: {
-      baseURL: getBaseUrl(context?.req, context?.base, context.env.API_HOST),
+      baseURL: getBaseUrl(context?.req, context?.base, context?.env?.API_HOST),
       headers: {
         ...(cookie ? { cookie } : {})
       }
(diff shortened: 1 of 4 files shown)


packages/core/core/src/utils/nuxt/_proxyUtils.ts
```diff
@@ -14,7 +14,7 @@ export const getBaseUrl = (req: IncomingMessage, basePath: string | undefined =
   const { headers } = req;
   const isHttps = require('is-https')(req);
   const scheme = isHttps ? 'https' : 'http';
-  const host = headers['x-forwarded-host'] || headers.host;
+  const host = process.env.API_HOST || headers['x-forwarded-host'] || headers.host;
```
bloodf, line 17: it would be nice, if we could change from direct access a `env` we can have a custom declaration in the settings file, and then use that as the detection

dawid-ziobro: if you pass this value in setting file then you won't be able to use the same codebase for different domains/subdomains where e.g. are different stores handled

packages/core/docs/getting-started/configuration.md
```diff
@@ -32,6 +32,14 @@ Depending on the integrations used, your project might include some additional c
 
 - [`.lintstagedrc`](https://github.com/okonet/lint-staged#configuration) configures lint runner that enforces code style before every Git commit. By default, it runs ESLint described above.
 
+## Environment variables
+
+Every integration might have its own environmental variables described in documentation for specific integration.
```
lukasborawski, line 37:
```suggestion
Every integration might have its own environmental variables described in the documentation.
```

packages/core/docs/getting-started/configuration.md
```diff
@@ -32,6 +32,14 @@ Depending on the integrations used, your project might include some additional c
...
+
+Those described here are common for every integration.
+
+- `API_HOST` - a host and port of the server that internal API calls are being made, e.g. `example.com` or `localhost:3000`. It is recommended to not leave it empty for security reasons.
```
lukasborawski, line 41:
```suggestion
- `API_HOST` - a host and port of the server that internal API calls are being made, e.g. `example.com` or `localhost:3000`. It is recommended **to not leave it empty** for security reasons.
```

lukasborawski (changes requested):

packages/core/core/src/utils/nuxt/_proxyUtils.ts
bloodf: Gotacha!

So I don't see a problem.

PR #6699, opened 2022-04-06
docs: update project-structure.md

knimavat:
The language section URL was redirecting to localhost instead of the Live site documentation, So I have updated the url to the proper page


filipsobol (approved):

filipsobol:
> The language section URL was redirecting to localhost instead of the Live site documentation, So I have updated the url to the proper page

Whoops. Thank you for your contribution, we really appreciate it 😊

PR #6700, opened 2022-04-07
docs: remove Spryker from integrations list

filipsobol:


Mateuszpietrusinski (approved):
Approve, in case of questions about that decision please contact me

PR #6717, opened 2022-05-09
fix(core): add option to disable the locale auto-redirect in i18n-cookies plugin

MarcinSulowski:
## Description
Added `autoRedirectByLocale` option in the i18n configuration object in nuxt.config.js to allow for bypassing the auto-redirect mechanism of the i18n-cookies plugin.

## Related Issue
[CT-192](https://vsf.atlassian.net/browse/CT-192)

## Motivation and Context
Fixes a scenario in which personal language preference set for a user's browser is different than the default locale (e.g. browser prefers `en` and the default locale is `de`) and results in an automatic redirect to the locale preferred by the browser.

## How Has This Been Tested?
Tested locally with a generic commercetools project.

## Types of changes

## Checklist:

#### Changelog

#### Tests

#### Code standards

#### Docs
- [x] My change requires a change to the documentation.

### DIFF
--- a/packages/core/nuxt-module/plugins/i18n-cookies.js
+++ b/packages/core/nuxt-module/plugins/i18n-cookies.js
@@ -6,6 +6,7 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
   const navigator = isServer ? undefined : window.navigator;
   const acceptedLanguage = app.context.req?.headers?.['accept-language'] || navigator?.languages || '';
   const isRouteValid = !!app.context.route.name;
+  const autoRedirectByLocale = i18nOptions.autoRedirectByLocale ?? true;
   const cookieNames = {
     currency: i18nOptions.cookies?.currencyCookieName || VSF_CURRENCY_COOKIE,
     country: i18nOptions.cookies?.countryCookieName || VSF_COUNTRY_COOKIE,
@@ -47,7 +48,7 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
       ...(autoChangeCookie.currency && { [cookieNames.currency]: getCurrencyByLocale(targetLocale) })
     };
 
-    if (redirectPath) {
+    if (autoRedirectByLocale && redirectPath) {
       redirect(redirectPath);
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/nuxt-module/plugins/i18n-cookies.js
```diff
@@ -6,6 +6,7 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
   const navigator = isServer ? undefined : window.navigator;
   const acceptedLanguage = app.context.req?.headers?.['accept-language'] || navigator?.languages || '';
   const isRouteValid = !!app.context.route.name;
+  const autoRedirectByLocale = i18nOptions.autoRedirectByLocale ?? true;
```
lukasborawski, line 9:
```suggestion
  const autoRedirectByLocale = i18nOptions.autoRedirectByLocale;
```

lukasborawski (changes requested):

packages/core/nuxt-module/plugins/i18n-cookies.js
dawid-ziobro: I'm afraid that this change would implement BC as
```
!!undefined = false
```
and
```
!!(undefined ?? true) = true
```
and default should be `true`

MarcinSulowski: Yes, the default should be `true`.

`const autoRedirectByLocale = i18nOptions.autoRedirectByLocale;` would actually introduce a breaking change since not having this option defined in nuxt.config.js would result in disabling redirects (which should be on by default).

So not having this option defined in nuxt.config.js should still return true, that's why:
`undefined ?? true = true`

Having it set to false should return false:
`false ?? true = false`

And having it set to true should return true:
`true ?? true = true`

dawid-ziobro (approved):

dawid-ziobro (approved):

packages/core/nuxt-module/plugins/i18n-cookies.js
lukasborawski: OK, understood. Thanks!

lukasborawski (approved):

PR #6752, opened 2022-06-14
feat: added flag allowing to change language without full page reload

dawid-ziobro:
## Description

This PR is introducing a flag that is responsible for disabling full page reload on language change. By default page reload is enabled to not to introduce breaking change.

## Related Issue

CT-300

## Types of changes

## Checklist:

#### Changelog

#### Tests

#### Code standards

#### Docs

### DIFF
--- a/packages/docs/getting-started/internationalization.md
+++ b/packages/docs/getting-started/internationalization.md
@@ -224,3 +224,25 @@ export default {
   },
 };
 ```
+## Page reload on language change
+
+By default, the page is reloading while changing language.
+
+To prevent this, you can set `reloadOnLanguageChange: false`  in the `i18n` configartion object in the `nuxt.config.js`.
+
+```js
+// nuxt.config.js
+
+export default {
+  i18n: {
+    reloadOnLanguageChange: false,
+  },
+};
+```
+
+::: tip
+Even though the page will not reload, this will not make all data to be translated automatically.
+:::
+
+Static strings will be translated out of the box, but for data coming from api it will not happen.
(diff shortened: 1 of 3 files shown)


packages/docs/getting-started/internationalization.md
```diff
@@ -224,3 +224,25 @@ export default {
   },
 };
 ```
+## Page reload on language change
+
+By default, the page is reloading while changing language.
+
+To prevent this, you can set `reloadOnLanguageChange: false`  in the `i18n` configartion object in the `nuxt.config.js`.
```
filipsobol, line 231:
```suggestion
By default, the page is reloading after changing the language. To prevent this, you can set `reloadOnLanguageChange: false`  in the `i18n` configuration object in the `nuxt.config.js`.
```

packages/docs/getting-started/internationalization.md
```diff
@@ -224,3 +224,25 @@ export default {
...
+  },
+};
+```
+
+::: tip
+Even though the page will not reload, this will not make all data to be translated automatically.
+:::
+
+Static strings will be translated out of the box, but for data coming from api it will not happen.
+You need to watch for language change and refetch data with new locale.
```
filipsobol, line 248:
```suggestion
Remember that some data will not be translated without a full page reload. Nuxt will automatically translate static strings, but it won't pull new data from the API. You must watch for language changes and re-fetch data with the new locale.
```

filipsobol (approved):

lukaszjedrasik (approved):

lukasborawski (approved):

PR #6760, opened 2022-06-15
docs: update Composition section

filipsobol:


bloodf (approved):

packages/docs/composition/getters.md
```diff
@@ -1,69 +1,88 @@
...
     });
 
     return {
-      cart,
-      cartTotalItems,
+      userFullName
     };
   }
-}
+};
```
matthewmaribojoc, line 60: Just an idea, but it might be helpful to have code comments that describe the steps. It would put the explanation directly with the associated line.

Open to hearing your thoughts.
```
    const { load, user } = useUser();
    
    // gets the full name from `user`
    const userFullName = computed(() => userGetters.getFullName(user.value));

    // will update `user` if successful
    useFetch(async () => {
      await load();
    });

    // make `userFullName` available in our template
    return {
      userFullName
    };
  

```

packages/docs/composition/error-handling.md
```diff
@@ -1,81 +1,107 @@
...
 
-:::details Where does the error come from?
+Instead, composables catch these errors internally and expose them in a computed property called `errors`.
 
-Inside each async method, we catch thrown errors and save them in the `error` object under the key corresponding to the called method.
+## Anatomy of the `errors` objects in composables
 
-```ts
-const { addItem, error } = useCart();
+The `errors` object in composables stores the last error thrown in each composable method. Its keys match the names of the methods, and values contain an error or `null` if no errors were thrown.
```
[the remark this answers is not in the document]
filipsobol: If the second call is successful, the error is cleared

packages/docs/composition/getters.md
filipsobol: Done in https://github.com/vuestorefront/vue-storefront/pull/6760/commits/5163869dfa088837388d950516ce571c0e18a74f

matthewmaribojoc (approved):

PR #6786, opened 2022-07-21
docs: add example for adding new API endpoints

filipsobol:


packages/docs/integrate/extending-integrations.md
```diff
@@ -48,18 +48,66 @@ To register an extension, add it to the array returned from the `extensions` fun
...
+          extendApiMethods: {
+            async customMethod(context, params) => { /* ... */ }
+          }
+        }
+      ],
+      customQueries: {}
+    }
+  }
+};
+```
```
Baroshem, line 90: This wont solve the issue of users IMO. They have the similar example already. I pasted the exactly working code and I think this kind of code should be also put here as an example. Then need exact copy/paste example as they already have a docs page that explains it and they still do not know how to use it.

filipsobol: The example you prepared is Vendure-specific and will not work in any other integration. That's why this section is more "generic".

packages/docs/integrate/extending-integrations.md
```diff
@@ -48,18 +48,66 @@ To register an extension, add it to the array returned from the `extensions` fun
...
+
+Because this is an abstract example that applies to all integrations, we intentionally used `'{INTEGRATION_NAME}'` as the name of the integration. However, for the sake of example, let's assume it's `sloth`.
+
+In such a case, registering the `customMethod` method in `extendApiMethods` would create a new `/api/sloth/customMethod` endpoint. This method accepts two parameters:
```
matthewmaribojoc, line 94: Because this is an abstract example that applies to all integrations, we intentionally used `{INTEGRATION_NAME}` as the name of the integration. For example, registering `customMethod` in `extendApiMethods` would create a new `/api/{INTEGRATION_NAME}/customMethod` endpoint. 

This method accepts two parameters:

packages/docs/integrate/extending-integrations.md
matthewmaribojoc: I see the purpose of having a "generic" example, but could we add some guidance to point someone to the next steps in terms of calling this endpoint from their app.

Even if it's just a sentence saying to check the specific integration guides to see how to call API endpoints for an integration.

packages/docs/integrate/extending-integrations.md
```diff
@@ -48,18 +48,66 @@ To register an extension, add it to the array returned from the `extensions` fun
 ```js
 module.exports = {
   integrations: {
-    <TAG NAME>: {
-      location: '@<integration-package>/server',
+    '{INTEGRATION_NAME}': {
+      location: '{SERVER_INTEGRATION}',
```
matthewmaribojoc, line 52: can we add a comment saying what `{SERVER_INTEGRATION}` is? We describe INTEGRATION_NAME later, but SERVER_INTEGRATION is never mentioned.

packages/docs/integrate/extending-integrations.md
```diff
@@ -48,18 +48,66 @@ To register an extension, add it to the array returned from the `extensions` fun
...
       ],
       customQueries: {}
     }
   }
 };
 ```
+
+## Example: Adding new API endpoints
+
+You cannot register new API endpoints directly but only through extensions to already registered integrations. Let's follow the examples above to add a new endpoint to the integration.
```
matthewmaribojoc, line 69: To register a new API endpoint, we can register a custom extension and use the `extendApiMethods` property. API endpoints cannot be registered directly.

Let's look at a generic example.

packages/docs/integrate/extending-integrations.md
```diff
@@ -48,18 +48,66 @@ To register an extension, add it to the array returned from the `extensions` fun
...
+  - `config` - integration configuration,
+  - `client` - API client created in `packages/api-client/src/index.server.ts`,
+  - `req` - HTTP request object,
+  - `res` - HTTP response object,
+  - `extensions` - extensions registered within integration,
+  - `customQueries` - custom GraphQL queries registered within integration (used only with GraphQL),
+  - `extendQuery` - helper function for handling custom queries (used only with GraphQL).
+- `params` - parameters passed.
+
+All HTTP requests to such endpoint must be POST, with arguments passed in an array:
```
matthewmaribojoc, line 106: All HTTP requests to endpoints created via `extendApiMethods` must use `POST`, with arguments passed in an array:

PR #6807, opened 2022-09-19
feat!: make on create an async function

bartoszherba:
## Description
The goal of this task was to make the `onCreate` an async function. To achieve that we must introduce BC change. The most intrusive change is that the`apiClientFactory` also becomes an async function.

Because the `onCreate` function is strongly coupled with the factory and lifecycles, there is no other way to make it an async function but to make the whole creation process asynchronous.

As a migration step, all `onCreate` functions in `index.server.ts` in all integrations must be updated to the async function. This is a huge change and will have an impact on all integrations.

## How Has This Been Tested?
I tested that for the Magento integration. Once the `onCreate` function was updated to the async function build and app started to work properly as before without any noticeable bugs or problems.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

#### Changelog

#### Tests

- [x] Default Theme
- [ ] Capybara Theme

#### Code standards

#### Docs

### DIFF
--- a/packages/core/src/factories/apiClientFactory/index.ts
+++ b/packages/core/src/factories/apiClientFactory/index.ts
@@ -11,7 +11,7 @@ import { applyContextToApi } from './context';
 const isFn = (x) => typeof x === 'function';
 
 const apiClientFactory = <ALL_SETTINGS extends ApiClientConfig, ALL_FUNCTIONS>(factoryParams: ApiClientFactoryParams<ALL_SETTINGS, ALL_FUNCTIONS>): ApiClientFactory => {
-  function createApiClient (config: any, customApi: any = {}): ApiInstance {
+  async function createApiClient (config: any, customApi: any = {}): Promise<ApiInstance> {
     const rawExtensions: ApiClientExtension[] = this?.middleware?.extensions || [];
     const lifecycles = Object.values(rawExtensions)
       .filter(ext => isFn(ext.hooks))
@@ -23,7 +23,7 @@ const apiClientFactory = <ALL_SETTINGS extends ApiClientConfig, ALL_FUNCTIONS>(f
       .filter(ext => isFn(ext.beforeCreate))
       .reduce((prev, curr) => curr.beforeCreate({ configuration: prev }), config);
 
-    const settings = factoryParams.onCreate ? factoryParams.onCreate(_config) : { config, client: config.client };
+    const settings = factoryParams.onCreate ? await factoryParams.onCreate(_config) : { config, client: config.client };
 
     Logger.debug('apiClientFactory.create', settings);
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -852,7 +852,7 @@ export interface MiddlewareConfig {
 export interface ApiClientFactoryParams<T, F = any> {
   api: F;
   isProxy?: boolean;
-  onCreate: (config: T, headers?: Record<string, string>) => { config: T; client: any };
+  onCreate: (config: T) => Promise<{ config: T; client: any }>;
   extensions?: ApiClientExtension[];
 }
 
@@ -863,7 +863,7 @@ export interface ApiInstance {
 }
 
 export type CreateApiProxyFn = (givenConfig: any, customApi?: any) => ApiInstance;
-export type CreateApiClientFn = (givenConfig: any, customApi?: any) => ApiInstance;
+export type CreateApiClientFn = (givenConfig: any, customApi?: any) => Promise<ApiInstance>;
 
 export interface ApiClientFactory {
   createApiClient: CreateApiClientFn;
--- a/packages/docs/integrate/integration-guide.md
+++ b/packages/docs/integrate/integration-guide.md
@@ -131,12 +131,14 @@ yarn add axios
 ```
 
 There are two functions you can use to create the client:
-- The `onCreate` is called on every request. You can use it to customize the request, response, or settings.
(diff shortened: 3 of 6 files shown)


packages/core/src/factories/apiClientFactory/index.ts
```diff
@@ -23,7 +23,7 @@ const apiClientFactory = <ALL_SETTINGS extends ApiClientConfig, ALL_FUNCTIONS>(f
       .filter(ext => isFn(ext.beforeCreate))
       .reduce((prev, curr) => curr.beforeCreate({ configuration: prev }), config);
 
-    const settings = factoryParams.onCreate ? factoryParams.onCreate(_config) : { config, client: config.client };
+    const settings = factoryParams.onCreate ? await factoryParams.onCreate(_config) : { config, client: config.client };
```
tlos-vsf, line 26: Here is the change I suggested during our daily, exposing the `req` and `res` properties on `onCreate`:

```suggestion
    const settings = factoryParams.onCreate ? await factoryParams.onCreate(_config, this.middleware) : { config, client: config.client };
```

In general, according to the `ApiClientFactoryParams` type, the `onCreate` function should take 2 params:

```
export interface ApiClientFactoryParams<T, F = any> {
  ...
  onCreate: (config: T, headers?: Record<string, string>) => Promise<{ config: T; client: any }>;
  ...
}
```

Without the above change, `headers` are `undefined` and for the purpose of multistore functionality, we need the `req` to pick up the origin hostname. In the **dev** environment, we also need `res`, as there are problems with CORS, and we need a proper `Access-Control-Allow-Origin` header dynamically generated and appended to the response based on the hostname where the request came from.

tlos-vsf (changes requested):
Great work! Your implementation aligns with my PoC for multistore functionality. There is only one thing missing, which I suggested in the comment below.

packages/core/src/factories/apiClientFactory/index.ts
bartoszherba: From what I understand, we need an optional req and/or res object. 
Can't we use already existing hooks like `beforeCreate` or `hook` function to set/get whatever you need?
The fact that we do not pass headers there is one thing, but passing the whole middleware will make everything even vaguer. I really do not want to do that. Middleware is a singleton for that process and so are the req and res objects on the middleware, both available in hooks. Please, consider such a solution.

tlos-vsf (approved):
Looks good to me. I agree that retrieving `req` and `res` objects through API hooks (namely `beforeCreate`) is a cleaner solution, so I'm dropping my initial suggestion regarding exposing the middleware object to the `onCreate`. Thank you for your work on that issue!

Frodigo (approved):

packages/docs/integrate/integration-guide.md
```diff
@@ -131,29 +131,32 @@ yarn add axios
...
 
-  return {
-    config: settings,
-    client: settings.client
-  };
+  const config = buildConfig(settings);
+
+  return new Promise((resolve) => {
+    resolve({ config, client: settings.client });
+  });
```
Fifciu, line 159:
```suggestion
  return Promise.resolve({ config, client: settings.client });
```

packages/core/src/types.ts
```diff
@@ -852,7 +852,7 @@ export interface MiddlewareConfig {
 export interface ApiClientFactoryParams<T, F = any> {
   api: F;
   isProxy?: boolean;
-  onCreate: (config: T, headers?: Record<string, string>) => { config: T; client: any };
+  onCreate: (config: T, headers?: Record<string, string>) => Promise<{ config: T; client: any }>;
```
Fifciu, line 855: I know it's not a scope of this task. But maybe we should remove `headers` argument as we already changing smth here? I do not see any place where it's used. Do I miss something?

bartoszherba: No, you are not missing anything, it looks like an old code or a dead code. I will remove it then.

packages/docs/integrate/integration-guide.md
bartoszherba: ok, updated

Fifciu (approved):

tlos-vsf (approved):

WojtekTheWebDev (changes requested):
Please give me a day more to review ;)

packages/core/src/types.ts
```diff
@@ -852,7 +852,7 @@ export interface MiddlewareConfig {
 export interface ApiClientFactoryParams<T, F = any> {
   api: F;
   isProxy?: boolean;
-  onCreate: (config: T, headers?: Record<string, string>) => { config: T; client: any };
+  onCreate: (config: T) => Promise<{ config: T; client: any }>;
```
WojtekTheWebDev, line 855: Why are the `headers` removed here?

Fifciu: @WojtekTheWebDev look here: https://github.com/vuestorefront/vue-storefront/pull/6807#discussion_r977422605

bartoszherba: Can we resolve this?

WojtekTheWebDev: Thx for explaining

WojtekTheWebDev (approved):
LGTM, now we need to prepare the release

bartoszherba:
@WojtekTheWebDev I can prepare the release, but do we have a procedure of how to do that?

WojtekTheWebDev:
@bartoszherba I've changed base branch to chore/release-2.8.0-canary so we can merge this one. I'll take care of final release

PR #6815, opened 2022-10-21
fix: integration api call security issue

bartoszherba:
## Description
During debugging another issue with CT env, I found out that it's pretty easy to take down our middleware.e.g. let's get a real endpoint like:
POST https://sapcc-dev.europe-west1.gcp.storefrontcloud.io/api/sapcc/addCartEntry .  Wheresapcc is the integration of our middleware but let's send a request to some not existing integration, replace sapcc with something and the request will take down the node.js app

K8s will restart the app fastly but it shouldn’t work like that.

Here is a code responsible for the issue:

```
  app.post('/:integrationName/:functionName', async (req: Request, res: Response) => {
    const { integrationName, functionName } = req.params as any as RequestParams;
    const { apiClient, configuration, extensions, customQueries, initConfig } = integrations[integrationName];
    const middlewareContext: MiddlewareContext = { req, res, extensions, customQueries };
    const createApiClient = apiClient.createApiClient.bind({ middleware: middlewareContext });
    const apiClientInstance = createApiClient({ ...configuration, ...initConfig });
    const apiFunction = apiClientInstance.api[functionName];
    try {
      const platformResponse = await apiFunction(...req.body);
      res.send(platformResponse);
    } catch (error) {
      res.status(getAgnosticStatusCode(error));
      res.send(error);
    }
  });
```
We are looking for properties in integrations[integrationName] but they might be undefined. We have to add an additional check for that.

The bug happens only if middleware is deployed separately from the Nuxt app. E.g. run it via node middleware.js. I can kill e.g. sapcc-dev instance this way.

## Types of changes

## Checklist:

#### Changelog

#### Tests

- [ ] Default Theme
- [ ] Capybara Theme

#### Code standards

#### Docs


packages/middleware/src/createServer.ts
```diff
@@ -50,11 +52,33 @@ async function createServer(config: MiddlewareConfig): Promise<Express> {
...
+        throw new Error(errMsg);
+      }
+
+      ({ apiClient, configuration, extensions, customQueries, initConfig } = integrations[integrationName]);
+    } catch (error) {
+      res.status(404);
+      res.send(error);
+
+      return app;
+    }
```
Fifciu, line 75:
```suggestion
    if (!integrations[integrationName]) {
        const errMsg = `"${integrationName}" integration is not configured. Please, check the request path or integration configuration.` +
          `Request host: ${req?.headers?.host}, origin: ${req.headers.origin}`;
        res.status(404);
        res.send(errMsg);
        return;
    }
    
    const { apiClient, configuration, extensions, customQueries, initConfig } = integrations[integrationName];
```

1. What about doing it this way? WDYT?
2. I understand you might want to check also for properties of `integrations[integrationName]` it might be redundant as `registerIntegrations` is supposed to return a certain shape of data anyway. 
3. I wonder why we return `app`, I see we've already had smth like this at the end of the function. Do you know maybe?

I feel like we could also prepare an integration test for this part of the code. It would be smth like [that](https://github.com/vuestorefront/elasticpath-commerce-cloud/pull/271/files) but instead of mocking EPCC, we would mock the package. TBH, I find it interesting and could cover this PR with an integration test in additional PR.

Fifciu (changes requested):

packages/middleware/src/createServer.ts
bartoszherba: 1. I did it on purpose, if destructuring will fall because one or more attributes are missing it will throw a verbose error. On the other hand, if you think that we can safely check only the integration then I am ok with the simplified solution.
2. This is basically the same question as 1.
3. IDK, I just did it to align with the rest and keep as small PR as possible (to avoid mistakes and bugs) It is worth checking just not in this PR maybe?
4. I would like to try the integration test on my own if you don't mind. :)

Fifciu: 1-2. Well... Look at the code of [`registerIntegrations`](https://github.com/vuestorefront/vue-storefront/blob/main/packages/middleware/src/integrations.ts#L71). It should always contain these fields. If a method like `resolveDependency` fails, it will throw an error anyway. If it wouldn't `{ apiClient, configuration, extensions, customQueries, initConfig }` would exist but with nullish value. That's why I think a simplified solution is better.

3. I agree
4. I agree. BTW I think these tests will increase our confidence here a lot. Don't hesitate to hit me for support if you need it. BTW in previous ones I used `nock` for intercepting and mocking HTTP traffic. However, now I would probably go for Mock Service Worker.

bartoszherba: Ok, I simplified this code.

packages/middleware/src/createServer.ts
```diff
@@ -50,11 +52,24 @@ async function createServer(config: MiddlewareConfig): Promise<Express> {
 
   app.post('/:integrationName/:functionName', async (req: Request, res: Response) => {
     const { integrationName, functionName } = req.params as any as RequestParams;
+
+    if (!integrations[integrationName]) {
+      const errMsg = `"${integrationName}" integration is not configured. Please, check the request path or integration configuration.` +
+        `Request host: ${req?.headers?.host}, origin: ${req.headers.origin}`;
```
Fifciu, line 58: 1. Why do we print host and origin headers? 
2. Why for `.host` we are using optional chaining but for `.origin` we don't?

bartoszherba: I did this because recently we had problems on the cloud with identifying the source of a malicious request (auth0 integration request). I can get rid of this part if you think it is not required. I would also expect to have it on logs on Nginx proxy, and as far as I know, the only reason we do not have it is that something in the cloud was upgraded and it breaks the logging mechanism. Ok, you no what, I will remove it. Introducing something because something we expect to work is not working does not make sense here.,

Fifciu (approved):
Thanks

packages/middleware/src/integrations.ts
```diff
@@ -29,7 +29,10 @@ type IntegrationsLoaded = Record<string, IntegrationLoaded>;
  */
 function resolveDependency<T>(name: string): T {
   try {
+    console.log('name', name);
```
WojtekTheWebDev, line 32: Console.log leftovers

WojtekTheWebDev (approved):
GJ @bartoszherba, thanks for adding integrations test here. I'm leaving my approve in advance but remove console logs 😄

packages/middleware/src/integrations.ts
WojtekTheWebDev: I removed them, resolving

PR #6817, opened 2022-10-28
feat(cli): added init command alias

bloodf:
## Description

added an alias command `init` mapped to `generate store` to the CLI to maintain consistency with old versions

## Motivation and Context

there is a lot of merchandise and printed materials with this alias command.

## How Has This Been Tested?
Local CLI tests and Jest

## Types of changes

## Checklist:

#### Changelog

#### Tests
- [x] I have written test cases for my code
- [x] I have added tests to cover my changes.

- [ ] Default Theme
- [ ] Capybara Theme

#### Code standards

#### Docs


packages/cli/src/commands/init.ts
```diff
@@ -0,0 +1,17 @@
...
+  static override description = t('command.generate_store.description');
+
+  static override examples = ['<%= config.bin %> <%= command.id %>'];
+
+  static override flags = {};
+
+  public async run(): Promise<void> {
+    const generateStore = new GenerateStore(this.argv, this.config);
+
+    return generateStore.run.bind(this).call(this);
```
filipsobol, line 15: I don't think we need both `bind` and `call`, right?

```suggestion
    return generateStore.run.call(this);
```

bloodf: It's a Class over another class, if we call with this, wont bind the this which were added as args :)

filipsobol (approved):

PR #6818, opened 2022-10-28
fix: passing x-forwarded-host in axios request

Fifciu:
Closed and moved to: https://github.com/vuestorefront/vue-storefront/pull/6869

### DIFF
--- a/packages/core/src/utils/nuxt/_proxyUtils.ts
+++ b/packages/core/src/utils/nuxt/_proxyUtils.ts
@@ -33,18 +33,16 @@ export const getIntegrationConfig = (context: NuxtContext, configuration: any) =
     Logger.info('Applied middlewareUrl as ', context.$config.middlewareUrl);
   }
 
+  const headers = {};
+  cookie && Object.assign(headers, { cookie });
+  const reqHeaders = context.req?.headers;
+  reqHeaders?.['x-forwarded-host'] && Object.assign(headers, { 'X-Forwarded-Host': reqHeaders['x-forwarded-host'] });
+  reqHeaders?.host && Object.assign(headers, { Host: reqHeaders.host });
+
   return merge({
     axios: {
       baseURL: new URL(/\/api\//gi.test(baseURL) ? '' : 'api', baseURL).toString(),
-      headers: {
-        ...(cookie ? { cookie } : {}),
-        ...(context.req ? {
-          ...(context.req.headers?.['x-forwarded-host']
-            ? { 'X-Forwarded-Host': context.req.headers['x-forwarded-host'] }
-            : {}),
-          Host: context.req.headers.host
-        } : {})
-      }
+      headers
     }
   }, configuration);
 };
--- a/packages/core/src/utils/nuxt/_proxyUtils.ts
+++ b/packages/core/src/utils/nuxt/_proxyUtils.ts
@@ -38,7 +38,7 @@ export const getIntegrationConfig = (context: NuxtContext, configuration: any) =
       baseURL: new URL(/\/api\//gi.test(baseURL) ? '' : 'api', baseURL).toString(),
       headers: {
         ...(cookie ? { cookie } : {}),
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/nuxt/_proxyUtils.ts
```diff
@@ -38,7 +38,10 @@ export const getIntegrationConfig = (context: NuxtContext, configuration: any) =
       baseURL: new URL(/\/api\//gi.test(baseURL) ? '' : 'api', baseURL).toString(),
       headers: {
         ...(cookie ? { cookie } : {}),
-        ...(context.req ? { Host: context.req.headers.host } : {})
+        ...(context.req ? {
+          'X-Forwarded-Host': context.req.headers['x-forwarded-host'],
```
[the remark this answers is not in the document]
Fifciu: We would have `'X-Forwarded-Host': undefined` provided as an argument to the lodash's `merge` which will remove this property: `Source properties that resolve to undefined are skipped if a destination value exists`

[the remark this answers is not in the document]
bartoszherba: 👌

bartoszherba (approved):

packages/core/src/utils/nuxt/_proxyUtils.ts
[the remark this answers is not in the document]
Fifciu: You know what... it works here well in unit test without mocking `merge`. But when I applied the same in EPCC it didn't work. Maybe it's caused be some versions mismatch. IDK. But I decided to add x-forwarded-host only if it's present

packages/core/src/utils/nuxt/_proxyUtils.ts
```diff
@@ -38,7 +38,12 @@ export const getIntegrationConfig = (context: NuxtContext, configuration: any) =
       baseURL: new URL(/\/api\//gi.test(baseURL) ? '' : 'api', baseURL).toString(),
       headers: {
         ...(cookie ? { cookie } : {}),
-        ...(context.req ? { Host: context.req.headers.host } : {})
+        ...(context.req ? {
```
lsliwaradioluz, line 41: These ternaries with destructuring terrify me. I think we could make it more readable... It's hard to tell what is going on there at first sight.

Approved, but please, consider my comment.

lsliwaradioluz (approved):

packages/core/src/utils/nuxt/_proxyUtils.ts
jeffpdotone: Could we do it like this instead?

```typescript
 const resolvedHost = context?.req?.headers?.['x-forwarded-host'] || context?.req?.headers?.host;

  return merge({
    axios: {
      baseURL: new URL(/\/api\//gi.test(baseURL) ? '' : 'api', baseURL).toString(),
      headers: {
        ...(cookie ? { cookie } : {}),
        ...(resolvedHost ? { Host: resolvedHost } : {})
      }
    }
  }, configuration);
```

And as I received comment from @WojtekTheWebDev in [PR#677](https://github.com/vuestorefront/bigcommerce/pull/677) we don't need 'x-forwarded-host' and needed to ensure Host is there coming from either x-forwarded-host or directly from host property of request headers.

Fifciu: Well @jeffpdotone, I don't really like the idea of merging two headers into one. If we have 2 headers let's leave them as they are during forwarding or we will end up having confusing code. To be honest, that was the primary goal of this PR.

jeffpdotone: All good then if the purpose is served and fixed the main issue.

jeffpdotone (approved):
All good just need to conflict fix.

lsliwaradioluz (approved):
Now better!

PR #6850, opened 2022-11-29
feat: allow overwriting i18n configuration through plugins

lsliwaradioluz:
## Description

As a developer, I want to be able to change the default currency and locale at runtime. An example might be a multistore setup where I might have a plugin loading the store configuration including default language and currency for the store (which is a pretty common scenario).

Unfortunately, the current implementation of the i18n-cookies plugin only relies on properties provided in `nuxt.config.js`. Therefore, influencing the outcome of this plugin at runtime is rather impossible - not good in a multistore setup where we have a single Vue Storefront application displaying several separate stores with different locale/currency settings.

This PR creates a gateway for Nuxt plugins to add `defaultLocale`, `defaultCurrency` and `defaultCountry` to the `context` object so that the `i18n-cookies` plugin picks these values up and does its job the usual way. An example of a flow supported by this PR could be:

1. Plugin A loads store configuration from Commerce API. The configuration contains default `language` and `currency` for the store. The plugin writes these values to the application `context` object.
2. Plugin b (i18n-cookies) reads these values from the `context`. If it does not find them, it falls back to the configuration defined in `nuxt.config.js`.

## Related Issue

[IN-1576](https://vsf.atlassian.net/browse/IN-1576)

## Types of changes

## Checklist:

#### Changelog

### DIFF
--- a/packages/nuxt-module/plugins/i18n-cookies.js
+++ b/packages/nuxt-module/plugins/i18n-cookies.js
@@ -23,7 +23,6 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
     ...i18nOptions.autoChangeCookie,
   };
 
-
   const getDefaultLocale = () => app.$config.defaultLocale;
   const getDefaultCurrency = () => app.$config.defaultCurrency;
   const getDefaultCountry = () => app.$config.defaultCountry;
--- a/packages/nuxt-module/plugins/i18n-cookies.js
+++ b/packages/nuxt-module/plugins/i18n-cookies.js
@@ -24,8 +24,7 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
   };
 
   const getCurrencyByLocale = (locale) =>
-    app.$config.defaultCurrency
-    || i18n.numberFormats?.[locale]?.currency?.currency
+    i18n.numberFormats?.[locale]?.currency?.currency
     || i18nOptions.currency
     || (i18nOptions.currencies.length && i18nOptions.currencies[0].name);
 
@@ -48,7 +47,7 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
   if (isServer) {
     app.i18n.cookieValues = {
       ...(autoChangeCookie.locale && { [cookieNames.locale]: targetLocale }),
-      ...(autoChangeCookie.currency && { [cookieNames.currency]: getCurrencyByLocale(targetLocale) })
+      ...(autoChangeCookie.currency && { [cookieNames.currency]: app.$config.defaultCurrency || getCurrencyByLocale(targetLocale) })
     };
 
     if (autoRedirectByLocale && redirectPath) {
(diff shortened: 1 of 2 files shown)


packages/nuxt-module/plugins/i18n-cookies.js
```diff
@@ -24,13 +24,14 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
   };
 
   const getCurrencyByLocale = (locale) =>
-    i18n.numberFormats?.[locale]?.currency?.currency
+    app.$config.defaultCurrency
+    || i18n.numberFormats?.[locale]?.currency?.currency
     || i18nOptions.currency
     || (i18nOptions.currencies.length && i18nOptions.currencies[0].name);
```
bartoszherba, line 30: I am not sure if I understand, but this function is supposed to get currency by locale and you added the logic to return default currency regardless of the locale parameter value. This is a surprising behavior. We should think if we want to set default value or locale-correlated value, or both. I mean, even if we want to keep the logic you introduced, the naming convention here is very misleading when it comes to understanding the code. Also I understand that this is a no bc change.

lsliwaradioluz: I think the easiest solution would be to simply remove `byLocale` from the function name.

lsliwaradioluz: WDYT @bartoszherba ?

bartoszherba: Then the parameter for the function is a big unknown. Everyone will expect to get the currency based on that parameter. Maybe it should be resolved to null by default and only in this case, we should pull the default currency.

lsliwaradioluz: In that case, what should make of the previous implementation? Here we also have only one of the three cases leveraging the `locale` param and the remaining two just pick the currency from `i18nOptions` without using the `locale` param at all.

```
  const getCurrencyByLocale = (locale) =>
    i18n.numberFormats?.[locale]?.currency?.currency
    || i18nOptions.currency
    || (i18nOptions.currencies.length && i18nOptions.currencies[0].name);
```

WojtekTheWebDev (approved):

packages/nuxt-module/plugins/i18n-cookies.js
lsliwaradioluz: @bartoszherba I've reverted `getCurrencyByLocale` to its initial state, could you please have a look once more?

packages/nuxt-module/plugins/i18n-cookies.js
```diff
@@ -47,7 +47,7 @@ const i18nCookiesPlugin = ({ $cookies, i18n, app, redirect }) => {
   if (isServer) {
     app.i18n.cookieValues = {
       ...(autoChangeCookie.locale && { [cookieNames.locale]: targetLocale }),
-      ...(autoChangeCookie.currency && { [cookieNames.currency]: getCurrencyByLocale(targetLocale) })
+      ...(autoChangeCookie.currency && { [cookieNames.currency]: app.$config.defaultCurrency || getCurrencyByLocale(targetLocale) })
```
bartoszherba, line 50: You can wrap this in `getDefaultCurrency` functions. It might seem like something that is not required but it will leave us with a more declarative code and there will be one place where default currency is resolved.

bartoszherba (approved):
One more comment, IMO worth to introduce but no something that should block this PR.

PR #6869, opened 2022-12-23
fix: passing x-forwarded-host in axios request

Fifciu:
## Description

Around a month ago, I added functionality that passes the original value of the Host header when sending a request from the Nuxt server to the middleware. This PR adds also an X-Forwarded-Host header as it's essential to use it if the infrastructure has a reverse proxy.

It's the following one with resolved conflicts: https://github.com/vuestorefront/vue-storefront/pull/6818


jeffpdotone (approved):
LGTM

WojtekTheWebDev (approved):
LGTM

PR #6873, opened 2022-12-27
chore: sync develop with main

Fifciu:
chore: sync develop with main


Fifciu:
I had to resolve the conflict locally

PR #6885, opened 2023-01-31
feat: deprecating middleware package

lukasborawski:
## Description
This PR provides info about `middleware` package deprecation.

The new middleware package was moved and active development is under [this](https://github.com/vuestorefront/middleware) repository.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

#### Code standards
- [x] My code follows the code style of this project

### DIFF
--- a/packages/middleware/_package.json
+++ b/packages/middleware/_package.json
@@ -1,3 +1,9 @@
+/**
+  WARNING: Package deprecated (from development).
+  Do not provide any changes here.
+  Please check the https://github.com/vuestorefront/middleware repository,
+  where the current middleware package lives.
+*/
 {
   "name": "@vue-storefront/middleware",
   "version": "2.7.5",
--- a/packages/middleware/rollup.config.js
+++ b/packages/middleware/rollup.config.js
@@ -1,4 +1,4 @@
-import pkg from './package.json';
+import pkg from './_package.json';
 import { generateBaseConfig } from '../rollup.base.config';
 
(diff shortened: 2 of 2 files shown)


packages/middleware/rollup.config.js
```diff
@@ -1,4 +1,4 @@
-import pkg from './package.json';
+import pkg from './_package.json';
```
Fifciu, line 1: Why? Isn't there an option called `deprecated`? Also, we could set it to `private: true`, I mean inside the package.json, so it won't be possible to publish it anymore

packages/middleware/_package.json
```diff
@@ -1,3 +1,9 @@
+/**
+  WARNING: Package deprecated (from development).
+  Do not provide any changes here.
+  Please check the https://github.com/vuestorefront/middleware repository,
+  where the current middleware package lives.
+*/
```
Fifciu, line 6: It's not a proper json file anymore. We shouldn't put comments here

Fifciu: No parser will handle it

Fifciu (changes requested):

packages/middleware/_package.json
lukasborawski: But why, you should not release it or parse it? That's the point.

Mateuszpietrusinski:
Changes require more discussion, closing for now.

PR #6894, opened 2023-03-21
fix: added node version check

skirianov:
Added Node version check to the CLI, since it may run into error while installing dependancies with Node 16+ version.

Fixed some minor issues

## Types of changes

## Checklist:

#### Changelog

#### Tests

- [ ] Default Theme
- [ ] Capybara Theme

#### Code standards

#### Docs

### DIFF
--- a/packages/cli/src/domains/magento2/functions/checkNode.ts
+++ b/packages/cli/src/domains/magento2/functions/checkNode.ts
@@ -5,13 +5,24 @@ import {
 } from './terminalHelpers';
 import { t } from 'i18next';
 
+const checkNodeVersion = (nodeString: string): boolean => {
+  const nodeVersion = nodeString.split('v')[1]?.split('.')[0];
+  const subNodeVersion = nodeString.split('v')[1]?.split('.')[1];
+
+  if (Number(nodeVersion) === 16 && Number(subNodeVersion) >= 13) {
+    return true;
+  }
+
+  return false;
+};
+
 /** Checking if Docker is installed and running on user's machine */
 const checkNode = async (): Promise<void> => {
   const node = spawn('node', ['-v']);
 
   return await new Promise((resolve) => {
     node.stdout.on('data', (data) => {
-      if (!data.toString().includes('v16')) {
+      if (!checkNodeVersion(data.toString())) {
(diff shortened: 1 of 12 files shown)


packages/cli/src/domains/magento2/functions/checkNode.ts
```diff
@@ -0,0 +1,38 @@
+import { spawn } from 'child_process';
+import {
+  logSimpleErrorMessage,
+  logSimpleSuccessMessage
+} from './terminalHelpers';
+import { t } from 'i18next';
+
+const checkNodeVersion = (nodeString: string): boolean => {
```
Igloczek, line 8: Not a fan of this solution, same as and forcing people to use Node.js version that will have [EoL in 5 months](https://endoflife.date/nodejs).

I'd say it would be much better to utilize built-it mechanisms, as described here https://stackoverflow.com/a/29349773/2859687

packages/cli/src/domains/magento2/functions/checkYarn.ts
```diff
@@ -0,0 +1,24 @@
+import { spawn } from 'child_process';
+import { logSimpleErrorMessage, simpleLog } from './terminalHelpers';
+
+/** Checking if Docker is installed and running on user's machine */
+const checkYarn = async (): Promise<void> => {
```
Igloczek, line 5: This can be checked automatically with `engines` in `package.json` same as Node.js version.

packages/cli/src/domains/magento2/functions/checkNode.ts
skirianov: Unfortunately, this is not my decision, out template runs on this version and we have to match it. It does not support version 17 or even 18. So, unless that is fixed, I can't do much

packages/cli/src/domains/magento2/functions/checkYarn.ts
skirianov: I may get it wrong, but I guess that setting it up in package json will affect the running of CLI in general. But I need to check that Yarn exists in order to install dependencies after the project is generated. 

So that user will only run `yarn dev`

Igloczek (approved):

Igloczek (approved):

PR #6960, opened 2023-09-19
feat: add pass headers guide

rohrig:
### ❓ Type of change

- [x ] 📖 Documentation (updates to the documentation, readme or JSDoc annotations)

### 📚 Description

A guide to show integrators how to pass headers from the frontend through to the external services

### 📝 Checklist


docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
...
+  return initSDK<typeof sdkConfig>(sdkConfig);
+};
+
+```
+::tip
+This guide uses a `useHeaders` composable. This composable manages both server-side and client-side headers for you. This is just an example, you can use any method you want to fetch the headers.
+::
+
+#step-6
+**Create a Composable to handle headers**
```
mattmaribojoc, line 155:
```suggestion
**Create a Composable to Handle Headers**
```
consistent capitalization with the other sections

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
...
+
+```javascript
+const { addHeadersToState, headerData } = useHeaders();
+addHeadersToState();
+```
+
+#step-4
+**Setting Up Axios Interceptor**
+
+Before making requests using axios, establish an interceptor to automatically append the desired headers.
```
mattmaribojoc, line 54:
```suggestion
Before making requests using axios, establish an interceptor to automatically append the desired headers.
```

mattmaribojoc (approved):
Looks good, very minor consistency suggestions

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
rohrig: Did you intend to suggest a change?
image

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
+# Pass Custom Headers in Requests
+
+In this tutorial, we'll demonstrate how to set up an interceptor to pass headers when making requests using the axios client inside the Vue Storefront SDK. This is useful when you need to append custom headers or cookies to your API calls.
```
filrak, line 3: Would be good to start such guide with some theoretical use-case scenario

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
...
+::danger
+Do not use createSharedComposable if creating a shared composable that uses the interceptor. This will cause the interceptor to only be created once, which will result in the headers being shared across instances.
+::
+
+## Prerequisites:
+Ensure you have the Vue Storefront SDK set up in your project.
+
+## Notes
+::info
+This guide uses the integrations boilerplate with the nuxt option. The same basic principles apply to other frameworks, but the implementation may vary.
```
filrak, line 16:
```suggestion
This guide uses the [Nuxt Storefront Boilerplate](https://github.com/vuestorefront/storefront-nuxt3-boilerplate). for demonstration purposes. The same basic principles apply to other frameworks using the Vue Storefront stack, but the implementation may slightly vary.
```

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
...
+::
+
+All SDK modules export a client instance of axios. This is the axios instance that is used to make requests. You can access this instance by importing the `client` from the SDK.
```
filrak, line 19:
```suggestion
All SDK modules export a client instance of [axios](https://axios-http.com/docs/intro), a popular HTTP Client. This is the axios instance that is used to make requests. You can access it by importing the `client` from the SDK.
```

filrak, line 19: Please give a short example of how to access it and try to avoid describing code without showing it

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Steps
```
filrak, line 21: Would be great to have some more in-depth explanations of what is happening here and why. Since implementations can vary it's important for end-user to understand WHY these things should happen even more than WHAT should happen. Then they can adjust it to their projects.

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
+# Pass Custom Headers in Requests
+
+In this tutorial, we'll demonstrate how to set up an interceptor to pass headers when making requests using the axios client inside the Vue Storefront SDK. This is useful when you need to append custom headers or cookies to your API calls.
+
+After passing the headers to the axios client, we'll also show you how to pass them to external services using the API Client.
+
+::danger
+Do not use createSharedComposable if creating a shared composable that uses the interceptor. This will cause the interceptor to only be created once, which will result in the headers being shared across instances.
```
filrak, line 8: It's a bit out of context here. A newcomer might have no idea what `createSharedComposable` is and why would he/she even has to use it. Perhaps this danger fits elsewhere?

skirianov (approved):

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
```diff
@@ -0,0 +1,284 @@
...
+::
+
+## Prerequisites:
```
filrak, line 11: Do we need to mention it if we're using Nuxt boilerplate that has SDK?

filrak (changes requested):
Maybe I got something wrong but since we're using Nuxt boilerplate shouldn't the guide focus more on how to modify the existing code for this particular use case rather than explaining how to set up the SDK itself?

filrak (changes requested):
requested some changes

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
rohrig: The guide does not use the nuxt 3 boilerplate. It uses the integrations boilerplate with the nuxt option.

rohrig:
> Maybe I got something wrong but since we're using Nuxt boilerplate shouldn't the guide focus more on how to modify the existing code for this particular use case rather than explaining how to set up the SDK itself?

The guide does not use the nuxt 3 boilerplate. It uses the integrations boilerplate with the nuxt option.

docs/content/5.integrations/2.guide/2.adding-headers-to-request.md
rohrig: I agree it's no longer necessary. We have since removed this composable from our projects.

lsliwaradioluz:
@rohrig, here are my suggestions: https://github.com/vuestorefront/vue-storefront/pull/6996/files

LMK if you need any help with understanding those!

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
```diff
@@ -0,0 +1,263 @@
...
+All SDK modules export a client instance of axios. This is the axios instance that is used to make requests to the Server Middleware. You can access this instance by importing the `client` from the SDK.
+
+## Steps
+::steps
+#step-1
+**Import Required Modules**
+
+```javascript
+import { initSDK, buildModule } from '@vue-storefront/sdk';
+import { client, boilerplateModule, BoilerplateModuleType } from '../../packages/sdk/src';
```
Fifciu, line 27: Is it proper path?

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
```diff
@@ -0,0 +1,263 @@
...
+import useHeaders from 'path-to-useHeaders-file'; // Replace with the correct path to the `useHeaders` composable.
```
Fifciu, line 28: Do we provide it for our storefronts?

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
```diff
@@ -0,0 +1,263 @@
...
+
+  addHeadersToState()
+
+  // If an interceptor is already set, eject it to remove it
+  // this prevents headers being shared across instances
+  if (interceptorId !== null) {
+    client.interceptors.request.eject(interceptorId);
+  }
+
+  interceptorId = client.interceptors.request.use(
```
Fifciu, line 127: I guess it will become flaky and unpredictable when there are many parallel connections to applications and shared client

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
Fifciu: Ah it's below

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Steps
+
+step 1.
+###  **Import Required Modules**
+
+```javascript
+import { initSDK, buildModule } from '@vue-storefront/sdk';
+import { client, boilerplateModule, BoilerplateModuleType } from '../../packages/sdk/src';
+import useHeaders from 'path-to-useHeaders-file'; // Replace with the correct path to the `useHeaders` composable.
```
michaelKurowski, line 28: I don't understand, who does provide the `useHeaders`?

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
```diff
@@ -0,0 +1,263 @@
...
+
+For the time being, there is no simple way to circumvent the issue. In integrations which leverage the Init Function, user-specific headers have to be set up within API Methods (where the final request to the external platform is created).
+::
+
+```javascript
+const devHeaderExtension = {
+  name: 'extension-dev-headers',
+  hooks: (req, res) => {
+    return {
+      afterCreate: ({ configuration }) => {
```
Fifciu, line 235: Why afterCreate instead of beforeCall?

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
Fifciu: Probably we should make it more clear

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
michaelKurowski: It also feels like it's strictly vue specific, is this what we want?

Fifciu (changes requested):
How is this whole guide related to risks of shared client mentioned by @bartoszherba recently? [Namely](https://vsf.atlassian.net/wiki/spaces/INT/pages/315392004/ADR013+-+Addressing+Singleton+Design+Issue+in+SDK+Modules+by+Prioritizing+New+Module+Development).

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    state.value.headerData = { isDevMode: cookieHeader };
+  }
+
+  return {
+    addHeadersToState,
+    ...toRefs(state.value),
+  }
+}
```
michaelKurowski, line 197: Should we provide it ootb, instead of asking user to write it themselves. We even have the code already.

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
michaelKurowski: I see that the customer is responsible for creation of the composable, but my question regarding the vue-centric docs remain.

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
```diff
@@ -0,0 +1,266 @@
...
+All SDK modules export a client instance of axios. This is the axios instance that is used to make requests to the Server Middleware. You can access this instance by importing the `client` from the SDK.
+
+## Steps
+
+step 1.
+###  **Import Required Modules**
+
+```javascript
+import { initSDK, buildModule } from '@vue-storefront/sdk';
+import { client, boilerplateModule, BoilerplateModuleType } from '../../packages/sdk/src';
```
michaelKurowski, line 27: I don't understand, where is the client coming from, is it exposed by every SDK module? Is it coordinated with the unified team, do they expose it as well the same way?

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
lsliwaradioluz: Could you elaborate? Not sure about the dangers if somebody uses the SDK server-side (e.g. in an Express application) but when it comes to the usage inside a storefront, how can SDK client be shared across multiple users running the storefront in their own browser? 🤔

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
lsliwaradioluz: This import path should be less ambigious. But long story short: the client is coming from an SDK module (e.g. sapcc-sdk), every module exports the client. Not sure this is going to be the case in a few weeks/months (considering the ongoing discussion initiated by @bartoszherba) but, for now, that's the state of things.

michaelKurowski: I really thing this should be agreed upon with unified. Consider how confused a customer on unified storefront would be after seeing these docs.

WojtekTheWebDev (changes requested):
exporting `client` as our singleton instance became problematic recently and I think we should discuss it during the refi before releasing this guide

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
lsliwaradioluz: I confirm that's the case.
image

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
rohrig: I agree, if we can create a solution that  framework agnostic that would be ideal. But I think given the differences in how Next and Nuxt handle ssr/state, I'm not sure that's possible. I'm not familiar enough with Next to know. 

For context, the guide was made in response to an integrator's query that was using Nuxt. This started as general enablement and turned into a guide when several more folks asked the same question and all were using Nuxt. 

If it's not possible to provide a framework agnostic solution, we can at least add examples in both support frameworks.

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
rohrig: I think that would be great. But I created the example quickly and I'm not certain there aren't edge cases that need to be addressed. This PR is a bit old, so Unified wasn't on my radar when I wrote the guide and we didn't have a storefront in which I could add it. But now we do, so I think it would indeed be great to have this ootb in Unified.

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
rohrig: This guide sat on the back burner for a while. Given the change in landscape since it's inception I agree there should be a consensus and standard way of doing this. RE: comments about ootb, I think it would be sensible for for the Unified team to implement the boilerplate to allow passing custom headers and allow the end dev to use a one-liner, if possible, to add headers to requests. The guide for an ootb solution would be a separate guide though. 

we still have two scenarios:

1.  Projects using Unified: for those we can have an ootb solution and a separate guide
2. Non-Unified:  this sort of guide will still be necessary.

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
michaelKurowski: I'm ok for both frameworks! :) Just including nuxt only solution felt a bit off

docs/content/3.cookbook/2.pass-custom-headers-and-cookies-to-requests.md
michaelKurowski: @jagoral wdyt

filrak:
@rohrig should I close it or do you want to update it?

rohrig:
closed because it's outdated, thanks for reminding me @filrak

PR #6965, opened 2023-09-25
feat(IN-3538): orchestration

WojtekTheWebDev:
### ❓ Type of change

### 📚 Description

Add orchestration possibility by extending the context with `integrations` and `getIntegration` properties.

### 📝 Checklist


packages/middleware/src/handlers/prepareApiFunction/index.ts
```diff
@@ -29,6 +29,40 @@ export function prepareApiFunction(
...
+          customQueries: innerCustomQueries,
+        };
+
+        const createInnerApiClient = innerApiClient.createApiClient.bind({
+          middleware: innerMiddlewareContext,
+        });
+
+        const apiClientInstance = createInnerApiClient({
+          ...innerConfiguration,
+          ...innerInitConfig,
```
michaelKurowski, line 61: Do we preserve the methods exposed on `context.api` or do we not have to worry about them here?

WojtekTheWebDev: Context.api method will be generated during innerCreateApiClient (same behaviour as with previous createApiClient), so nothing changes here

michaelKurowski (approved):
Its ok from my side but check tests firsts as you have breaking change in types.

packages/middleware/src/createServer.ts
```diff
@@ -14,20 +14,21 @@ import {
...
   TIntegrationContext extends Record<string, IntegrationContext>
 >(config: MiddlewareConfig<TIntegrationContext>): Promise<Express> {
+  const app = express();
+
+  app.use(express.json());
+  app.use(cookieParser());
+  app.use(
+    cors({
+      credentials: true,
+      origin: "http://localhost:3000",
```
Fifciu, line 27: When I look at it, I feel like we should introduce a second optional argument to the createServer, where we could pass e.g. config to the cors middleware. But it's outside of the scope of this task, just a food for thought

Fifciu (approved):
Nitpicks to add to batch and commit, looks good. Great work Wojtek

bartoszherba (commented):
@WojtekTheWebDev Are we going to build changelog based on the changeset messages? Because these messages are meaningless.

WojtekTheWebDev:
@bartoszherba Yeah I know, I'm going to fix it as soon as I'll close the prerelease mode to not do it twice

PR #6985, opened 2023-11-02
fix: run command build / install cli

rohrig:
### ❓ Type of change

- [ x] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📝 Checklist

### DIFF
--- a/packages/cli/src/utils/installDeps.ts
+++ b/packages/cli/src/utils/installDeps.ts
@@ -22,6 +22,7 @@ export const installDeps = async (
     sp.start(picocolors.cyan(entryMessage));
 
     await execa(packageManager, ["install"], options);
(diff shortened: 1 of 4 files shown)


packages/cli/src/utils/installDeps.ts
```diff
@@ -22,6 +22,7 @@ export const installDeps = async (
     sp.start(picocolors.cyan(entryMessage));
 
     await execa(packageManager, ["install"], options);
+    await execa(packageManager, ["install"], { cwd: `${options.cwd}/cli` });
```
skirianov, line 25: Two installs?

skirianov (approved):

packages/cli/src/utils/installDeps.ts
rohrig: yes, the cli is not part of the main packages

PR #7054, opened 2024-02-22
feat: add customizable error handler

WojtekTheWebDev:
### ❓ Type of change

- [x] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

Adds customizable error handler.

### 📝 Checklist


packages/sdk/src/modules/moduleFromEndpoints/types.ts
```diff
@@ -64,6 +64,11 @@ export type HTTPClient = (
   config: HTTPClientConfig
 ) => Promise<any>;
 
+/**
+ * Error handler abstraction.
+ */
+export type ErrorHandler = (error: any) => Promise<any>;
```
jagoral, line 70: I don't think that `error` would be enough to provide a solid error handling. User should be aware what was the method which caused the error, what params were passed etc

packages/sdk/src/modules/moduleFromEndpoints/types.ts
```diff
@@ -125,6 +130,23 @@ export type Options = {
...
+   * Custom error handler for the requests.
+   *
+   * It's optional and it will use the default error handler if it's not provided.
+   *
+   * @example
+   * ```ts
+   * const options: Options = {
+   *   apiUrl: "https://api.example.com",
+   *   errorHandler: (error) => {
+   *     return refreshAndRetry(error);
```
jagoral, line 144: what is `refreshAndRetry`? if this is just an example, I don't think that its possible to retry when just an error is known

WojtekTheWebDev: @jagoral it's possible, we did it in the SAPCC but anyway it's just the example :)

packages/sdk/src/modules/moduleFromEndpoints/types.ts
WojtekTheWebDev: can you list what is the most important for you to have in the error handling? I don't see any objections just want to know what to add.

jagoral: I think we could place here:
- error
- method name
- passed arguments
- httpClient

packages/sdk/src/modules/moduleFromEndpoints/types.ts
```diff
@@ -81,6 +81,38 @@ export type HTTPClient = (
...
+   */
+  error: any;
+  /**
+   * The URL of the HTTP request that resulted in an error.
+   */
+  url: string;
+  /**
+   * The parameters passed to the HTTP request. Can be of any type, hence the use of `any[]`.
+   */
+  params: any[];
```
jagoral, line 99: wdyt about using unknown instead of any?

packages/sdk/src/modules/moduleFromEndpoints/types.ts
```diff
@@ -134,6 +166,32 @@ export type Options = {
...
+   * ```typescript
+   * const options: Options = {
+   *   apiUrl: "https://api.example.com",
+   *   errorHandler: async ({ error, url, params, config, httpClient }) => {
+   *     if (error.status === 401) {
+   *       // Refresh token
+   *       await refreshToken();
+   *       // Retry the request
+   *       return httpClient(url, params, config);
+   *     }
```
jagoral, line 187: I think it might be quite common to do conditions basing on the method name - e.g.

```suggestion
   *     const methodName = url.split('/').at(-1);
   *     if (error.status === 401 && methodName !== "loginCustomer") {
   *       // Refresh token
   *       await refreshToken();
   *       // Retry the request
   *       return httpClient(url, params, config);
   *     }
```

hence I think it would be more convenient to pass also `methodName` in `ErrorHandlerContext`, but leaving it up to you

packages/sdk/src/modules/moduleFromEndpoints/types.ts
WojtekTheWebDev: Added.

packages/sdk/src/modules/moduleFromEndpoints/types.ts
WojtekTheWebDev: Done

packages/sdk/src/modules/moduleFromEndpoints/types.ts
```diff
@@ -134,6 +173,32 @@ export type Options = {
    * Default request config for each request.
    */
   defaultRequestConfig?: IncomingConfig;
+
+  /**
+   * An optional custom error handler for HTTP requests.
+   *
+   * If not provided, errors will be thrown as is.
```
lsliwaradioluz, line 180: For the sake of our API Reference docs, I would keep the main description short (An optional custom error handler for HTTP requests.) and include everything else in @remarks.

```suggestion
   * @remarks
   * If not provided, errors will be thrown as is.
```

Otherwise, we will end up with lengthy texts in the `Description` section

image

lsliwaradioluz (approved):

packages/sdk/src/modules/moduleFromEndpoints/types.ts
WojtekTheWebDev: Done

PR #7056, opened 2024-02-22
fix: issue with custom methods in new module

WojtekTheWebDev:
### ❓ Type of change

- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

Fix the issue with extensions and new module.
Additionally, rename getHttpClient into getRequestSender to avoid conflicts.

### 📝 Checklist


packages/sdk/src/modules/moduleFromEndpoints/utils/getRequestSender.ts
```diff
@@ -1,13 +1,21 @@
...
   ErrorHandler,
+  RequestSender,
 } from "../types";
 
-export const getHTTPClient = (options: Options) => {
+/**
+ * Generates a `RequestSender` function configured according to the provided options.
+ *
+ * @remarks
+ * This function abstracts away the details of constructing request URLs, merging configurations,
```
lsliwaradioluz, line 15: Can we just link to the `RequestSender` interface docs? I feel like you're repeating here what you've just written there

packages/sdk/src/modules/moduleFromEndpoints/types.ts
```diff
@@ -52,32 +56,60 @@ export type ComputedConfig = BaseConfig & {
...
+ * Represents a function type for sending HTTP requests, abstracting the complexity of request configuration.
+ *
+ * @remarks
+ * This type is created via a factory function that configures it with common settings, such as base URLs and default headers.
+ *
+ * It simplifies making HTTP requests by handling URL construction, parameter serialization, and applying default and overridden configurations.
+ */
+export type RequestSender = (
+  /**
+   * Name of the method that was called to make the HTTP request.
```
lsliwaradioluz, line 79:
```suggestion
   * Name of the SDK method that was called to make the HTTP request.
```

lsliwaradioluz (approved):

packages/sdk/src/modules/moduleFromEndpoints/utils/getRequestSender.ts
WojtekTheWebDev: imo this function might be used directly by the developers so it would be good to have it in here as well :)

PR #7074, opened 2024-03-15
fix: throw new SdkHttpClient error from default HTTP Client

WojtekTheWebDev:
### ❓ Type of change

### 📝 Checklist


packages/sdk/src/__tests__/integration/modules/middlewareModule.spec.ts
```diff
@@ -491,4 +495,72 @@ describe("middlewareModule", () => {
...
+            });
+
+            return data;
+          } catch (err: any) {
+            throw new SdkHttpError({
+              statusCode: err?.response?.status || 500,
+              message: err?.response?.data?.message || err.message,
+              cause: err,
+            });
+          }
```
michaelKurowski, line 545: I've thought you wanted to return the error like I've shown you. Did you change your mind or is it an accident?

jagoral (approved):

packages/sdk/src/__tests__/integration/modules/middlewareModule.spec.ts
```diff
@@ -491,4 +495,72 @@ describe("middlewareModule", () => {
...
+        isSpecificSdkHttpError(err, {
+          statusCode: (statusCode) => statusCode === 401,
+        })
+      ).toBe(true);
+      expect(isSdkRequestError(err)).toBe(true);
+      expect(isSdkUnauthorizedError(err)).toBe(true);
+    }
+  });
+
+  it("should throw an error as an instance of SdkHttpError class with axios as a custom http client", async () => {
```
michaelKurowski, line 524: What you're really testing here is 
"should make a method throw if `httpClient` callback throws"
Neither axios nor the error is part of the SDK, instead it's part of the test's code.

packages/sdk/src/__tests__/integration/modules/middlewareModule.spec.ts
```diff
@@ -491,4 +495,72 @@ describe("middlewareModule", () => {
...
+    expect.assertions(6);
+
+    const sdk = initSDK({
+      commerce: buildModule(middlewareModule<Endpoints>, {
+        apiUrl: "http://localhost:8181/commerce",
+        httpClient: async (url, params, config) => {
```
michaelKurowski, line 530: I'm also thinking, it looks like it's `fetch`, not a client. Client suggests that it's an instance of something, instead, it's a fetcher.

michaelKurowski (approved):
I've posted few things that may be helpful, but they're minor and I won't block the PR because of them. I trust your judgement in what to do with them.

packages/sdk/src/__tests__/integration/modules/middlewareModule.spec.ts
WojtekTheWebDev: I abandoned this idea as the response of fetch, axios etc are almost always any and this trick does not work :(

packages/sdk/src/__tests__/integration/modules/middlewareModule.spec.ts
WojtekTheWebDev: Applied.

PR #7104, opened 2024-04-05
feat: configuration options in create server

jagoral:
### 🔗 Linked issue

- [ES-411](https://alokai.atlassian.net/browse/ES-411)

### ❓ Type of change

### 📚 Description

- Added options to `createServer` which allows to customize the `cors`, `cookieParser`, and `bodyParser` middlewares

Context:
The change is related to the _hacky_ cors config customization which we've got atm in templates. Currently, the middleware app is run mostly on port 4000, because Storefront app occupies port 3000 (default port for Next.js), so I could change the  `origin` to accept also `localhost:4000`, however there are also multistores which are run locally and other cases, when the current cors config is enough. The workaround which we do in the templates looks as follows

```ts
  const CORS_MIDDLEWARE_NAME = "corsMiddleware";

  const corsMiddleware = app._router.stack.find(
    (middleware: any) => middleware.name === CORS_MIDDLEWARE_NAME,
  );
  const multistoreUrls = listOfStoresDomains.map((storeUrl) => `https://${storeUrl}`);
  corsMiddleware.handle = cors({
    origin: [
      "https://dev.local",
      "http://localhost:3000",
      "http://localhost:3333",
      ...multistoreUrls,
    ],
    optionsSuccessStatus: 200,
    credentials: true,
  });
```

I propose to instead by default allow cors for all origins, and pass an option to configure it when calling `createServer`.

I've added also an option to pass config for `cookieParser`, and `bodyParser`, but I'm open to exclude it from this PR, because it extends our API layer.

### 📝 Checklist

### DIFF
--- a/packages/middleware/src/createServer.ts
+++ b/packages/middleware/src/createServer.ts
@@ -6,7 +6,12 @@ import express from "express";
 import type { HelmetOptions } from "helmet";
 import helmet from "helmet";
 import { registerIntegrations } from "./integrations";
-import type { Helmet, IntegrationContext, MiddlewareConfig } from "./types";
+import type {
+  Helmet,
+  IntegrationContext,
+  MiddlewareConfig,
+  CreateServerOptions,
+} from "./types";
 import {
   prepareApiFunction,
   prepareErrorHandler,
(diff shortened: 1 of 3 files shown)


packages/middleware/src/createServer.ts
WojtekTheWebDev, line 1: I like the change a lot, but using it with no default options (`options: CreateServerOptions = {}`) may cause some unexpected behaviour in customer's apps (not sure if I'm right). Maybe we could use old setup as default that is overwritten by options?

WojtekTheWebDev (changes requested):

packages/middleware/src/createServer.ts
jagoral: ok, I've reverted to the previous options, but also added `http://localhost:4000`

WojtekTheWebDev (approved):
LGTM now :)

PR #7105, opened 2024-04-05
docs: improve middleware extensions guide

WojtekTheWebDev:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSDoc annotations)

### 📚 Description

Improves middleware extensions guide by describing usage on the frontend.

### 📝 Checklist


docs/content/3.middleware/2.guides/3.extensions.md
jagoral, line 1: I've got a problem with this guide. From one side we use typescript to export the types to the storefront, but from the other side, when we show the middleware config and how to define the extension method, it is completely untyped. ESLint will be yelling at this. I think the reason is that it's not easy to type right now, the middleware config. Maybe it's time to introduce the helpers such as `defineIntegration`, `defineExtension`, etc. which will be exported from the `@vsf-enterprise/<integration>-api` package? Then user will be able to: 

```ts
import { defineExtension } from "@vsf-enterprise/sapcc-api";

export const customExtension = defineExtension({
  name "customExtension",
  extendApiMethods: {
    baseSites: async (context, params) => {
      const { data } = await context.client.get("/basesites");
      return data;
    },
  },
});
```

instead of

```ts
import { ApiClientExtension } from "@vue-storefront/middleware";
import { ApiMethods, SapccIntegrationContext } from "@vsf-enterprise/sapcc-api";

export const customExtension = defineExtension({
  name "customExtension",
  extendApiMethods: {
    baseSites: async (context, params) => {
      const { data } = await context.client.get("/basesites");
      return data;
    },
  },
}) satisfies ApiClientExtension<ApiMethods, SapccIntegrationContext>;
```

PR #7155, opened 2024-05-16
IN-4121 - async hooks / api client

Olbix:
### ❓ Type of change

### 📝 Checklist


packages/middleware/src/handlers/prepareApiFunction/index.ts
```diff
@@ -76,19 +74,14 @@ export function prepareApiFunction(
...
-      );
-    } catch (e) {
-      res.status(404);
-      res.send(e.message);
-
-      return;
-    }
-
-    next();
+    getApiFunction(apiClientInstance, functionName, extensionName)
```
WojtekTheWebDev, line 77: nit pick but I'd like to have async/await syntax in here as well (easier debugging imho), but you'd need to use this lib: https://www.npmjs.com/package/express-async-handler

packages/middleware/__tests__/integration/orchestration/orchestration.spec.ts
WojtekTheWebDev, line 1: Can you make `onCreate` async in `packages/middleware/__tests__/integration/bootstrap/server.ts` to validate async behaviour?

WojtekTheWebDev (changes requested):

packages/middleware/src/handlers/prepareApiFunction/index.ts
```diff
@@ -1,11 +1,10 @@
-import type { RequestHandler } from "express";
 import { IntegrationsLoaded, MiddlewareContext } from "../../types";
 import { getApiFunction } from "./getApiFunction";
 
 export function prepareApiFunction(
   integrations: IntegrationsLoaded
-): RequestHandler {
-  return (req, res, next) => {
+): (req, res, next) => Promise<any> {
+  return async (req, res, next) => {
```
WojtekTheWebDev, line 7: I don't like this hack in here ;p I'd go with https://www.npmjs.com/package/express-async-handler and avoid defining the return type, then express will validate if it returns what's expected.
```suggestion
) {
  return asyncHandler(async (req, res, next) => { ... });
```

packages/middleware/src/apiClientFactory/index.ts
```diff
@@ -20,89 +20,91 @@ const apiClientFactory = <
...
           isFunction(extension?.beforeCreate)
         )
-        .reduce(
-          (configSoFar, extension) =>
-            extension.beforeCreate({ configuration: configSoFar }),
-          config
-        );
+        .reduce(async (configSoFar, extension) => {
+          const resolvedConfig = await configSoFar;
+          return extension.beforeCreate({ configuration: resolvedConfig });
```
WojtekTheWebDev, line 47: await on lifecycles seems fine but I think you missed async behaviour on hooks like `beforeCreate`, `afterCreate`, `beforeCall` and `afterCall`

packages/middleware/src/handlers/prepareApiFunction/index.ts
WojtekTheWebDev: You could also try without the lib and without the Promise<any> in return type

WojtekTheWebDev (changes requested):
Still, missing the integration tests to validate full scenario with extensions.

packages/middleware/__tests__/unit/apiClientFactory/apiClientFactory.spec.ts
```diff
@@ -55,13 +55,15 @@ describe("apiClientFactory", () => {
     const clientConfig = { config: defaultConfig, client: defaultClient };
     const { createApiClient } = apiClientFactory<any, any>(params as any);
 
-    const { settings } = createApiClient(clientConfig);
+    const { settings } = await createApiClient(clientConfig);
+
+    await settings.client;
```
lsliwaradioluz, line 60: why?

packages/middleware/__tests__/unit/apiClientFactory/apiClientFactory.spec.ts
```diff
@@ -76,11 +78,11 @@ describe("apiClientFactory", () => {
     };
 
     const { createApiClient } = apiClientFactory<any, any>(params as any);
-    const extensions = (createApiClient as any)._predefinedExtensions;
+    const extensions = ((await createApiClient) as any)._predefinedExtensions;
```
lsliwaradioluz, line 81: I'd split that
```suggestion
    const apiClient = (await createApiClient) as any;
    const extensions = apiClient._predefinedExtensions;
```

packages/middleware/__tests__/unit/apiClientFactory/apiClientFactory.spec.ts
```diff
@@ -76,11 +78,11 @@ describe("apiClientFactory", () => {
...
 
-    createApiClient.bind({ middleware: { req: null, res: null, extensions } })(
-      {}
-    );
+    await createApiClient.bind({
```
lsliwaradioluz, line 83: Since you're calling this function immediately anyway, perhaps you could replace `.bind() ` with [.call()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call)?

packages/middleware/__tests__/unit/apiClientFactory/apiClientFactory.spec.ts
```diff
@@ -106,7 +108,7 @@ describe("apiClientFactory", () => {
     const { createApiClient } = apiClientFactory<any, any>(params as any);
     const extensions = (createApiClient as any)._predefinedExtensions;
 
-    const apiWithContext: any = createApiClient.bind({
+    const apiWithContext: any = await createApiClient.bind({
```
lsliwaradioluz, line 111: > Since you're calling this function immediately anyway, perhaps you could replace .bind()  with [.call()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call)?

same here

packages/middleware/__tests__/unit/apiClientFactory/factoryParams.spec.ts
lsliwaradioluz, line 1: what is this suite testing, actually? :P

lsliwaradioluz (approved):

packages/middleware/__tests__/integration/apiClientFactory.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  beforeAll(async () => {
+    app = await createServer({
+      integrations: {
+        test_integration: {
+          configuration: {},
+          location: "./__tests__/integration/bootstrap/server",
+          extensions() {
+            return [myExtension as any];
```
WojtekTheWebDev, line 56: why as any?

WojtekTheWebDev (approved):
Approved in advance but type casting to any might be redundant

PR #7157, opened 2024-05-16
Update packages to require Node 18

sethidden:
Upgrades all packages to require node 18

This also moves the storefronts packages that required Node 18 but were previously in a separate workspace


WojtekTheWebDev (approved):

WojtekTheWebDev (changes requested):
CI is failing

WojtekTheWebDev (approved):

PR #7215, opened 2024-07-09
Add Prometheus resource usage metrics for middleware

sethidden:
Sending a GET to /metrics will now output Prometheus metrics implemented by https://github.com/siimon/prom-client/
https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors

- [ ] Expose metrics on separate port
- [ ] Add env var to disable prometheus

### DIFF
--- a/packages/middleware/src/helpers/metrics.ts
+++ b/packages/middleware/src/helpers/metrics.ts
@@ -1,0 +1,25 @@
+import { RequestHandler } from "express";
+import prometheusClient from "prom-client";
+
+const prometheusRegister = new prometheusClient.Registry();
+prometheusRegister.setDefaultLabels({
+  app: "middleware",
+});
+prometheusClient.collectDefaultMetrics({ register: prometheusRegister });
+
+/**
+ * Express API handler that should be registered on the /metrics route
+ * so that Prometheus can periodically call the middleware's /metrics endpoint
+ * and scrape the metrics
+ * @example `app.get('/metrics', prometheusMetricsHandler)`
+ */
+const prometheusMetricsHandler: RequestHandler = async (_, res) => {
+  res.contentType(prometheusRegister.contentType);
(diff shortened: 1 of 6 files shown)


packages/middleware/src/helpers/metrics.ts
```diff
@@ -0,0 +1,25 @@
...
+const prometheusRegister = new prometheusClient.Registry();
+prometheusRegister.setDefaultLabels({
+  app: "middleware",
+});
+prometheusClient.collectDefaultMetrics({ register: prometheusRegister });
+
+/**
+ * Express API handler that should be registered on the /metrics route
+ * so that Prometheus can periodically call the middleware's /metrics endpoint
+ * and scrape the metrics
```
bartoszherba, line 13: What kind of metrics? This endpoint will be open (no token or whatever), anyone can hit it. Is it safe to expose such data?

packages/middleware/src/helpers/metrics.ts
```diff
@@ -0,0 +1,25 @@
+import { RequestHandler } from "express";
+import prometheusClient from "prom-client";
+
+const prometheusRegister = new prometheusClient.Registry();
+prometheusRegister.setDefaultLabels({
+  app: "middleware",
+});
+prometheusClient.collectDefaultMetrics({ register: prometheusRegister });
```
michaelKurowski, line 8: I think default metrics are metrics that are largely already available in console (CPU/Memory) no?

michaelKurowski: What would be useful is to keep average response time per endpoint

michaelKurowski: GC and heap could be useful I suppose, but yeah average response time per endpoint would definitely be useful as well. The page views initiative will also rely on prometheus

PR #7223, opened 2024-07-23
feat(ust-1534): add global state manager

dawid-ziobro:
### 🔗 Linked issue

https://alokai.atlassian.net/browse/UST-1534

### ❓ Type of change

### 📚 Description

This PR is introducing the global state manager as proposed with this RFC -> https://alokai.atlassian.net/wiki/spaces/KB/pages/472678415/Storefront+Contract+State+management

For next.js the `SdkContext.Provider` is now combined with `SfStateProvider` and it's renamed to `AlokaiProvider`. This is not a breaking change, exported name is inside of array so on the storefront side you can use whatever name you want.
This WILL be a breaking change if we'll decide to change the name of exported `createSdkContext` method from the package to `createAlokaiContext`. This would be logical tho.

Name changes require updating docs. The best would be to use versioning of docs. I'll prepare docs changes on a separate branch and we'll release them when we'll be ready for it.

### 📝 Checklist


.changeset/giant-dogs-crash.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vue-storefront/next": major
+"@vue-storefront/nuxt": major
```
dawid-ziobro, line 3: It might be a `minor` for nuxt as there're no breaking changes.

packages/next/src/sdk/client.tsx
```diff
@@ -15,20 +25,24 @@ import { CreateSdkContextReturn, SdkProviderProps } from "./types";
...
  * ```
- * Then use the `SdkProvider` in the root component of your application.
+ * Then use the `AlokaiProvider` in the root component of your application.
  * For Pages Router it would be the `pages/_app.tsx` file,
  * and for the App Router it would be the `app/layout.tsx` file.
- * Finally you can use the `useSdk` in any client component of your application.
+ * Finally, you can use the `useSdk` in any client component of your application.
  * @returns [SdkProvider, useSdk] - The SDK provider and the `useSdk` hook.
  */
 export function createSdkContext<
```
dawid-ziobro, line 36: This function's name could be changed to `createAlokaiContext` yet it will be a breaking change.

.changeset/giant-dogs-crash.md
grixu: We could go with major, and use it to remove deprecated stuff from this module as the support for VSF prefix in runtime config.

packages/next/src/sdk/client.tsx
grixu: Sooner or later we will need to do that, so maybe at this point when we making a lot of changes in Next.js ;)

dawid-ziobro: Yeah, i've already changed that ;)

packages/next/src/state.tsx
```diff
@@ -0,0 +1,136 @@
...
+  useContext,
+  useRef,
+} from "react";
+import { type StoreApi, createStore, useStore } from "zustand";
+
+export interface SfStateProps {
+  currencies: SfCurrency[];
+  currency: SfCurrency;
+  locale: string;
+  locales: string[];
```
grixu, line 20: If we would like to be super-typed - we can create an interface that contains languages in form:
```ts
interface AlokaiLocales {
  en: 'en';
}
```

so user can extend this in his `d.ts` file, and then in `setLocale` TypeScript doesn't allow you to set an language that are not defined

```suggestion
  locale: keyof AlokaiLocales;
  locales: (keyof AlokaiLocales)[];
```

packages/next/src/state.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  return (
+    <SfStateContext.Provider value={stateReference.current}>
+      {children}
+    </SfStateContext.Provider>
+  );
+}
+
+function getSfContext() {
```
grixu, line 77:
```suggestion
function getSfStateContext() {
```

To be more consistent ;)

packages/next/src/state.tsx
dawid-ziobro: Ah yes, good catch. Mateusz Ostafil was pointing that out during RFC session.

dawid-ziobro: Added

packages/next/src/state.tsx
```diff
@@ -0,0 +1,153 @@
...
+  createContext,
+  useContext,
+  useRef,
+} from "react";
+import { createStore, useStore } from "zustand";
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface SfLocales {}
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface SfCart {}
```
dawid-ziobro, line 15: Types cannot be taken from UDL package as those can be extended on the middleware layer. Using generics was also not possible due to the fact that `createContext` cannot be wrapped with a function and it's impossible to pass generics to consts. It's created in line 52 of this file.

.changeset/giant-dogs-crash.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+[ADDED] global state management
```
grixu, line 6: Please add a bit more information about it and mark this release as a breaking change 

**References:**
- https://alokai.atlassian.net/wiki/spaces/KB/pages/335216697/Standardising+release+notes+and+migration+guides
- https://github.com/vuestorefront/vue-storefront/pull/7227/files#diff-7101df80574d26d88450f5116b0651b92b3db3307c397136eda549b939b2ad86

packages/nuxt/src/module.ts
```diff
@@ -28,13 +29,15 @@ export default defineNuxtModule<SdkModuleOptions>({
...
     const projectLayer = nuxt.options._layers[0];
 
     const projectRootResolver = createResolver(projectLayer.config.rootDir);
     const buildDirectoryResolver = createResolver(nuxt.options.buildDir);
     const localResolver = createResolver(import.meta.url);
 
+    await installModule("@pinia/nuxt");
+
     nuxt.options.runtimeConfig.public.alokai = defu(
       nuxt.options.runtimeConfig.public?.vsf as any,
```
grixu, line 42: If we're making breaking change release, please remove this line - we shoudn't support the vsf prefix in our environment variables anymore

grixu: As well as line 49, 50

packages/nuxt/src/module.ts
grixu, line 1: I'm also wondering about the `configKey` — to change it to the `alokai`. As for now, we don't use it so badly — as we migrate to runtime envs, it's hidden in the default storefront config. And user can migrate to v4.1.0 to enable the runtime envs without breaking changes. So, I see this as a good moment to change it along with the similar changes in Next.js

packages/next/src/state.tsx
```diff
@@ -0,0 +1,153 @@
...
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface SfCurrency {}
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface SfCustomer {}
```
grixu, line 19: Nice, I'm only wondering what the `.d.ts` on the Storefront side will look like?

dawid-ziobro: Unfortunately this solution is now changed to use the generics.

packages/next/src/state.tsx
dawid-ziobro: Not applicable anymore. It has been changed.

packages/nuxt/src/module.ts
dawid-ziobro: Changed

packages/nuxt/src/module.ts
dawid-ziobro: Changed

grixu (approved):

FRSgit (approved):

PR #7238, opened 2024-08-22
feat: logger

bartoszherba:
closing https://alokai.atlassian.net/jira/polaris/projects/PROD/ideas/view/1113287?selectedIssue=PROD-91

### ❓ Type of change

### 📝 Checklist


.eslintrc.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
   extends: "@vue-storefront/eslint-config-integrations",
+  rules: {
+    "class-methods-use-this": "off",
```
WojtekTheWebDev, line 4: This is good enough, but I think that we could add it to the `@vue-storefront/eslint-config-integrations` package directly

packages/logger/src/LoggerFactory.ts
```diff
@@ -0,0 +1,16 @@
+import { ConsolaStructuredLogger } from "./ConsolaStructuredLogger";
+import { GCPStructuredLog } from "./structuredLog/GCPStructuredLog";
+import { LoggerOptions } from "./interfaces/LoggerOptions";
+
+type LoggerType = "consola-structured"; // | "winston" | "pino";
+
+export class LoggerFactory {
+  static create(type: LoggerType, options: LoggerOptions = {}) {
+    switch (type) {
```
WojtekTheWebDev, line 9: Can we assume that we have an opinion about the default logger and use it if someone won't pass the type?

WojtekTheWebDev (commented):
eslint can be ignored but we could start a discussion about the default factory result

packages/logger/src/LoggerFactory.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Creates a logger based on the type
+ * Available types:
+ * - consola-gcp (Consola logger with GCP structured log)
+ *
+ * @param type The type of logger to create
+ */
+export class LoggerFactory {
+  static create(
+    type: LoggerType,
+    options: LoggerOptions = {
```
dawid-ziobro, line 19: Maybe we should merge default options with those that are passed through params?

bartoszherba: ok, fixed

packages/logger/__tests__/unit/consolaStructuredLogger.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+describe("createConsolaStructuredLogger", () => {
+  const mockStructuredLog: StructuredLog = {
+    createLog: jest.fn(),
+  };
+
+  const logData: LogData = "Test log message";
+  const metadata: Metadata = { userId: "12345" };
+
+  it("should create a logger with the correct set of methods", () => {
```
Fifciu, line 15: nit pick: I usually find `should` in the description of tests redundant. `It creates`, It logs` would be enough

packages/logger/__tests__/unit/loggerFactory.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { LoggerFactory } from "../../src/LoggerFactory";
+
+describe("LoggerFactory try to create type '$type'", () => {
+  // We can add more in the future
+  it.each([
+    {
+      type: "consola-gcp",
+      expected: "Object",
```
Fifciu, line 8: It doesn't really check if consola-gcp type has been created

packages/logger/__tests__/unit/structuredLog/gcpStructuredLog.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      expected: {
+        message:
+          "(/var/www/apps/storefront-unified-nextjs/.next/server/chunks/689.js:67:9)",
+        severity: "INFO",
+        timestamp: expect.any(String),
+        labels: "other-label",
+      },
+    },
+  ])(
+    "should create a GCP structured log: #ID $id",
```
Fifciu, line 153: I get the idea of using .each. However, I always find it overwhelming, especially here when we have 10 cases. This test was much harder to grasp than others. Imagine looking at it at midnight when you have to fix smth in a hurry, or just 6 months from now. Please consider refactoring this test to a few granular ones with a detailed description.

packages/logger/src/ConsolaStructuredLogger.ts
```diff
@@ -0,0 +1,136 @@
...
+    metadata?: Metadata
+  ): void => {
+    const structuredLogEntry = structuredLog.createLog(
+      logData,
+      options,
+      level,
+      metadata
+    );
+
+    const consolaLevel = mapToConsolaLevel(level);
```
Fifciu, line 78:
```suggestion
    // A few LogLevel's map to the same ConsolaLevel, so we had to operate on ConsolaLevel's to DRY
    const consolaLevel = mapToConsolaLevel(level);
```

This line and switch made me think a minute about "why". Maybe we could add comment or some human-readable object like (I prefer latter, what about you?):
```ts
const CONSOLA_LEVELS = {
  ERROR: 0,
  WARN: 1,
};

// then
case ERROR:
logger.error({ structuredLog: structuredLogEntry });
        break;
```

packages/logger/src/ConsolaStructuredLogger.ts
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  const logger = createConsola({
+    level: levelMap?.[options.level] ?? levelMap.info,
+    reporters: [jsonReporter],
+  });
+
+  if (options.reporters) {
+    logger.setReporters(options.reporters);
+  }
```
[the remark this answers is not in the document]
bartoszherba: ok, addressed both problems

packages/logger/src/structuredLog/GCPStructuredLog.ts
```diff
@@ -0,0 +1,118 @@
...
+    critical: "CRITICAL",
+    error: "ERROR",
+    warning: "WARNING",
+    notice: "NOTICE",
+    info: "INFO",
+    debug: "DEBUG",
+  };
+
+  /**
+   * The function to remove ANSI codes from the log message.
```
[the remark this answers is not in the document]
bartoszherba: fixed

packages/logger/__tests__/unit/consolaStructuredLogger.spec.ts
bartoszherba: ok

packages/logger/__tests__/unit/loggerFactory.spec.ts
bartoszherba: I know, since i changed it from the class => function factory I do not know how to do it without some kind of metadata on the object which will be there only for this test.

packages/logger/__tests__/unit/structuredLog/gcpStructuredLog.spec.ts
bartoszherba: I think that the matrix for a multiple inputs for the same test is something that we should not fight with. I am very used to such tests. Imagine more input data. This is exactly what it is for and the only argument is that from your POV it is less readable. Please, create a ticket in the jest repo :)

packages/logger/src/ConsolaStructuredLogger.ts
bartoszherba: I have added a comment.

packages/logger/__tests__/unit/loggerFactory.spec.ts
Fifciu: Remove it then

filrak (changes requested):
Looking at the examples in the docs I think we should consider extracting logger via `this.logger.x` instead. More explanation in the comments

docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,336 @@
...
+        const logger = getLogger(params);
+        logger.info("smth");
+        return params;
+      },
+    };
+  },
+};
+```
+
+#### Caveat: Accessing logger via closure
```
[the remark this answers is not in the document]
michaelKurowski: I did discuss it with Bartek, I honestly don't know why do we have to have this as a function and the argument that "well we may want to put it in a different place later" doesn't really speak to me. It should either be context or req.
Think about this that way - now user has to import it to even use it, rather than simply access is, so it increases the chances that the person will use `console` instead of our logger.

`this` would be a new convention in our product, so I'm against this.


I won't die on this hill tho, the logger was in a development already for quite some time.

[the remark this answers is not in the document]
filrak: Agree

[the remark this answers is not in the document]
bartoszherba: Firstly, do not use colorized sentences like "well we may want to put it in a different place later". You put us in a place where we do not have arguments, where it is not true. It is neither a quote, neither our position in this case. We have provided a logic arguments and you respond with what you think which is a word against word at best.

1. it is not possible to logger only on the context or req (req is unusable at all and we must use res as only res have locals to store some custom data).
2. in some places only context is available, in some only res
3. in some places, like hooks, which is a great example for you, we can't use neither of them because of race conditions in hooks execution. We can't inject logger into singletons.
4. we have discussion with DX about the approach and we showed alternatives
5. logger is set on:

- res.locals.alokai.logger
- context.logger
- someObject.res
- container.logger (injected into hooks to avoid race conditions issues)

So the user will have to remember all these use cases to be able to get the logger.
Also, if we change the path, it will be a breaking change for all users and for all our packages. 

To address these issues we introduced getLogger function that hide all that complexity which which, in addition, prevents from bc.
Even now your request implies changes in 20 packages at least, and only 20 because we did not add logger everywhere yet.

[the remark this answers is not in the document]
michaelKurowski: I wasn't attempting a strawman, sorry if it looked this way. This was the final argument I remembered from our call after the conversation we've had. Not exactly in this phrasing but the summary of it.
- The fact that req is unusable I don't remember at all, if you said it it wasn't back on our calls. 
- Putting it on res is something completely new to me.
- We already had a discussion that there don't have to be any singletons, we have a full control over what do we pass to parameters of hooks and methods, we can build objects on the fly.

That being said, it doesn't really matter at the end of the day. We would go back and argue in circles again and it is greenlighted already.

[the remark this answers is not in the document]
bartoszherba: `The fact that req is unusable I don't remember at all, if you said it it wasn't back on our calls.` 
You can't remember that because we did not know that at that time.
`We already had a discussion that there don't have to be any singletons, we have a full control over what do we pass to parameters of hooks and methods, we can build objects on the fly.`
But the req/res object is and will be a singleton - and must be a singleton. And the same req/res is used in parallel. We have control of what we pass but we can't pass it into these object in all places (eg. in hooks, we just cant due to condition race).

We have considered `this.logger` but it is far from being a perfect solution. It will not work for arrow functions which is the major of our code, we can't bind this. Also, using `this` in such way is an antipattern.

We can also consider attaching getLogger to the global object - which is also an anti pattern but, at least, it is not "this".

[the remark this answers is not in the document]
Fifciu: @filrak Indeed, it would be pleasant to have it always available via `this`. However, we cannot bind `this` for arrow functions. Example (last 2 lines are most important):
```ts
const obj = {
  value: 'Object',
  regularFunction: function() {
    console.log(this.value);
  },
  arrowFunction: () => {
    console.log(this.value);
  }
};

obj.regularFunction(); // Will print 'Object'
obj.arrowFunction(); // Will print 'undefined' in the browser or error in Node.js

// New binding of `this` with `bind`
const newRegularFunction = obj.regularFunction.bind({ value: 'New Object' });
newRegularFunction(); // Will print 'New Object'

const newArrowFunction = obj.arrowFunction.bind({ value: 'New Object' });
newArrowFunction(); // Will still print 'undefined' in the browser or error in Node.js

```

So this change would require rewriting every handler using logger in every integration/module/extension to be classic non-arrow function. We would also need to have constraint to only use classic functions in some places of our software. 

What's more, it doesn't solve the problem of accessing the wrong logger via closure, e.g:
```ts
const someExtension = {
  name: "some-extension",
  hooks(req, res, alokai) {
    // const hooksLogger = getLogger(alokai);
    const hooksLogger = this.logger;
    hooksLogger.info("hooks");
    return {
      beforeCall(params) {
        // Never access via closure a logger belonging to the hooks factory:
        hooksLogger.info("smth");
        // Instead use own logger of every hook function:
        const logger = this.logger;
        // const logger = getLogger(params);
        logger.info("smth");
        return params.args;
      }
    };
  },
};
```

Which unfortunately, invalidates an idea of passing logger via `this`.

BTW I don't understand mentioning LoggerManager. It's not public thing anymore, it's used only internally. So end-developer won't be overwhelmed by new thing to learn.

@michaelKurowski Yup, we learned a lot more during work on the logger and haven't had a chance yet to tell about the details of our findings. 

> I wasn’t aware that there are places exposed to the user that only have res available, looking at the code now, looks like you probably refer to the after hook?

Bartek used mental shortcut, he meant we have access to either context or req+res pair.

BTW At first, I also thought `after hook` has access to the `response: Response` from express. However, it has value of response.body. 

### Our final approach 
It mostly involves injecting a logger wrapped with proper metadata in every possible place. Every place has its own set of metadata attached to make it resistant to singleton-related issues and race conditions. We couldn't set it in `context.res` and attach metadata on the fly, as it causes issues in case of [parallel calls of multiple handlers in the scope of one request](https://docs.alokai.com/middleware/guides/orchestration#example-1-fetching-custom-product-properties-from-legacy-systems) and would make code really hard to grasp.

I didn't want to be too overwhelming with explanation so I've cut to the chase. If you wish to know more about challenges, tell me, as I am open to writing additional longer comment.

docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,336 @@
...
+  extendApp(params) {
+    const logger = getLogger(params);
+    logger.info("Logged succesfully!");
+  }
+};
+```
+
+### hooks
+
+Logger is available in hooks factory:
```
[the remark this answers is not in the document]
Fifciu: Great idea!

docs/content/3.middleware/2.guides/10.working-with-logger.md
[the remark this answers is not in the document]
filrak: Can you elaborate more on this final approach? 

I get the argument to not use `this` and indeed one less import is not worth refactoring all packages. Though If we can still have it available directly from the function parameters instead of relying on `getLogger` import and having possibility to use it wrong when we don't inject it on every level would be a step forward in reducing complexity. Can we have that?


Also" it's used only internally. So end-developer won't be overwhelmed by new thing to learn.". Wrong. Integration API is also a public API. Customers often write their own integrations with various systems.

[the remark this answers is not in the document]
bartoszherba: I will let @Fifciu to answer it but:
`Though If we can still have it available directly from the function parameters instead of relying on getLogger`
the logger itself is in very different places depending on if this is a hook, extension, or other extensibility place. End user will have to remember where and how to get it for all these places. `getLogger` hides this complexity.
`and having possibility to use it wrong when we don't inject it on every level would be a step forward in reducing complexity.` I do not understand the rest of it; What possibility to use it wrong? Where we do not inject the logger? Could you provide some examples where `getLogger` leads to issues and function-less approach does not?

As I described, here are 3 places where logger can be accessed, depends on the extension point:

res.locals.alokai.logger - if res if available and we do not have to care about race condition
context.logger - if context is available but res is not
alokai.logger - injected into hooks to avoid race conditions issues

`getLogger` not only hides that complexity, protect clients from bc but also help with types inference - it would be a hard or impossible to rework broken typing in middleware in such way that the logger will be properly typed in all these places.

`Also" it's used only internally. So end-developer won't be overwhelmed by new thing to learn.". Wrong. Integration API is also a public API. Customers often write their own integrations with various systems.` This is about something completely different, Filip was referring to some internal things which are irrelevant here. We still use loggerManager internally.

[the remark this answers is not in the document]
bartoszherba: @filrak This is our final list of arguments, if you do not think that these are valid points we will do as you ask.

- Users do not need to remember where to obtain the logger, as it is encapsulated within the function itself.
- The function can provide documentation within a docblock, immediately clarifying the objects it operates on (beyond TypeScript types). This eliminates the need to leave the IDE to refer to external documentation.
- By encapsulating specific logic within the function, we make it easier to modify in the future and protect clients from breaking changes.
- This approach enables us to maintain proper typing, which might not be achievable in our middleware under other circumstances (to be validated).
- Every method of accessing the logger must be documented. When using a function, the docblock and brief in-documentation note provide clear information, ensuring comprehensive documentation.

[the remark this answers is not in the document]
bartoszherba: @filrak @michaelKurowski 
Guys, what is your final decision? Can we use `getLogger` or we should get rid of it from the public interface?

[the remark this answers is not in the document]
filrak: Thanks for giving the arguments, they are convincing.

Frankly I think given your arguments choosing one or the other will result in tradeoffs coming purely from the weakness of our implementation rather than thinking what is the best possible API for this kind of feature, and I'm sure there is an approach that won't have them (some form of context-aware `getLogger` from fn context/params ). Though, if you don't see such solution at first glance lets proceed with your proposal.

docs/content/3.middleware/2.guides/10.logging.md
```diff
@@ -0,0 +1,340 @@
...
+- **info**: Logs general informational messages indicating normal operation.
+- **debug**: The most detailed log level, useful for tracing and debugging the application.
+
+**Example:**
+Setting the log level to `error` ensures only error messages and above (i.e., critical, alert, emergency) will be captured:
+
+```json
+{
+  "logger": {
+    "level": "error"
```
filrak, line 240:
```suggestion
    "level": "error"
```
one last catch, wouldn't `verbosity` be more descriptive as we're talking specifically about log verbosity levels here/

filrak (changes requested):

filrak (approved):

packages/logger/src/interfaces/LoggerInterface.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export type Metadata = {
+  /**
+   * Arbitrary metadata.
+   */
+  [key: string]: any;
+  /**
+   * The error that caused the log entry with some additional information.
+   */
+  troubleshoot?: Record<string, any> | string[];
```
Fifciu, line 12: We've never used it eventually. Let's remove it and add later if needed

Fifciu (changes requested):

docs/content/3.middleware/2.guides/10.logging.md
```diff
@@ -0,0 +1,358 @@
...
+    "includeStackTrace": true
+  }
+}
+```
+
+### Global configuration
+
+To configure the logger globally, add a `logger` object to the top-level configuration of the middleware. This configuration will apply to all integrations and parts of the application unless specifically overridden. Here’s an example of a global configuration where the stack trace is included in the logs, and the log level is set to `debug`:
+
+```js[middleware.config.js]
```
dawid-ziobro, line 315: Example should be wrapped with 
```
export const config = {
```

docs/content/3.middleware/2.guides/10.logging.md
```diff
@@ -0,0 +1,358 @@
...
+This global logger configuration ensures that all logs, regardless of the integration or scope, will follow these settings.
+
+### Per integration configuration
+
+If you need different logging settings for specific integrations, you can configure the logger separately within each integration. The logger configuration should be added at the same level as `location` and `configuration` keys within the integration’s object.
+
+For example, here is how you would configure the logger for an integration with the key `commerce`:
+
+```js[middleware.config.js]
+{
```
dawid-ziobro, line 337: Also here should be wrapped with 
```
export const config = {
```

docs/content/3.middleware/2.guides/10.logging.md
```diff
@@ -0,0 +1,358 @@
...
+
+The middleware application provides a logger instance that automatically attaches metadata related to the scope of each call. By adding this contextual data, the logger makes it significantly easier to trace the origin of logs or errors, simplifying the debugging and monitoring process across the application.
+
+## Using logger
+
+The middleware application provides access to the logger in various parts of the system, such as in extensions and integrations. This flexibility allows you to log important events, errors, and other data throughout the lifecycle of your application.
+
+In this section, we will explore how to access and use the logger in different parts of the application.
+
+The below examples will demonstrate how to use the application logger in a way that ensures that correct metadata is conveyed in the logs. To achieve that we will use `getLogger` function that will extract the logger instance from the passed argument.
```
dawid-ziobro, line 11:
```suggestion
The examples below demonstrate how to use the application logger to ensure that correct metadata is conveyed in the logs. To achieve that, we will use the `getLogger` function to extract the logger instance from the passed argument.
```

docs/content/3.middleware/2.guides/10.logging.md
```diff
@@ -0,0 +1,358 @@
...
+  const logger = getLogger(alokai);
+  logger.info("You can log inside onCreate when creating a new integration");
+
+  return {
+    // ...
+  };
+};
+```
+
+### init
```
dawid-ziobro, line 224:
```suggestion
### Init
```

docs/content/3.middleware/2.guides/10.logging.md
bartoszherba: ok, fixed

docs/content/3.middleware/2.guides/10.logging.md
bartoszherba: fixed

docs/content/3.middleware/2.guides/10.logging.md
bartoszherba: fixed

docs/content/3.middleware/2.guides/10.logging.md
```diff
@@ -0,0 +1,358 @@
...
+
+```json
+{
+  "message": "I've just fetched a user!",
+  "timestamp": "2024-10-22T19:04:18.791Z",
+  "severity": "INFO",
+  "alokai": {
+    "context": "middleware",
+    "scope": {
+      ...
```
dawid-ziobro, line 64:
```suggestion
      // ...
```

docs/content/3.middleware/2.guides/10.logging.md
```diff
@@ -0,0 +1,358 @@
...
+
+```json
+{
+  "message": "Middleware is running!",
+  "timestamp": "2024-10-22T19:04:18.791Z",
+  "severity": "INFO",
+  "alokai": {
+    "context": "middleware",
+    "scope": {
+      ...
```
dawid-ziobro, line 39:
```suggestion
      // ...
```

packages/logger/__tests__/unit/structuredLog/gcpStructuredLog.spec.ts
```diff
@@ -0,0 +1,166 @@
+import { Metadata } from "../../../src/interfaces/LoggerInterface";
+import { LogVerbosity } from "../../../src/interfaces/LogVerbosity";
+import { GCPStructuredLog } from "../../../src/structuredLog/GCPStructuredLog";
+
+// We do not want to serialize the log payload at this point
+process.env.NODE_ENV = "dev";
```
dawid-ziobro, line 6: Shouldn't it be `development`?

packages/middleware/src/integrations/helpers/resolveDependency.ts
```diff
@@ -1,14 +1,14 @@
-import consola from "consola";
+import { AlokaiContainer } from "../../types";
```
dawid-ziobro, line 1:
```suggestion
import type { AlokaiContainer } from "../../types";
```

packages/middleware/src/logger/injectMetadata.ts
```diff
@@ -0,0 +1,83 @@
+import merge from "lodash.merge";
+import type { LoggerInterface } from "@vue-storefront/logger";
+import { AlokaiLocal } from "../types";
```
dawid-ziobro, line 3:
```suggestion
import type { AlokaiLocal } from "../types";
```

packages/middleware/src/types/common.ts
```diff
@@ -1,4 +1,6 @@
 import type { Express, Request, Response } from "express";
+import { LoggerInterface } from "@vue-storefront/logger";
```
dawid-ziobro, line 2:
```suggestion
import type { LoggerInterface } from "@vue-storefront/logger";
```

packages/middleware/src/types/common.ts
```diff
@@ -1,4 +1,6 @@
...
+import { LoggerOptions } from "./config";
```
dawid-ziobro, line 3:
```suggestion
import type { LoggerOptions } from "./config";
```

packages/middleware/src/apiClientFactory/wrapFnWithErrorBoundary.ts
```diff
@@ -0,0 +1,47 @@
+import { LogScope } from "../types";
```
dawid-ziobro, line 1:
```suggestion
import type { LogScope } from "../types";
```

dawid-ziobro (approved):

Fifciu (approved):

.changeset/blue-suits-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vue-storefront/middleware": minor
+---
+
+block overwriting alokai metadata
```
WojtekTheWebDev, line 5: All those changesets messages will appear in the final changelog, you should consolidate them into a single one or just remove those with messages like `rc`.

packages/logger/CHANGELOG.md
```diff
@@ -0,0 +1,38 @@
+# @vue-storefront/logger
+
+## 1.0.0-rc.5
```
WojtekTheWebDev, line 3: Imho no point of having those messages in here as well, before merging into main let's clean it up.

packages/logger/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "changesets:version": "yarn changeset version && YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn install",
+    "changesets:publish": "yarn build && yarn changeset publish"
+  },
+  "packageManager": "yarn@1.22.10",
+  "engines": {
+    "node": ">=18"
+  },
+  "dependencies": {
+    "consola": "^3",
+    "dotenv": "^16"
```
WojtekTheWebDev, line 28: Do you need `dotenv` dependnecy in here? Shouldn't it be a devDep or peerDep?

packages/middleware/CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
 # Change log
 
+## 5.1.0-rc.4
```
WojtekTheWebDev, line 3: Another changelog to cleanup

WojtekTheWebDev (approved):
Minor comments to consider before merging from my side

packages/logger/__tests__/unit/structuredLog/gcpStructuredLog.spec.ts
bartoszherba: it should, fixed

Tofandel:
Since this change I get a `Failed to resolve apiClient or function: Logger instance could not be determined` on all endpoints

PR #7259, opened 2024-09-18
ES-1089 Use correct separator for merging cookies

sethidden:
If cookies come both from api client and defaultRequestConfig (set in
sdk config), the sdk will merge them like Cookie: a=1,b=2 which is
wrong, because according to the spec Cookies must be separated by `;`,
not `,`

### DIFF
--- a/.changeset/forty-kings-tan.md
+++ b/.changeset/forty-kings-tan.md
@@ -2,6 +2,6 @@
 "@vue-storefront/sdk": patch
 ---
 
-[CHANGED] Headers with same name will now have their values merged using the ";' delimiter, rather than ",". This is seen when middleware and defaultRequestConfig set the same header - e.g. `Cookie`. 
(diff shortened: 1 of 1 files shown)


.changeset/forty-kings-tan.md
```diff
@@ -0,0 +1,7 @@
+---
+"@vue-storefront/sdk": patch
+---
+
+[CHANGED] Headers with same name will now have their values merged using the ";' delimiter, rather than ",". This is seen when middleware and defaultRequestConfig set the same header - e.g. `Cookie`.
```
WojtekTheWebDev, line 5:
```suggestion
[FIXED] Headers with same name will now have their values merged using the ";' delimiter, rather than ",". This is seen when middleware and defaultRequestConfig set the same header - e.g. `Cookie`. 
```

WojtekTheWebDev (approved):

PR #7267, opened 2024-10-04
feat: add logger module to the SDK

dawid-ziobro:
This PR is adding a logger module to the SDK. Then it will be used within `@vue-storefront/next` and `@vue-storefront/nuxt` packages.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -19,6 +19,9 @@
     "changesets:version": "yarn changeset version && YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn install",
     "changesets:publish": "yarn run build && yarn changeset publish"
   },
+  "dependencies": {
+    "@vue-storefront/logger": "1.0.0-rc.0"
+  },
   "devDependencies": {
     "@types/jest": "^29.0.3",
     "@types/node": "^18.11.17",
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -20,7 +20,7 @@
     "changesets:publish": "yarn run build && yarn changeset publish"
   },
   "dependencies": {
-    "@vue-storefront/logger": "1.0.0-rc.1"
+    "@vue-storefront/logger": "1.0.0-rc.2"
   },
   "devDependencies": {
     "@types/jest": "^29.0.3",
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -20,7 +20,7 @@
     "changesets:publish": "yarn run build && yarn changeset publish"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


packages/sdk/package.json
```diff
@@ -19,6 +19,9 @@
     "changesets:version": "yarn changeset version && YARN_ENABLE_IMMUTABLE_INSTALLS=false yarn install",
     "changesets:publish": "yarn run build && yarn changeset publish"
   },
+  "dependencies": {
+    "@vue-storefront/logger": "1.0.0-rc.0"
```
dawid-ziobro, line 23: To be updated with rc.1

packages/sdk/src/modules/loggerModule/index.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  LoggerFactory,
+  LoggerOptions,
+  LoggerType,
+} from "@vue-storefront/logger";
+import type { Module } from "../../types";
+
+export const loggerModule = (options?: LoggerOptions) => {
```
bartoszherba, line 8: Do not get logger options directly from the logger itself, loggerOptions are not necessary exactly the same options that we might want to expose via sdk module options, this is just a coincidence at this stage. And also, changing the logger options might break code that is using the module but not necessary the module. If I add some server related options to the logger it will be reflected in the module but it does not make any sense. We need to have separate options type for the module.

dawid-ziobro: Ok that makes sense. I'll add one

packages/sdk/package.json
dawid-ziobro: Updated

packages/sdk/src/modules/loggerModule/index.ts
dawid-ziobro: Changed to what we've discussed.

packages/sdk/src/modules/loggerModule/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  warning(logData: LogData, metadata?: Metadata): void;
+  notice(logData: LogData, metadata?: Metadata): void;
+  info(logData: LogData, metadata?: Metadata): void;
+  debug(logData: LogData, metadata?: Metadata): void;
+}
+
+export type LoggerModuleConfig = Partial<{
+  level: LogLevel;
+  includeStackTrace: boolean;
+  environment: Environment;
```
bartoszherba, line 28: We dont need this.

dawid-ziobro: Yup, removed.

bartoszherba (changes requested):
Do not forget to add changeset!

PR #7269, opened 2024-10-04
feat: middleware logger

Fifciu:


packages/middleware/src/types/config.ts
```diff
@@ -16,12 +16,15 @@ export type Integrations<TIntegrationsContext extends TObject = TObject> = {
     : never;
 };
 
+export type LoggerConfig = unknown;
```
bartoszherba, line 19: Is this a mistake? We know the config interface: 
```
export const loggerConfig = {
  level: LogLevel.Error, //optional
    includeStackTrace: true, //optional,
    [key: string]: any, //optional
      handler: { // optional
        info: ...
        warn: ....
        // and so on
      }
    }
};
```

packages/middleware/src/types/common.ts
```diff
@@ -67,8 +73,13 @@ export interface ApiClientExtension<API = any, CONTEXT = any, CONFIG = any> {
   }: {
     app: Express;
     configuration: any;
+    logger: LoggerInterface;
   }) => Promise<void> | void;
-  hooks?: (req: Request, res: Response) => ApiClientExtensionHooks;
+  hooks?: (
+    req: Request,
+    res: Response,
+    alokai: AlokaiContainer
```
bartoszherba, line 81:
```suggestion
    hooksContext: AlokaiContainer
```

packages/middleware/src/types/common.ts
```diff
@@ -107,6 +119,7 @@ export interface LoadInitConfigProps {
   apiClient: ApiClientFactory;
   integration: Integration;
   tag: string;
+  alokai: AlokaiContainer;
```
bartoszherba, line 122:
```suggestion
  integrationContext: AlokaiContainer;
```
I am not sure ;/ this alokai thing is soooooooo bad ;/ So is the overusing the context name. And I think that there might be a different context already injected? I am not sure what are these all types for eh

packages/middleware/src/types/common.ts
```diff
@@ -9,6 +10,7 @@ import {
 import { WithRequired } from "./index";
 import { ApiClient, ApiClientConfig, ApiClientFactory } from "./server";
 
+export type LoggerFactoryConfig = Parameters<typeof LoggerFactory.create>[1];
```
bartoszherba, line 13: Is this some kind of a weird inference of `LoggerOptions` type which is already exported?
Also, if this is meant to be a public interface then it should not be copied from the logger package (or infered in any way) but must be created manually, even if it will be a c/p, to not couple packages on the user interface level so we can avoid bc in the future if eg. logger options will change.

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+    return this.instances[GLOBAL_MIDDLEWARE_CFG_KEY];
+  }
+}
+
+export function getLogger(
+  source: AlokaiContainer | Response | { res: Response }
+): LoggerInstance {
+  if (!source) {
+    throw new Error("Parameter for getLogger function not provided");
+  }
```
bartoszherba, line 63: How is it possible? source is a required parameter.

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+  if ("logger" in source) {
+    return source.logger;
+  }
+  let base: Response;
+  if ("res" in source) {
+    base = source.res;
+  } else {
+    base = source;
+  }
```
bartoszherba, line 72:
```suggestion
 const base = source?.res ?? source;
```

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+    );
+  }
+
+  // getting reference to global instance of logger
+  public getGlobal() {
+    return this.instances[GLOBAL_MIDDLEWARE_CFG_KEY];
+  }
+}
+
+export function getLogger(
```
bartoszherba, line 58:
```
export function getLogger(
  source: AlokaiContainer | Response | { res: Response }
): LoggerInstance {
  if (source?.logger) {
    return source.logger;
  }

  const base = source.res || source;

  if (base?.locals?.alokai?.logger) {
    return base.locals.alokai.logger;
  }

  throw new Error("Logger instance could not be determined");
}
```

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
+import type { Response } from "express";
+import { LoggerInterface } from "@vue-storefront/logger";
+import type {
+  AlokaiContainer,
+  IntegrationContext,
+  MiddlewareConfig,
+} from "./types";
+
+type LoggerInstance = LoggerInterface;
```
bartoszherba, line 9: what is it for?

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+const GLOBAL_MIDDLEWARE_CFG_KEY = Symbol("GLOBAL_MIDDLEWARE_CFG_KEY");
+export class LoggerManager<
```
bartoszherba, line 11: Add docblock

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+      if (this.hasCustomLoggerConfig(integrationEntry)) {
+        // every integration inside `configuration` object can have own logger too
+        this.instances[integrationName] = buildLogger(
+          integrationEntry.logger as TLoggerConfig
+        );
+      }
+    }
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
bartoszherba, line 39: Update eslint-config-integration package, it is added there, this can be removed

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+  // getting reference to instance of logger by integration tag name
+  public get(integrationName) {
+    return (
+      this.instances[integrationName] ||
+      this.instances[GLOBAL_MIDDLEWARE_CFG_KEY]
+    );
+  }
+
+  // getting reference to global instance of logger
+  public getGlobal() {
```
bartoszherba, line 53: why do we need this?, just make get(integrationName?: string) and keep the rest, if you will not pass an arg it will get the global, there is no point to have a separate public method.

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+import type {
+  AlokaiContainer,
+  IntegrationContext,
+  MiddlewareConfig,
+} from "./types";
+
+type LoggerInstance = LoggerInterface;
+const GLOBAL_MIDDLEWARE_CFG_KEY = Symbol("GLOBAL_MIDDLEWARE_CFG_KEY");
+export class LoggerManager<
+  TLoggerConfig = any,
```
bartoszherba, line 12: Don't any plx

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
+import type { Response } from "express";
+import { LoggerInterface } from "@vue-storefront/logger";
```
bartoszherba, line 2:
```suggestion
import type { LoggerInterface } from "@vue-storefront/logger";
```

packages/middleware/src/integrations/registerIntegrations.ts
```diff
@@ -12,35 +11,55 @@ import type {
...
+              extensionName: name,
+              hookName: "extendApp",
+            },
+          }));
+          await extendApp({ app, configuration, logger });
         }
       }
 
-      consola.success(`- Integration: ${tag} loaded!`);
+      logger.notice(`- Integration: ${tag} loaded!`);
```
bartoszherba, line 62:
```suggestion
      logger.debug(`- Integration: ${tag} loaded!`);
```

packages/middleware/src/integrations/registerIntegrations.ts
```diff
@@ -12,35 +11,55 @@ import type {
...
+        alokai,
+      });
       const configuration = {
         ...integration.configuration,
         integrationName: tag,
       };
 
       for (const { name, extendApp } of extensions) {
-        consola.info(`- Loading: ${tag} extension: ${name}`);
+        logger.info(`- Loading: ${tag} extension: ${name}`);
```
bartoszherba, line 48:
```suggestion
        logger.debug(`- Loading: ${tag} extension: ${name}`);
```

packages/middleware/src/integrations/registerIntegrations.ts
```diff
@@ -12,35 +11,55 @@ import type {
...
 ): Promise<IntegrationsLoaded> {
   return await Object.entries(integrations).reduce(
     async (prevAsync, [tag, integration]) => {
-      consola.info(`- Loading: ${tag} ${integration.location}`);
+      const rawLogger = loggerManager.get(tag);
+      const logger = wrapLogger(rawLogger, () => ({
+        context: "middleware",
+      }));
+      const alokai = { logger };
+      logger.info(`- Loading: ${tag} ${integration.location}`);
```
bartoszherba, line 28:
```suggestion
      logger.debug(`- Loading: ${tag} ${integration.location}`);
```

packages/middleware/src/integrations/registerIntegrations.ts
```diff
@@ -12,35 +11,55 @@ import type {
...
 export async function registerIntegrations(
   app: Express,
-  integrations: Integrations
+  integrations: Integrations,
+  loggerManager: LoggerManager
 ): Promise<IntegrationsLoaded> {
   return await Object.entries(integrations).reduce(
     async (prevAsync, [tag, integration]) => {
-      consola.info(`- Loading: ${tag} ${integration.location}`);
+      const rawLogger = loggerManager.get(tag);
```
bartoszherba, line 23:
```suggestion
      const logger = loggerManager.get(tag);
```
meat can be raw, or data but not an object, what does it mean?

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+  "alert",
+  "critical",
+  "error",
+  "warning",
+  "notice",
+  "info",
+  "debug",
+];
+
+export function wrapLogger(
```
bartoszherba, line 87: injectMetadata? loggerWithMetadata? enhanceLogger?
wrapLogger is very meaningless

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+  logger: LoggerInstance,
+  metadataGetter: (metadata: Record<string, any>) => Record<string, any>
```
bartoszherba, line 89: Why do we need a function? Why can't we simply merge metadata? Isn't it an overengineering? The function produces data that we merge anyway, why not inject the data itself?

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+): LoggerInstance {
+  return new Proxy(logger, {
+    get(target, prop) {
+      if (
+        typeof target[prop] === "function" &&
+        methodsToOverwrite.includes(prop as string)
```
bartoszherba, line 95: We should loop over all methods not just hardcoded set of names, we can't miss something if someone will add new one. This will not work for him.

packages/middleware/src/loggerManager.ts
```diff
@@ -0,0 +1,108 @@
...
+  let base: Response;
+  if ("res" in source) {
+    base = source.res;
+  } else {
+    base = source;
+  }
+  return base.locals.alokai.logger;
+}
+
+const methodsToOverwrite = [
```
bartoszherba, line 76: Kill it with fire, read comment below.

packages/middleware/src/integrations/registerIntegrations.ts
```diff
@@ -12,35 +11,55 @@ import type {
   IntegrationsLoaded,
 } from "../types";
 import { defaultErrorHandler } from "../errors/defaultErrorHandler";
+import { LoggerManager, wrapLogger } from "../loggerManager";
 
 export async function registerIntegrations(
```
bartoszherba, line 16: This function is SO unreadable. Please, extract functions, give them names, split operations into block, make it readable.

packages/middleware/src/integrations/helpers/resolveDependency.ts
```diff
@@ -1,14 +1,14 @@
-import consola from "consola";
+import { AlokaiContainer } from "../../types";
 
 /**
  * Resolves dependencies based on the current working directory, not relative to this package.
  */
-export function resolveDependency<T>(name: string): T {
+export function resolveDependency<T>(name: string, alokai: AlokaiContainer): T {
```
bartoszherba, line 6: We need to figure out the better name, container is used only to not use context name but both works as a random entities container. I would suggest to add scoped names.

packages/middleware/src/integrations/helpers/getInitConfig.ts
```diff
@@ -1,17 +1,20 @@
...
   tag,
   integration,
+  alokai,
 }: LoadInitConfigProps): Promise<TObject> {
   if (isFunction(apiClient?.init)) {
     try {
-      consola.success(`- Integration: ${tag} init function Start!`);
-      const initConfig = await apiClient.init(integration.configuration);
-      consola.success(`- Integration: ${tag} init function Done!`);
+      alokai.logger.notice(`- Integration: ${tag} init function Start!`);
```
bartoszherba, line 12:
```suggestion
      alokai.logger.debug(`- Integration: ${tag} init function Start!`);
```

packages/middleware/src/integrations/helpers/getInitConfig.ts
```diff
@@ -1,17 +1,20 @@
...
+      const initConfig = await apiClient.init(
+        integration.configuration,
+        alokai
+      );
+      alokai.logger.notice(`- Integration: ${tag} init function Done!`);
```
bartoszherba, line 17:
```suggestion
      alokai.logger.debug(`- Integration: ${tag} init function Done!`);
```

packages/middleware/src/handlers/prepareMetadataStorage/index.ts
```diff
@@ -0,0 +1,16 @@
+import type { Response, Request, NextFunction } from "express";
+
+export async function prepareMetadataStorage(
```
bartoszherba, line 3: Explain in docblock and how to populate/add more data.

packages/middleware/src/handlers/prepareLogger/index.ts
```diff
@@ -0,0 +1,24 @@
+import type { Response, Request, NextFunction } from "express";
+import { LoggerManager, wrapLogger } from "../../loggerManager";
+
+export function prepareLogger(loggerManager: LoggerManager) {
+  return function (req: Request, res: Response, next: NextFunction) {
+    if (!res.locals.alokai) {
```
bartoszherba, line 6: it can explode before .alokai, add ??

packages/middleware/src/handlers/prepareLogger/index.ts
```diff
@@ -0,0 +1,24 @@
...
+      res.locals.alokai = {};
+    }
+    const logger = loggerManager.get(req.params.integrationName);
```
bartoszherba, line 9: is it possible to have undefined params? can we make it safer? eg. req?.params?.integrationName ?? ''

it will not hurt to do

packages/middleware/src/handlers/prepareLogger/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    const proxiedLogger = wrapLogger(logger, (metadata) => {
```
bartoszherba, line 10: proxiedLogger is meaningless, 
loggerWithMetadata?

packages/middleware/src/integrations/registerIntegrations.ts
```diff
@@ -12,35 +11,55 @@ import type {
...
   app: Express,
-  integrations: Integrations
+  integrations: Integrations,
+  loggerManager: LoggerManager
 ): Promise<IntegrationsLoaded> {
   return await Object.entries(integrations).reduce(
     async (prevAsync, [tag, integration]) => {
-      consola.info(`- Loading: ${tag} ${integration.location}`);
+      const rawLogger = loggerManager.get(tag);
+      const logger = wrapLogger(rawLogger, () => ({
```
bartoszherba, line 24: loggerWithMetadata?

packages/middleware/src/createServer.ts
```diff
@@ -33,6 +37,15 @@ async function createServer<
   config: MiddlewareConfig<TIntegrationContext>,
   options: CreateServerOptions = {}
 ): Promise<Server> {
+  const loggerManager = new LoggerManager<LoggerFactoryConfig>(
+    config,
+    (loggerConfig) => LoggerFactory.create(LoggerType.ConsolaGcp, loggerConfig)
+  );
+  const rawGlobalLogger = loggerManager.getGlobal();
```
bartoszherba, line 44:
```suggestion
  const logger = loggerManager.getGlobal();
```
What is the point to distinguish it here with a name? It is the same interface and you assume that you know internals of getter function. All you care about should be to get loggerInterface typed object.

packages/middleware/src/createServer.ts
```diff
@@ -33,6 +37,15 @@ async function createServer<
...
+  const globalLogger = wrapLogger(rawGlobalLogger, () => ({
```
bartoszherba, line 45: loggerWithMetadata? enhancedLogger?
globalLogger is bad for 2 reasons:
1.you already have a global logger just with dumb name
2.it does not indicate the ehnancement/metadata

packages/middleware/src/createServer.ts
```diff
@@ -59,22 +72,30 @@ async function createServer<
     (config.helmet && Object.keys(config.helmet).length > 0);
   if (isHelmetEnabled) {
     app.use(helmet(helmetOptions));
-    consola.info("VSF `Helmet` middleware added");
+    globalLogger.info("VSF `Helmet` middleware added");
```
bartoszherba, line 75:
```suggestion
    globalLogger.debug("VSF `Helmet` middleware added");
```

packages/middleware/src/createServer.ts
```diff
@@ -59,22 +72,30 @@ async function createServer<
...
   }
 
-  consola.info("Loading integrations...");
-  const integrations = await registerIntegrations(app, config.integrations);
-  consola.success("Integrations loaded!");
+  globalLogger.info("Loading integrations...");
```
bartoszherba, line 78:
```suggestion
  globalLogger.debug("Loading integrations...");
```

packages/middleware/src/createServer.ts
```diff
@@ -59,22 +72,30 @@ async function createServer<
...
+  const integrations = await registerIntegrations(
+    app,
+    config.integrations,
+    loggerManager
+  );
+  globalLogger.notice("Integrations loaded!");
```
bartoszherba, line 84:
```suggestion
  globalLogger.debug("Integrations loaded!");
```

packages/middleware/src/createServer.ts
```diff
@@ -88,7 +109,7 @@ async function createServer<
 
   const server = http.createServer(app);
   createTerminus(server, createTerminusOptions(options.readinessProbes));
-  consola.success("Middleware created!");
+  globalLogger.notice("Middleware created!");
```
bartoszherba, line 112:
```suggestion
  globalLogger.info("Middleware created!");
```

packages/middleware/src/createServer.ts
```diff
@@ -1,11 +1,13 @@
...
 import cookieParser from "cookie-parser";
 import cors from "cors";
 import express from "express";
 import type { HelmetOptions } from "helmet";
 import helmet from "helmet";
 import http, { Server } from "node:http";
 import { createTerminus } from "@godaddy/terminus";
+import { LoggerFactory, LoggerType } from "@vue-storefront/logger";
+import { LoggerManager, wrapLogger } from "./loggerManager";
+import { LoggerFactoryConfig } from "./types";
```
bartoszherba, line 10:
```suggestion
import type { LoggerFactoryConfig } from "./types";
```

packages/middleware/src/createServer.ts
```diff
@@ -1,11 +1,13 @@
-import consola from "consola";
 import cookieParser from "cookie-parser";
 import cors from "cors";
 import express from "express";
 import type { HelmetOptions } from "helmet";
 import helmet from "helmet";
 import http, { Server } from "node:http";
 import { createTerminus } from "@godaddy/terminus";
+import { LoggerFactory, LoggerType } from "@vue-storefront/logger";
```
bartoszherba, line 8:
```suggestion
import { LoggerFactory, type LoggerType } from "@vue-storefront/logger";
```

packages/middleware/src/createServer.ts
```diff
@@ -1,11 +1,13 @@
...
+import { LoggerManager, wrapLogger } from "./loggerManager";
```
bartoszherba, line 9: I am not sure why wrapLogger is exported from the logger manager, loggerManager must export only the class + types eventually, no more than 1 object/class.

packages/middleware/src/apiClientFactory/index.ts
```diff
@@ -35,15 +49,31 @@ const apiClientFactory = <
...
           )
-          .map(async ({ hooks }) =>
-            hooks(this?.middleware?.req, this?.middleware?.res)
-          )
+          .map(async ({ name, hooks }) => {
+            // Attaching extension related metadata to the logger
+            // We cannot assign it to res.locals as we would end up
+            // with incorrect logger for hook functions (like beforeCreate)
+            // in case of multiple extensions using hooks property
+            const wrappedLogger = wrapLogger(logger, (metadata) => ({
```
bartoszherba, line 64: I do the review down-up... so please, check further comments.

packages/middleware/src/loggerManager.ts
bartoszherba, line 58: It mustn't be exported from the loggerManager. There is another comment about it smewhere. Only one entity must be exported from a file. (+ types eventually)

bartoszherba: Or instead of throw new Error use console.log as a fallback but add log info about the problem

packages/middleware/package.json
```diff
@@ -25,7 +25,8 @@
     "cors": "^2.8.5",
     "express": "^4.18.1",
     "helmet": "^5.1.1",
-    "@godaddy/terminus": "^4.12.1"
+    "@godaddy/terminus": "^4.12.1",
+    "@vue-storefront/logger": "1.0.0-rc.0"
```
bartoszherba, line 29:
```suggestion
    "@vue-storefront/logger": "1.0.0-rc.3"
```

packages/middleware/src/apiClientFactory/applyContextToApi.ts
```diff
@@ -36,6 +36,10 @@ const applyContextToApi = <
         const extendQuery = createExtendQuery(context);
         const transformedArgs = await hooks.before({ callName, args });
         const apiClientContext = { ...context, extendQuery };
+        if ("_extensionName" in fn) {
+          context.res.locals.alokai.metadata.scope.extensionName =
```
bartoszherba, line 40: make it safer, it can explode much faster in the chain

packages/middleware/src/apiClientFactory/index.ts
```diff
@@ -10,9 +11,22 @@ import {
...
+  FnMarkedWithExtensionName,
 } from "../types";
 import { applyContextToApi } from "./applyContextToApi";
 
+/**
+ * Function marking endpoint added or overwritten by extension's extendApiMethod hook
+ * with information about source extension's name
+ */
+function markWithExtensionName(
+  apiMethod: FnMarkedWithExtensionName,
```
bartoszherba, line 23: Isnt FnMarkedWithExtensionName too complex? it is all about adding prop to object (fn is an object). Should we care what do you pass?

packages/middleware/src/types/config.ts
Fifciu: I see that `ConsolaLoggerOptions` is not exported, only `LoggerOptions` is available. Should I use the latter then?

bartoszherba: You should write it with your own hands, We do not want to leak internals of subpackages to the public user interface.

packages/middleware/src/loggerManager.ts
Fifciu: TypeScript won't allow that (accessing `.res`). `"res" in source` is a way of selecting type from types intersection that TypeScript understands. If you really want to make it one liner, it could be:
```ts
const base = "res" in source ? source?.res : source;
return base.locals.alokai.logger;
```

But I also find it a little inelegant. Maybe smth like that?:
```ts
type ContextSubset = { res: Response };
type LoggerSource = AlokaiContainer | Response | ContextSubset;
function isAlokaiContainer(source: LoggerSource): source is AlokaiContainer {
  return "logger" in source;
}
function isContext(source: LoggerSource): source is ContextSubset {
  return "res" in source;
}

export function getLogger(source: LoggerSource): LoggerInterface {
  if (isAlokaiContainer(source)) {
    return source.logger;
  }
  const base = isContext(source) ? source.res : source;
  return base.locals.alokai.logger;
}
```

Fifciu: Also, as a part of other tip, getLogger would be moved to another dedicated file, and only getLogger would be exported. So it's functions won't decrease readability of the logger manager's file.

packages/middleware/src/loggerManager.ts
Fifciu: I prefer an approach with throwing. The idea with fallback requires mapping different severity of our logger's methods to console.log/warn/error etc. The probability of that scenario is near 0% and even if it happens, app won't start, so it won't be deployed and information from error will be visible.

Fifciu: As a next step of [this approach](https://github.com/vuestorefront/vue-storefront/pull/7269#discussion_r1791770263), we could do smth like:
```ts
type ContextSubset = { res: Response };
type LoggerSource = AlokaiContainer | Response | ContextSubset;
function isAlokaiContainer(source: LoggerSource): source is AlokaiContainer {
  return "logger" in source;
}
function isContext(source: LoggerSource): source is ContextSubset {
  return "res" in source;
}
function findLogger(source: LoggerSource): LoggerInterface {
  if (isAlokaiContainer(source)) {
    return source?.logger;
  }
  const base = isContext(source) ? source?.res : source;
  return base?.locals?.alokai?.logger;
}
export function getLogger(source: LoggerSource): LoggerInterface {
  const logger = findLogger(source);
  if (!logger) {
    throw new Error("Logger instance could not be determined");
  }
  return logger;
}
```

packages/middleware/src/loggerManager.ts
Fifciu: Relict of past

packages/middleware/src/loggerManager.ts
Fifciu: Ok

packages/middleware/src/loggerManager.ts
bartoszherba: both code versions are better than initial one so feel free to pick whatever you like more :)

packages/middleware/src/loggerManager.ts
Fifciu: injectMetadata sounds proper. `loggerWithMetadata` could be what's injectMetadata return

packages/middleware/src/loggerManager.ts
Fifciu: Take a look at the following middleware:
```ts
export function prepareLogger(loggerManager: LoggerManager) {
  return function (req: Request, res: Response, next: NextFunction) {
    if (!res.locals.alokai) {
      res.locals.alokai = {};
    }
    const logger = loggerManager.get(req.params.integrationName);
    const proxiedLogger = injectMetadata(logger, (metadata) => {
      return {
        ...res.locals.alokai.metadata,
        ...metadata,
        scope: {
          ...res.locals.alokai.metadata?.scope,
          ...metadata?.scope,
        },
      };
    });
    res.locals.alokai.logger = proxiedLogger;

    next();
  };
}
```

This middleware is the second one that's called when someone calls any of our endpoints (first is creating res.locals.alokai.metadata object). In that moment, `res.locals.alokai.metadata` doesn't have many information. We adjust it's value in different moments of request's flow. If we passed ` ...res.locals.alokai.metadata` here, without getter, then it would be instantly resolved (won't have data added later in request's flow). Getter assures us, we always get actual value of `res.locals.alokai.metadata` (including fields added later like functionName, integrationName, extensionName).

packages/middleware/src/apiClientFactory/index.ts
Fifciu: It's useful. It prevents us from passing primitive values and unlocks only specified fields. We could rename it `CustomEndpointHandler`.

I also considered `object` type as it's exactly for locking primitive values but it blocks `_extensionName` field.

packages/middleware/src/loggerManager.ts
Fifciu: The overwritten method has to match the interface anyway. I've just found an exposed `log` method that doesn't match it. So I reverted logic to specify methods to skip.

```ts

const METHODS_TO_SKIP = ["log"];

export function injectMetadata(
  logger: LoggerInterface,
  metadataGetter: (metadata: Record<string, any>) => Record<string, any>
): LoggerInterface {
  return new Proxy(logger, {
    get(target, prop) {
      const shouldSkipMethod =
        typeof target[prop] !== "function" ||
        METHODS_TO_SKIP.includes(prop as string);

      if (!shouldSkipMethod) {
        return (...args: any[]) => {
          const [message, metadata] = args;
          target[prop](message, {
            ...metadata,
            ...metadataGetter(metadata),
          });
        };
      }
      return target[prop];
    },
  });
}
```

packages/middleware/src/handlers/prepareLogger/index.ts
bartoszherba: But, does it hurt to write safe code? Did you ever had situation that you thought something is not possible and then it happened?

packages/middleware/src/loggerManager.ts
bartoszherba: Yes, this is also what I have proposed in the SDK, to have blacklist rather than whitelist. At some point we might want to expose it as a configurable options then it should be a shorter list than whitelist.

packages/middleware/src/apiClientFactory/applyContextToApi.ts
```diff
@@ -36,6 +38,17 @@ const applyContextToApi = <
         const extendQuery = createExtendQuery(context);
         const transformedArgs = await hooks.before({ callName, args });
         const apiClientContext = { ...context, extendQuery };
+        if (isCustomEndpointHandler(fn)) {
+          if (context.res.locals?.alokai?.metadata?.scope) {
+            context.res.locals.alokai.metadata.scope.extensionName =
+              fn._extensionName;
```
dawid-ziobro, line 44: If `fn` stands for function than it's weird to provide some property to it. Could you pls change the `fn` name to something more accurate?

packages/middleware/src/handlers/prepareMetadataStorage/index.ts
```diff
@@ -0,0 +1,46 @@
...
+ * // or
+ * res.locals.alokai.metadata.myNewField = 'hello';
+ * ```
+ */
+export async function prepareMetadataStorage(
+  req: Request,
+  res: Response,
+  next: NextFunction
+) {
+  if (!res.locals.alokai) {
```
dawid-ziobro, line 38: Is `locals` always available here?

packages/middleware/src/apiClientFactory/applyContextToApi.ts
bartoszherba: There is no better way to achieve this and in js fn is an object and can store properties - and it does in fact. I do no like that patter but considering the fact we are using such language and it will a BC if we will do it other way it is a lesser evil to me.

packages/middleware/src/types/config.ts
Fifciu: It is not exported anymore as I just use 
```ts
import {
  type LoggerOptions,
} from "@vue-storefront/logger";
```

I am not reexporting this type outside of the package. Would you still want me to copy paste it to the package? (LoggerOptions)

packages/middleware/package.json
```diff
@@ -25,7 +25,8 @@
     "cors": "^2.8.5",
     "express": "^4.18.1",
     "helmet": "^5.1.1",
-    "@godaddy/terminus": "^4.12.1"
+    "@godaddy/terminus": "^4.12.1",
+    "@vue-storefront/logger": "1.0.0-rc.3"
```
bartoszherba, line 29: I know it is very picky from my side, but can we move our packages to the top of deps list? Also, I think that the consola can be removed at this stage?

packages/middleware/src/apiClientFactory/applyContextToApi.ts
```diff
@@ -36,6 +38,17 @@ const applyContextToApi = <
...
+          } else {
+            const logger = getLogger(context.res);
+            logger.warning(
```
bartoszherba, line 47: Should be a debug imo, ant the info can be more technical as this is a very meaningless information that will be printed for every request eventually.
```suggestion
            logger.debug(
```

packages/middleware/src/createServer.ts
```diff
@@ -33,6 +40,15 @@ async function createServer<
...
   options: CreateServerOptions = {}
 ): Promise<Server> {
+  const loggerManager = new LoggerManager<LoggerOptions>(
+    config,
+    (loggerConfig) => LoggerFactory.create(LoggerType.ConsolaGcp, loggerConfig)
+  );
+  const logger = loggerManager.get();
+  const loggerWithMetadata = injectMetadata(logger, () => ({
+    context: "middleware",
+  }));
```
bartoszherba, line 50:
```
const loggerWithMetadata = injectMetadata(loggerManager.get(), () => ({
    context: "middleware",
  }));
```

packages/middleware/src/handlers/prepareLogger/index.ts
Fifciu: Ok, let it be like that if someone every creates new route and tries to use it:
```ts
const logger = loggerManager.get(req?.params?.integrationName);
    if (!req?.params?.integrationName) {
      console.error("prepareLogger middleware used for unsupported route");
    }
```

packages/middleware/src/handlers/prepareMetadataStorage/index.ts
Fifciu: Updated to:
```ts
if (!res.locals) {
      res.locals = {};
    }
    if (!res.locals.alokai) {
      res.locals.alokai = {};
    }
```
Just for sure

packages/middleware/src/apiClientFactory/applyContextToApi.ts
Fifciu: It means something is incorrect internally. It decreases the quality of metadata. A client should act so I think warning is perfectly fine, as app is still working but they need to act.

packages/middleware/src/types/common.ts
Fifciu: We've decided together, we will adjust it in next task

packages/middleware/src/apiClientFactory/applyContextToApi.ts
```diff
@@ -1,4 +1,5 @@
 // @ts-check
+import { isCustomEndpointHandler } from "../helpers";
```
bartoszherba, line 2:
```suggestion
import { isExtensionEndpointHandler } from "../helpers";
```
Because we also use extensions which are not a custom code I think - custom = client?

packages/middleware/src/apiClientFactory/applyContextToApi.ts
```diff
@@ -36,6 +38,17 @@ const applyContextToApi = <
...
         const transformedArgs = await hooks.before({ callName, args });
         const apiClientContext = { ...context, extendQuery };
+        if (isCustomEndpointHandler(fn)) {
+          if (context.res.locals?.alokai?.metadata?.scope) {
+            context.res.locals.alokai.metadata.scope.extensionName =
+              fn._extensionName;
+          } else {
+            const logger = getLogger(context.res);
+            logger.warning(
+              "Local Alokai's metadata object is missing. Check your source code or contact Alokai's team."
```
bartoszherba, line 48: maybe
```
Alokai's metadata object is missing in 'res.locals.alokai'. This could impact extension-specific functionality, 
  such as scope tracking. Ensure 'res.locals.alokai.metadata' is properly initialized. 
  Verify middleware setup or contact the Alokai team for assistance.
  Steps to troubleshoot:
  Check if .... bla bla....
```

current message is very unclear

packages/middleware/src/handlers/prepareApiFunction/index.ts
```diff
@@ -16,6 +15,15 @@ export function prepareApiFunction(
       return;
     }
 
+    res.locals.alokai.metadata = {
+      ...res.locals.alokai.metadata,
```
bartoszherba, line 19: can it be possibly empty? assuming that we even added a warning for that previously it can, somehow... can we add .? safeguards with {} fallback?

bartoszherba (approved):
Add changeset, address mentioned issues and fire.

PR #7276, opened 2024-10-08
feat: sdk logger wrapper

dawid-ziobro:
Wrap logger with predefined data.


packages/sdk/src/modules/loggerModule/index.ts
```diff
@@ -1,14 +1,28 @@
...
+import { wrapLogger } from "./wrapper";
+import { isInvalidConfig } from "./utils";
 
 export const loggerModule = (options?: LoggerModuleConfig) => {
   const logger =
     options?.handler ?? LoggerFactory.create(LoggerType.ConsolaGcp, options);
 
+  if (isInvalidConfig(options)) {
+    console.warn(
+      "Both handler and options are provided. Using handler, options will be ignored."
```
Fifciu, line 13:
```suggestion
      "Both handler and options are provided to logger's options. Using handler, options will be ignored."
```

packages/sdk/src/modules/loggerModule/wrapper.ts
```diff
@@ -0,0 +1,27 @@
+import type { LoggerInterface } from "./types";
+
+const methodsToExclude = ["log"];
+
+export function wrapLogger(
```
Fifciu, line 5: We renamed it `injectMetadata` in middleware as it's more meaningful. Please adjust it here.

Fifciu: Generally, it looks like that there:
```ts

const METHODS_TO_SKIP = ["log"];

export function injectMetadata(
  logger: LoggerInterface,
  metadataGetter: (metadata: Record<string, any>) => Record<string, any>
): LoggerInterface {
  return new Proxy(logger, {
    get(target, prop) {
      const shouldSkipMethod =
        typeof target[prop] !== "function" ||
        METHODS_TO_SKIP.includes(prop as string);

      if (!shouldSkipMethod) {
        return (...args: any[]) => {
          const [message, metadata] = args;
          target[prop](message, {
            ...metadata,
            ...metadataGetter(metadata),
          });
        };
      }
      return target[prop];
    },
  });
}
```

Maybe we could copy paste it here at least for now? So same thing won't be different at least now.

packages/sdk/src/modules/loggerModule/index.ts
```diff
@@ -1,14 +1,28 @@
...
+    );
+  }
+
+  const wrappedLogger = wrapLogger(logger, {
```
Fifciu, line 17:
```suggestion
  const loggerWithMetadata = wrapLogger(logger, {
```

packages/sdk/src/modules/loggerModule/index.ts
```diff
@@ -1,14 +1,28 @@
...
+    alokai: {
+      context: "storefront",
+    },
+  });
+
   return {
     connector: {
-      ...logger,
+      ...wrappedLogger,
```
Fifciu, line 25: We can't use spread operator here as wrapLogger/injectMetadata wraps logger in Proxy's instance. And spread destroys Proxy.

Fifciu (changes requested):

packages/sdk/src/modules/loggerModule/index.ts
dawid-ziobro: Added

packages/sdk/src/modules/loggerModule/index.ts
dawid-ziobro: Changed.

packages/sdk/src/modules/loggerModule/index.ts
dawid-ziobro: Actually it works pretty well :)

dawid-ziobro: I was testing it before and after your comment @Fifciu . I've created a snippet to reproduce this behavior -> https://www.typescriptlang.org/play/?#code/MYewdgzgLgBFCGAnA5gU1gXhgbxvAXDAIwA0MARoQExnCEDMMAvgNwBQoksAFvGACYAbVIhhZsbGDDRRCAChDkAVvj4BPMgAdEITarBqAlGIB8OSVJicIIYQDpBIZHIAGAQWDBUECAEswyDDauiJQajAAJNjBmkwuhuyWMIjoAK6IYDCKSgDaMQC6MADUMADkAPrBAB5q5aXFMAAi8FCodmAgAO5yCRZMJGysbBzg0EE6NWIwYKidMAAKE2pyCCjoZLwCwoi9I5C2bY7O1Wp28LsA9BdWo7AQ2qjw-ADyyqjAmDgwdj8nzOxsK43fb2I5ye4pJ6vJTvKB2ci9CDoAAqvgAtqgQKkoHIeqZzElrLBcMBmFMTolLNYDg4nHJgL1+lQAAyswxsIA

packages/sdk/src/modules/loggerModule/wrapper.ts
dawid-ziobro: Changed to `injectMetadata` the rest is pretty much the same.

packages/sdk/src/modules/loggerModule/index.ts
Fifciu: On the surface, it indeed looks like it works but if you dive deeper you can spot it doesn't, e.g.: https://www.typescriptlang.org/play/?#code/MYewdgzgLgBFCGAnA5gU1gXhgbxvAXDAIwA0MARoQExnCEDMMAvgNwBQoksAFvGACYAbVIhhZsbGDDRRCAChDkAVvj4BPMgAdEITarBqAlGIB8OSVJicIIYQDpBIZHIAGAERABrcAFclqGH4QaFRNHxhtAC9USIidAA81GAASbG1dJhdDdksYRHQfRDAYRSUAbXTNAF0YAGoYAHIAfXTEpoa6mDd4KFQ7MBAAdzlsiyYSNlY2DnBoOJBEsRgwVEGYAAUEtTkEFHQyXgFhRFHJa1gIbVR4fgB5ZVRgTBwYOzfWpNYYAHpvmAAJEQBUqPHjwCAUVCoYrARwrfh4AQwACWUAaEIGsA+iLUAFsQPk7DAAMJwsDIsDIebAVAQCFQRDI5BofIIgBEMjZJWKqAAbiIkp5UEkoLw0RDBtwkoMAhAoTB6PE8rSfIJYCAAGZWWa2PqOZBnHX2fVyS75G73fxPOzwbI-P63GFk1AIkFPMgyvDAGl0mC8+CCHzArVs+BcyoiKBJPgI6y6hxOFEYkCwBlMlkunH4-JkcEwM3XO4PJ5J5YpvDzRbqbN9SS7GQ2pYAVgADC2WPaYAAZdDomD4-jIjVJf2B4MlYuwHwCESVtSGyDxk0Fi2Tm1234wACCEJXRatYIh5ChxXyNkE-IRFP7IFx0PVWth4HJlMRV-FZdg8DnByBME9MjEC0WxNN0vT9EMIxkJiMCti2TQAFQAIQLueepOHIHzrh2m5bki5qCHO+ZQMigiEcgyL8hCeZwcBCxqKBPR9AMwyGJIQA

packages/sdk/src/modules/loggerModule/index.ts
```diff
@@ -1,14 +1,28 @@
...
+import { injectMetadata } from "./injectMetadata";
+import { isInvalidConfig } from "./utils";
 
 export const loggerModule = (options?: LoggerModuleConfig) => {
   const logger =
     options?.handler ?? LoggerFactory.create(LoggerType.ConsolaGcp, options);
 
+  if (isInvalidConfig(options)) {
+    console.warn(
+      "Both handler and options are provided to logger's options. Using handler, options will be ignored."
```
bartoszherba, line 13: Please, use this file as a reference: we want to have more verbose message with a clear steps of what to do and where to look
https://github.com/vuestorefront/vue-storefront/blob/553dd5144c9d0fdeb6e2c32b1783b85a6efd10aa/packages/middleware/src/apiClientFactory/applyContextToApi.ts

packages/sdk/src/modules/loggerModule/index.ts
```diff
@@ -1,14 +1,28 @@
...
+  const loggerWithMetadata = injectMetadata(logger, {
+    alokai: {
+      context: "storefront",
+    },
+  });
+
   return {
     connector: {
-      ...logger,
+      ...loggerWithMetadata,
```
bartoszherba, line 25: It will not work, let's talk about it on the daily.

packages/sdk/src/modules/loggerModule/injectMetadata.ts
```diff
@@ -0,0 +1,27 @@
+import type { LoggerInterface } from "./types";
+
+const methodsToExclude = ["log"];
```
bartoszherba, line 3: This is not required anymore.

packages/sdk/src/modules/loggerModule/utils.ts
```diff
@@ -0,0 +1,13 @@
+import type { LoggerModuleConfig } from "./types";
+
+export const isEmptyObject = (obj = {}) => {
+  return Object.keys(obj).length === 0;
+};
+
+export const isInvalidConfig = (config?: LoggerModuleConfig) => {
```
bartoszherba, line 7: Please, add a docblock that will shortly explain what is "invalid" for us, because it is not invalid in common sense but it does not apply to custom logger.

bartoszherba (changes requested):

Fifciu (commented):
PR looks good expect Proxy spread let's talk about it on daily

Fifciu (approved):
During an analysis on daily meeting, we identified some issues with spreading proxy but as @dawid-ziobro said earlier, they are not affecting our use case. So eventually, we agreed to use spread.

packages/sdk/src/modules/loggerModule/index.ts
dawid-ziobro: As discussed, it stays as it is.

packages/sdk/src/modules/loggerModule/utils.ts
dawid-ziobro: Good idea, we could forget about what is invalid in some time. 
It's added now.

packages/sdk/src/modules/loggerModule/injectMetadata.ts
dawid-ziobro: Yup, removed.

packages/sdk/src/modules/loggerModule/index.ts
dawid-ziobro: Added, @bartoszherba please take a look.

PR #7282, opened 2024-10-18
docs(tfal-56): next/nuxt logger implementation

dawid-ziobro:
https://alokai.atlassian.net/browse/TFAL-56

### DIFF
--- a/docs/content/4.sdk/2.getting-started/4.logger.md
+++ b/docs/content/4.sdk/2.getting-started/4.logger.md
@@ -51,7 +51,8 @@ export default defineSdkConfig(({ buildModule, loggerModule, middlewareModule, g
 
 ## Configuration
 
-Available options:
+The default configuration is already provided, but you can customize it by providing the following options:
+
 * `level` - the minimum level of the message to be logged. The default value is `info`. Possible values are:
   * `emergency` 
   * `alert`
@@ -64,7 +65,7 @@ Available options:
 * `includeStackTrace` - a boolean value that determines if the stack trace should be included in the log message. The default value is `true`.
 * `handler` - a custom handler that will be called when the message is logged.
 
-To provide the configuration for the Logger, you need to update the SDK config with the Logger module configuration by providing an object with options to the `buildModule` function, f.e:
+To provide the configuration for the Logger, you need to update the SDK config with the Logger module configuration by providing an object with options to the `buildModule` function, e.g.:
 ```ts
   // ...
   logger: buildModule(loggerModule, {
@@ -75,7 +76,7 @@ To provide the configuration for the Logger, you need to update the SDK config w
 ```
 
 ### Custom handler
-There might a certain use case where you want to customize the way the Logger logs messages f.e. you want to change a log provider to a different one.
+There might a certain use case where you want to customize the way the Logger logs messages e.g. you want to change a log provider to a different one.
 Then you can provide your own custom handler for the Logger. The handler must implement the `LoggerInterface` interface provided in the section [types](#type).
 
 ::warning
--- a/docs/content/4.sdk/2.getting-started/4.logger.md
+++ b/docs/content/4.sdk/2.getting-started/4.logger.md
@@ -1,0 +1,171 @@
+# Logger
+
+The Logger provides a logging utility for Alokai Storefront projects. It allows you to log messages at various levels
+and ability to provide additional context. 
+
+It is created to provide a unified way of logging messages across the project along with providing additional data out of the box.
+
+## Installation
+
+If the Logger is already included in the Alokai Storefront, you can skip this step.
+
+The Logger is provided by the SDK and framework specific packages.
+
+In order to install the Logger, you need to update following packages to at least the following versions:
+* `@vue-storefront/sdk` to `3.3.0`
+* `@vue-storefront/nuxt` to `6.2.0`
+* `@vue-storefront/next` to `4.3.0`
+  
+After updating the packages, you need to provide the Logger module to the SDK config.
+
+::tabs{:titles='["Next.js", "Nuxt"]' class="mt-8"}
+
+#tab-1
+
+```ts
+// apps/storefront-unified-nextjs/sdk/config.ts
+
+// ...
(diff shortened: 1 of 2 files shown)


docs/content/4.sdk/2.getting-started/4.logger.md
```diff
@@ -0,0 +1,168 @@
+# Logger
+
+The Logger provides a logging utility for Alokai Storefront projects. It allows you to log messages at various levels and ability to provide additional context. 
+
+It is created to provide a unified way of logging messages across the project along with providing additional data out of the box.
+
+## Installation
+
+If the Logger is already included in the Alokai Storefront, you can skip this step.
```
bartoszherba, line 9: If it is not, then what? In which circumstances it might be missing?

docs/content/4.sdk/2.getting-started/4.logger.md
```diff
@@ -0,0 +1,168 @@
+# Logger
+
+The Logger provides a logging utility for Alokai Storefront projects. It allows you to log messages at various levels and ability to provide additional context. 
+
+It is created to provide a unified way of logging messages across the project along with providing additional data out of the box.
```
bartoszherba, line 5: What additional data?

docs/content/4.sdk/2.getting-started/4.logger.md
```diff
@@ -0,0 +1,168 @@
...
+## Installation
+
+If the Logger is already included in the Alokai Storefront, you can skip this step.
+
+The Logger is provided by the SDK and framework specific packages.
+
+In order to install the Logger, you need to update following packages to at least the following versions:
+* `@vue-storefront/sdk` to `3.3.0`
+* `@vue-storefront/nuxt` to `6.2.0`
+* `@vue-storefront/next` to `4.3.0`
```
bartoszherba, line 16: Is it common to have migration guide in the getting started guide?
Storefronts does not have any migrations?

docs/content/4.sdk/2.getting-started/4.logger.md
```diff
@@ -0,0 +1,168 @@
...
+  * `critical`
+  * `error`
+  * `warning`
+  * `notice`
+  * `info`
+  * `debug`
+* `includeStackTrace` - a boolean value that determines if the stack trace should be included in the log message. The default value is `true`.
+* `handler` - a custom handler that will be called when the message is logged.
+
+To provide the configuration for the Logger, you need to update the SDK config with the Logger module configuration by providing an object with options to the `buildModule` function, f.e:
```
bartoszherba, line 67:
```suggestion
To provide the configuration for the Logger, you need to update the SDK config with the Logger module configuration by providing an object with options to the `buildModule` function, e.g.:
```

bartoszherba: It is worth to mention that the configuration is optional and provide default configuration.

docs/content/4.sdk/2.getting-started/4.logger.md
```diff
@@ -0,0 +1,168 @@
...
+  // ...
+  logger: buildModule(loggerModule, {
+    level: 'debug',
+    includeStackTrace: false,
+    handler: myCustomHandler,
+  }),
+```
+
+### Custom handler
+There might a certain use case where you want to customize the way the Logger logs messages f.e. you want to change a log provider to a different one.
```
bartoszherba, line 78:
```suggestion
There might a certain use case where you want to customize the way the Logger logs messages e.g. you want to change a log provider to a different one.
```

bartoszherba (changes requested):

docs/content/4.sdk/2.getting-started/4.logger.md
dawid-ziobro: It might be the case that a client already has a delivered storefront that doesn't yet have logger implemented. For such case it has to be installed, otherwise this step is not required.

docs/content/4.sdk/2.getting-started/4.logger.md
dawid-ziobro: We don't provide migration guides for storefronts, even tho those would be available internally.

dawid-ziobro: For this case, i wanted to specify a minimal version of packages when the logger becomes available.

docs/content/4.sdk/2.getting-started/4.logger.md
```diff
@@ -0,0 +1,169 @@
...
+* `includeStackTrace` - a boolean value that determines if the stack trace should be included in the log message. The default value is `true`.
+* `handler` - a custom handler that will be called when the message is logged.
+
+To provide the configuration for the Logger, you need to update the SDK config with the Logger module configuration by providing an object with options to the `buildModule` function, e.g.:
+```ts
+  // ...
+  logger: buildModule(loggerModule, {
+    level: 'debug',
+    includeStackTrace: false,
+    handler: myCustomHandler,
```
Fifciu, line 74: Providing `handler` makes other configuration properties redundant. I think we should either present other configuration properties or only handler property here.

docs/content/4.sdk/2.getting-started/4.logger.md
```diff
@@ -0,0 +1,169 @@
...
+// component
+<script setup>
+useSdk().logger.info('Example log');
+<script>
+```
+::
+
+Now instead of using `console.log` you can use the Logger to log messages at different levels.
+
+## Type
```
Fifciu, line 136: I think parts of this section should be placed in more meaningful places. E.g. LogLevel where you tell about log levels, logger interface where you tell about custom handler or just available methods of logger and so on.

docs/content/4.sdk/2.getting-started/4.logger.md
dawid-ziobro: My idea was to put all available options here but yeah, this might be a better idea. Especially that all options are listed above and handler has its own section below.

Fifciu (approved):

PR #7285, opened 2024-10-21
docs: working with logger

Fifciu:
[TFAL-55](https://alokai.atlassian.net/browse/TFAL-55)

[TFAL-55]: https://alokai.atlassian.net/browse/TFAL-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/content/3.middleware/2.guides/10.working-with-logger.md
+++ b/docs/content/3.middleware/2.guides/10.working-with-logger.md
@@ -1,0 +1,336 @@
+# Working with logger
+
+The middleware application provides a logger instance that automatically attaches metadata related to the scope of each call. By adding this contextual data, the logger makes it significantly easier to trace the origin of logs or errors, simplifying the debugging and monitoring process across the application.
+
+## Using logger
+
+The middleware application provides access to the logger in various parts of the system, such as in extensions and integrations. This flexibility allows you to log important events, errors, and other data throughout the lifecycle of your application. Whether you’re working in an extension or an integration, you can easily utilize the logger to capture and track necessary information.
+
+```json[Example log]
+{
+  "message": "log message we print",
+  "timestamp": "2024-10-22T19:04:18.791Z",
+  "severity": "INFO",
+  "context": "middleware",
+  "scope": {
+    "integrationName": "adyen",
+    "functionName": "addedCustomEndpoint",
+    "hookName": "afterCall",
+    "type": "requestHook",
+    "extensionName": "testing-adyen-extension"
+  }
+}
+```
+
+In this section, we will explore how to access and use the logger in different parts of the application, ensuring you have the tools to log data efficiently and consistently. 
+
+These examples will demonstrate the proper ways to obtain and interact with the logger, ensuring that you are able to implement consistent and effective logging practices across your middleware application. As you will see, `getLogger` function solves the problem in most cases.
+
+### Example log
+
+### extendApp
+
+```ts
+import { getLogger } from "@vue-storefront/middleware";
+
+const someExtension = {
+  name: "some-extension",
+  extendApp(params) {
+    const logger = getLogger(params);
+    logger.info("Logged succesfully!");
+  }
+};
+```
+
+### hooks
+
+Logger is available in hooks factory:
+
+```ts
+import { getLogger } from "@vue-storefront/middleware";
+
+const someExtension = {
+  name: "some-extension",
+  hooks(req, res, alokai) {
+    const logger = getLogger(alokai);
+    logger.info("hooks");
+    return {
+      // ...
+    };
+  },
+};
+```
+
+In order, to use logger in certain hooks like `beforeCall`, `beforeCreate`, `afterCall`, `afterCreate` obtain it from it's params:
+
+```ts
+import { getLogger } from "@vue-storefront/middleware";
+
+const someExtension = {
+  name: "some-extension",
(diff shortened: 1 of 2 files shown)


docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,311 @@
+# Working with logger
+
+The middleware application provides a logger instance that automatically attaches metadata related to the scope of each call. This scope includes crucial information such as the name of the integration, the extension (if applicable), the function name (if applicable), the hook name (if applicable), and the type of the action being logged (e.g., endpoint, bootstrapHook, requestHook). By adding this contextual data, the logger makes it significantly easier to trace the origin of logs or errors, simplifying the debugging and monitoring process across the application.
```
bartoszherba, line 3: I wonder if we have be that specific here? Specific information are usually outdated fast and must be adjusted. Maybe, instead of listing what we add, we just describe in general what we add? (no details)

docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Using logger
+
+The middleware application provides access to the logger in various parts of the system, such as in extensions and integrations. This flexibility allows you to log important events, errors, and other data throughout the lifecycle of your application. Whether you’re working in an extension or an integration, you can easily utilize the logger to capture and track necessary information.
+
+In this section, we will explore how to access and use the logger in different parts of the application, ensuring you have the tools to log data efficiently and consistently. 
+
+These examples will demonstrate the proper ways to obtain and interact with the logger, ensuring that you are able to implement consistent and effective logging practices across your middleware application. As you will see, `getLogger` function solves the problem in most cases.
```
bartoszherba, line 11: Maybe we can add output examples, it would be nice to show how our logger improve contextual data. If we do not want to add such example in all cases (that might be too specific as in my first comment), maybe we can add at least one such example.

docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,311 @@
...
+In order, to use logger in certain hooks like `beforeCall`, `beforeCreate`, `afterCall`, `afterCreate` obtain it from it's params like that:
+
+```ts
+import { getLogger } from "@vue-storefront/middleware";
+
+const someExtension = {
+  name: "some-extension",
+  hooks(req, res, alokai) {
+    return {
+      beforeCall(params) {
```
bartoszherba, line 55: that `params` is kind of misleading? Or it is just me that I interpret it as request parameters?

docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,311 @@
...
+        return params;
+      },
+    };
+  },
+};
+```
+
+#### Caveat: Accessing logger via closure
+
+Consider the following snippet:
```
bartoszherba, line 82: Some explanation is missing, I am not sure what is actually forbidden

docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,311 @@
...
+export interface LoggerOptions {
+  /**
+   * The log level aligned with RFC5424.
+   */
+  level?: LogLevel;
+
+  /**
+   * Whether to include the stack trace in the log message.
+   */
+  includeStackTrace?: boolean;
```
bartoszherba, line 188: There is no `handler?: LoggerInterface` option?

docs/content/3.middleware/2.guides/10.working-with-logger.md
```diff
@@ -0,0 +1,311 @@
...
+
+```ts
+/**
+ * Options for the logger.
+ */
+export interface LoggerOptions {
+  /**
+   * The log level aligned with RFC5424.
+   */
+  level?: LogLevel;
```
bartoszherba, line 183: Can we add default values?

bartoszherba (changes requested):

docs/content/3.middleware/2.guides/10.working-with-logger.md
Fifciu: Honestly, I didn't feel like it's related to router in any way and I am not sure if it's the case. I believe it's a cognitive bias and not most devs think about router's params. It's about function's parameters. We also called it like that in types:

```ts
  beforeCreate?: (params: HookParams<C>) => C;
  afterCreate?: (params: HookParams<C>) => C;
  beforeCall?: <ARGS>(params: BeforeCallParams<C, ARGS>) => BeforeCallArgs;
  afterCall?: <ARGS, RESPONSE>(
    params: AfterCallParams<C, ARGS, RESPONSE>
  ) => AfterCallArgs;
}
```

docs/content/3.middleware/2.guides/10.working-with-logger.md
Fifciu: Below snippet you have explaination but I also adjusted comments

bartoszherba (approved):

PR #7292, opened 2024-10-28
chore: export types from logger module

bartoszherba:
[TFAL-71](https://alokai.atlassian.net/browse/TFAL-71)

[TFAL-71]: https://alokai.atlassian.net/browse/TFAL-71?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sdk/src/modules/loggerModule/index.ts
```diff
@@ -1,32 +1,3 @@
...
-  });
-
-  return {
-    connector: {
-      ...loggerWithMetadata,
-    },
-  } satisfies Module;
-};
+export { loggerModule } from "./module";
+export * from "./utils";
```
dawid-ziobro, line 2: There's no need to export `utils`, this is internal only.

dawid-ziobro (approved):

.changeset/eighty-turkeys-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vue-storefront/sdk": major
```
Fifciu, line 2: Why major?

packages/sdk/src/modules/loggerModule/index.ts
bartoszherba: Fixed

.changeset/eighty-turkeys-marry.md
bartoszherba: Changed to patch, I thought it does not matter if we are on the prerelease mode.

PR #7302, opened 2024-11-04
fix(ES-1214): fix XSS vulnerability

WojtekTheWebDev:
A potential XSS (Cross-Site Scripting) vulnerability in the middleware has been addressed.

Each param is now validated before being used in the middleware.

### DIFF
--- a/packages/middleware/src/handlers/validateParams/index.ts
+++ b/packages/middleware/src/handlers/validateParams/index.ts
@@ -4,9 +4,10 @@ import { IntegrationsLoaded } from "../../types";
 
 export function validateParams(integrations: IntegrationsLoaded) {
   return (req: Request, res: Response, next: NextFunction) => {
-    const { method } = req;
-    const { integrationName } = req.params;
+    const { method, params } = req;
+    const { integrationName } = params;
 
+    // Validate the request method
     if (method !== "GET" && method !== "POST") {
       res.status(405);
       res.send(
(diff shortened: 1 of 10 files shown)


packages/middleware/src/handlers/validateParams/index.ts
```diff
@@ -0,0 +1,37 @@
...
+      res.status(405);
+      res.send(
+        `Method ${method} is not allowed. Please, use GET or POST method.`
+      );
+
+      return;
+    }
+
+    // Validate the integration
+    if (!integrations || !integrations[integrationName]) {
```
bartoszherba, line 21: If I remember correctly, previously - if the method name was a malicious string, our middleware was trying to call it inside `getApiFunction`. It recognized that the function is not registered and throws. Error handler then send back that exact message which was a malicious html snippet.

You have prevented that in lines 30-33, that is ok. 

But here I think is similar situation. If someone will make a call with malicious code as an `integrationName` it will be sent back as it is. This is a very similar situation, You do not sanitize the integration name which also comes from params. Also, there are not tests for such case. Integration name is an user input that must be sanitized in the same way as other params before you eventually send it back to the browser.

WojtekTheWebDev: ty for finding this, fixed in recent commit

Fifciu (approved):

bartoszherba (approved):

PR #7312, opened 2024-11-12
Expose createExpress

Tofandel:
In v4 of @vue-storefront/middleware we used to get an express app from createServer

Since I self host in a complex setup I was using the express server directly instead of an http server, It was very usefull because I could also do requests to it internal to the process, which you can't with a node server

So this PR splits the createServer into createExpress and createServer so that we can still use createExpress to get the express app directly

### DIFF
--- a/packages/middleware/src/createServer.ts
+++ b/packages/middleware/src/createServer.ts
@@ -1,6 +1,6 @@
 import cookieParser from "cookie-parser";
 import cors from "cors";
-import express from "express";
+import express, { Express } from "express";
 import type { HelmetOptions } from "helmet";
 import helmet from "helmet";
 import http, { Server } from "node:http";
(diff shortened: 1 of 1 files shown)


WojtekTheWebDev:
Hi @Tofandel thanks for the PR! The change you proposed is a good direction, however only `createServer` applies `createTerminus` function and exporting `createExpress` separately is allowing to skip it. I'm debugging right now if we can return both `server` and `app` from `createServer` function to have a cake and eat it. Will keep you updated!

Tofandel:
Thanks in my case it's not a problem because I don't use terminus and since I will expose the express server and not the node one, I will be skipping it anyways but yes exposing both would still work

WojtekTheWebDev:
Hey @Tofandel, once again, we had some internal discussions and decided that extending Express in this way is not a recommended solution on our end. Express is an internal dependency, and such behavior could make future migrations more challenging if we decide to switch from Express to something else. However, we still allow access to the Express instance through our [extensions](https://docs.alokai.com/middleware/guides/extensions) using the `extendApp` feature.

PR #7323, opened 2024-11-21
feat: add upload functionality in the middleware

bartoszherba:
Summary

The issue involves enhancing the middleware express app by allowing the addition of multiple body parsers. Currently, only the JSON body parser is available, which limits functionality for developers needing to implement file uploads.

Context

The middleware express app is currently configured to use only a JSON body parser. This restriction poses challenges for developers who require different body parsers for specific functionalities, such as file uploads.


.changeset/hot-maps-unite.md
```diff
@@ -0,0 +1,29 @@
...
+  // Do something with files
+
+  return Promise.resolve({
+    status: 200,
+    message: "ok",
+  });
+};
+```
+
+Available options are:
```
jagoral, line 23: You could add some context here - Available options for what?

packages/middleware/__tests__/integration/bootstrap/serverWithUpload.ts
```diff
@@ -0,0 +1,23 @@
+import { apiClientFactory } from "../../../src/apiClientFactory";
+import * as api from "./api";
+import { AlokaiContainer, getLogger } from "../../../src";
+
+const onCreate = async (
+  config: Record<string, unknown> = {},
+  alokai: AlokaiContainer
+) => {
+  const logger = getLogger(alokai);
+  logger.info("oncreate");
```
jagoral, line 10: I know that it's just a test integration, but what is the purpose of this log?

packages/middleware/__tests__/integration/upload.spec.ts
jagoral, line 1: maybe we could add a test case `should allow sending non multipart/form-data requests when fileUpload is enabled`. If you feel that it's an overkill, feel free to skip it

packages/middleware/package.json
```diff
@@ -27,7 +27,9 @@
     "helmet": "^5.1.1",
     "@godaddy/terminus": "^4.12.1",
     "lodash.merge": "^4.6.2",
-    "xss": "^1.0.15"
+    "xss": "^1.0.15",
+    "multer": "^1.4.5-lts.1",
+    "@types/multer": "^1.4.7"
```
jagoral, line 32: shouldn't `@types/multer` be in devDependencies?

packages/middleware/src/types/fileUpload.ts
```diff
@@ -0,0 +1,7 @@
+export interface FileUploadOptions {
+  enabled?: boolean;
+  maxFileSize?: number; // in bytes
+  maxFiles?: number;
+  allowedMimeTypes?: string[];
+  fieldNames?: string[]; // specific field names to accept
```
jagoral, line 6: could you add default values?

packages/middleware/src/types/fileUpload.ts
```diff
@@ -0,0 +1,7 @@
+export interface FileUploadOptions {
+  enabled?: boolean;
```
jagoral, line 2: hmm just thinking out loud - wouldn't be easier to remove `enabled` (whenever the file upload config is provided, file upload is enabled) flag or set it to true `enabled: true`?

packages/middleware/src/createServer.ts
```diff
@@ -50,6 +51,8 @@ async function createServer<
 
   const app = express();
 
+  configureFileUpload(app, options.fileUpload);
```
jagoral, line 54: (nitpick, feel free to skip this comment) imo we could improve readiness, by changing this line to sth like this:

```
  app.use(createMulterMiddleware(options.fileUpload))
```

as it's clear in most of the file which middlewares we register: `cookieParser`, `jsonBodyParser` etc. `configureFileUpload(app)` hides this informtion, and we have to navigate to separate file to get to know that multer is used under the hood.

packages/middleware/src/types/fileUpload.ts
jagoral: @bartoszherba I meant that we could set the type to accept only `true` value

```ts
export interface FileUploadOptions {
  enabled: true;
```

so there is no way to set `enable: false` and provide other configuration options

packages/middleware/src/types/common.ts
```diff
@@ -22,6 +22,14 @@ export type ResponseWithAlokaiLocals = Response<
     [key: string]: any;
   }
 >;
+export type RequestWithFiles = Request & {
```
lsliwaradioluz, line 25: I am afraid if we decide to enhance the Request object even further in the future, we're gonna have a problem with that type's name.

Currently it's just Request + Files but perhaps we will add another request-enhancing middleware in the future (e.g. [session](https://github.com/expressjs/session)). Then we will have to rename it to `RequestWithFilesAndSession`) and the more middlewares we add, the longer the name it will become.

Can we use this intersection type in line 107?

lsliwaradioluz: Or maybe we can just come up with some more generic name, e.g. `AlokaiRequest`

lsliwaradioluz (approved):

PR #7330, opened 2024-11-26
fix: upload multiple files

bartoszherba:


jagoral (approved):
Could you just add a test which will protect us before the regression

PR #7335, opened 2024-11-27
chore: loosen up eslint rules

FRSgit:
https://alokai.atlassian.net/browse/UST-1825

### DIFF
--- a/engineering-toolkit/eslint-config/src/ecma.js
+++ b/engineering-toolkit/eslint-config/src/ecma.js
@@ -42,7 +42,7 @@ export function ecma(config, ...overrides) {
             },
             name: "import",
             rules: {
-              "import/no-anonymous-default-export": "warn",
+              "import/no-anonymous-default-export": "off",
             },
             settings: {
               "import/extensions": [".ts", ".tsx", ".mts", ".cts", ".mtsx", ".ctsx", ".js", ".jsx", ".mjs", ".cjs"],
--- a/engineering-toolkit/eslint-config/src/typescript.js
+++ b/engineering-toolkit/eslint-config/src/typescript.js
@@ -61,10 +61,10 @@ export function typescript(config, ...overrides) {
         "@typescript-eslint/explicit-module-boundary-types": "warn",
         "@typescript-eslint/naming-convention": "off",
         "@typescript-eslint/no-empty-interface": "off",
-        "@typescript-eslint/no-empty-object-type": "off",
+        "@typescript-eslint/no-empty-object-type": "warn",
         "@typescript-eslint/no-explicit-any": "warn",
         "@typescript-eslint/no-inferrable-types": "warn",
-        "@typescript-eslint/no-magic-numbers": "off",
+        "@typescript-eslint/no-magic-numbers": "warn",
         "@typescript-eslint/no-namespace": "warn",
         "@typescript-eslint/no-non-null-assertion": "off",
         // allow unused vars starting with `_` https://typescript-eslint.io/rules/no-unused-vars
@@ -103,6 +103,6 @@ export function typescript(config, ...overrides) {
           }),
         ]
       : [],
-    overrides
+    overrides,
   );
 }
--- a/engineering-toolkit/eslint-config/src/typescript.js
+++ b/engineering-toolkit/eslint-config/src/typescript.js
@@ -26,7 +26,7 @@ export function typescript(config, ...overrides) {
             },
             name: "import",
             rules: {
-              "import/no-anonymous-default-export": "warn",
+              "import/no-anonymous-default-export": "off",
             },
             settings: {
               "import/resolver": {
@@ -63,7 +63,7 @@ export function typescript(config, ...overrides) {
         "@typescript-eslint/no-empty-interface": "off",
         "@typescript-eslint/no-explicit-any": "warn",
         "@typescript-eslint/no-inferrable-types": "warn",
-        "@typescript-eslint/no-magic-numbers": "warn",
+        "@typescript-eslint/no-magic-numbers": "off",
         "@typescript-eslint/no-namespace": "warn",
         "@typescript-eslint/no-non-null-assertion": "off",
         // allow unused vars starting with `_` https://typescript-eslint.io/rules/no-unused-vars
@@ -76,7 +76,9 @@ export function typescript(config, ...overrides) {
             varsIgnorePattern: "^_",
           },
         ],
-        "@typescript-eslint/no-use-before-define": "warn",
+        "@typescript-eslint/no-use-before-define": ["warn", {
+          functions: false,
+        }],
         "@typescript-eslint/prefer-function-type": "warn",
         "@typescript-eslint/no-empty-object-type": "warn",
         "default-param-last": "off",
--- a/engineering-toolkit/eslint-config/src/typescript.js
+++ b/engineering-toolkit/eslint-config/src/typescript.js
@@ -61,6 +61,7 @@ export function typescript(config, ...overrides) {
         "@typescript-eslint/explicit-module-boundary-types": "warn",
         "@typescript-eslint/naming-convention": "off",
         "@typescript-eslint/no-empty-interface": "off",
+        "@typescript-eslint/no-empty-object-type": "off",
         "@typescript-eslint/no-explicit-any": "warn",
         "@typescript-eslint/no-inferrable-types": "warn",
         "@typescript-eslint/no-magic-numbers": "off",
@@ -76,11 +77,13 @@ export function typescript(config, ...overrides) {
             varsIgnorePattern: "^_",
           },
         ],
-        "@typescript-eslint/no-use-before-define": ["warn", {
-          functions: false,
-        }],
+        "@typescript-eslint/no-use-before-define": [
+          "warn",
+          {
+            functions: false,
+          },
+        ],
         "@typescript-eslint/prefer-function-type": "warn",
-        "@typescript-eslint/no-empty-object-type": "warn",
         "default-param-last": "off",
         "no-magic-numbers": "off",
         "no-use-before-define": "off",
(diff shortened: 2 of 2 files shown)


Razz21 (commented):
I was wondering if wouldn't be feasible to provide each config there with a set of **strict** rules (enabled conditionally) which **would** enable some of these "blocking" rules. 
These rules would be intended for a development, while for demos/ clients projects we would deliver configs with strict rules disabled (or even better - without strict enabled)

engineering-toolkit/eslint-config/src/typescript.js
```diff
@@ -63,7 +63,7 @@ export function typescript(config, ...overrides) {
         "@typescript-eslint/no-empty-interface": "off",
         "@typescript-eslint/no-explicit-any": "warn",
         "@typescript-eslint/no-inferrable-types": "warn",
-        "@typescript-eslint/no-magic-numbers": "warn",
+        "@typescript-eslint/no-magic-numbers": "off",
```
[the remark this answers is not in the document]
Razz21: please test this on all gitignored or autogenerated files, because some rules throws only in unified-demo repo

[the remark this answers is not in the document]
FRSgit: Thanks for pointing that out: I've disabled `@typescript-eslint/no-empty-object-type` 9e27d830e99bb0a42c96580335b96f2eb9c57f19

With this configuration I've been able to come down to 15 warnings in total in next/nuxt/middleware/playwright apps Out of those 15 warnings 10 is `@typescript-eslint/no-explicit-any` and 4 are `vue/no-v-html`. So I'd say these changes are enough 😄 

On the side not - I think `@shared/unified-api-eslint-config` from `unified-storefronts` shouldn't be used anymore. Now it's a dependency only in unified-api-sfcc, for every other package we use biome.

bartoszherba (commented):
Why do we change these rules?

engineering-toolkit/eslint-config/src/typescript.js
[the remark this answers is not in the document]
bartoszherba: I do not like that we remove this rule, it helps to keep code clean and readable. What are the arguments in favor of removing this rule besides that our code does not align with it? (in what cases?). In fact, this is one of the first rules about clean code you can possibly learn, no magic number, no strings without variables and so on.

engineering-toolkit/eslint-config/src/ecma.js
```diff
@@ -42,7 +42,7 @@ export function ecma(config, ...overrides) {
             },
             name: "import",
             rules: {
-              "import/no-anonymous-default-export": "warn",
+              "import/no-anonymous-default-export": "off",
```
bartoszherba, line 45: Why do we disable it? Default exports are bad pattern in the first place as it allows to have the same thing under unlimited amount of aliases in the code. Having this even if the module author did not export things as default is even worse as it not only inherit the first complain but also allows to write a code that is not aligned with its intentions.

engineering-toolkit/eslint-config/src/typescript.js
```diff
@@ -61,9 +61,10 @@ export function typescript(config, ...overrides) {
         "@typescript-eslint/explicit-module-boundary-types": "warn",
         "@typescript-eslint/naming-convention": "off",
         "@typescript-eslint/no-empty-interface": "off",
+        "@typescript-eslint/no-empty-object-type": "off",
```
bartoszherba, line 64: It should be at least a warning because in 99% of cases interface like `interface Empty {}` is a mistake, such interface is redundant.

bartoszherba (changes requested):
Changes like that affects the whole company. It will have a great impact on the code quality and it should be not merged in such task. I understand the problem of reducing warnings when we lint the code but the goal should be do align with the best possible set of rules that will help us maintain the code, not to remove rules to address the threshold requirement. Such changes deserves RFC with an explanation of why do we things like we do. Currently warns are mostly harmless and it is a shame that we ignore them in the first place. If we turn off relative rules it will be nothing more like turning your back to the problem.

engineering-toolkit/eslint-config/src/ecma.js
bartoszherba: But you must import it with the original name and then consciously add an alias. In case of defaults you can make a mistake and you might never notice it.

engineering-toolkit/eslint-config/src/typescript.js
lukasborawski: What's the use case here, it's for extending?

lukasborawski:
Can we get use cases and reasoning? The statement that some rules are problematic is not so mature here. We need a wider discussion about these rules, some of them are a real pain in the ass, but at the same time, they're quite important, and someone decided to use them not on top of a hunch, right?

engineering-toolkit/eslint-config/src/ecma.js
lukasborawski: What's wrong with that one?

engineering-toolkit/eslint-config/src/typescript.js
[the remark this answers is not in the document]
FRSgit: No magic number eslint rule is unfortunately not implemented well. 99% of the time it shows a false-positives (Especially with react hooks/vue reactivity). Currently in unified-storefronts is shows more than a hundred of warnings (100% of those are false positives). Here some examples from unified-storefront codebase:

Middleware:
image
image
image

Next.js:
image
image
image
image
image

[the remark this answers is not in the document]
FRSgit: All of those I believe to be self-explanatory, but eslint is printing a waning about every single one.

Though, if you believe this rule is important for your team, we can always turn this rule off directly in the storefronts/modules repositories. What do you think @bartoszherba ?

engineering-toolkit/eslint-config/src/ecma.js
FRSgit: I'm not sure if we have understanding of what this rule actually checks.
This is okay for the rule: `export default function Something() {`
and this is bad: `export default function () {`

But in both cases it's still a default export and you can import it like `import WhatheverNameILike from '...'`.

So IMO this rule just does not make sense - what's the difference if you name something before making a default export or not? Especially that it tends to display false-positives for configuration files.

Can we agree that it makes sense to disable this one or should I revert is as well?

engineering-toolkit/eslint-config/src/typescript.js
FRSgit: We can move this one to storefronts codebase. We have more than 20 cases of things like this:
image
image

It's code supposed to be extended by the end user (or sometimes just to create "our" own exported type - in case we need to change it in the future we don't just reexport typings from subpackages.

But as I said - if it's problematic I can move this to storefronts

FRSgit:
@lukasborawski @lsliwaradioluz @bartoszherba Sorry, I thought those rules will be rather insignificant for you.
If you could - please write here which ones can be merged into this repository in your opinion.

I'll move everything else into storefronts repo as I don't want to waste too much of your time on eslint-rules-related discussions 😄 

Sorry again, should've pinged you for your opinion

engineering-toolkit/eslint-config/src/ecma.js
bartoszherba: Then the question is, why do we want to allow anonymous functions?

engineering-toolkit/eslint-config/src/typescript.js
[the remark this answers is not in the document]
bartoszherba: In fact all these examples can be reduced to (minValue is just an example):
```
const MIN_QUANTITY = 1;
minValue = MIN_QUANTITY;
```

- Enhances code readability and maintainability.
- Provides a single point of change if the value ever needs to be updated.
- Helps document the intent of the 1.

But all of them seems like no-magic-number false because of the context. I agree it is a "philosophical" area in such cases. I also think that if we remove the rule we would need to be super aware of it on CR.

bartoszherba (approved):

engineering-toolkit/eslint-config/src/ecma.js
FRSgit: I think it's just unnecessary requirement - it does not bring anything good imo, just a stylistic change.
On the other hand, we get few warnings about e.g. our [next.config.js](https://github.com/vuestorefront/unified-storefronts/blob/main/apps/storefront-unified-nextjs/next.config.mjs#L29) and boilerplate for i18n: [barrel export function](https://github.com/vuestorefront/unified-storefronts/blob/main/apps/storefront-unified-nextjs/lang/en/index.ts#L4-L6)

IMO these cases are not something I'd like to see in my eslint output 🤷 

FYI: I think it's a minor and I see you gave me a thumbs up, so if I don't get any response to this comment till tomorrow I'll just merge this PR with this rule turned off.

engineering-toolkit/eslint-config/src/typescript.js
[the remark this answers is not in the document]
FRSgit: that's fair. I personally prefer to look out for those during the review, because it's not that important matter that eslint needs to notify me of it every time and as you said - it's kind of philosophical and imo arguable.

So, I think we can agree that this rule should be moved directly to the storefronts repo as it's a team decision to ignore it - it shouldn't affect whole comapny

engineering-toolkit/eslint-config/src/typescript.js
lukasborawski: Yeah, extending, classic issue.

lukasborawski (approved):

engineering-toolkit/eslint-config/src/ecma.js
FRSgit: okay, leaving this one in the PR

PR #7353, opened 2024-12-12
feat(ES-1378): export default error handler

jagoral:
[ES-1378](https://alokai.atlassian.net/browse/ES-1378)

Export the `defaultErrorHandler`. It might be useful when someone wants to do something before delegating the job to our built-in error handler.

[ES-1378]: https://alokai.atlassian.net/browse/ES-1378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/middleware/src/errors/defaultErrorHandler.ts
+++ b/packages/middleware/src/errors/defaultErrorHandler.ts
@@ -1,32 +1,21 @@
-import type { Request } from "express";
 import { getAgnosticStatusCode } from "../helpers";
-import type { AlokaiResponse } from "../types";
-
-type ClientSideError = {
-  message?: string;
-};
+import type { Integration } from "../types";
 
 /**
  * Default error handler for the middleware
- *
- * @param error
- * @param req
- * @param res
  */
-export const defaultErrorHandler = (
-  error: ClientSideError,
-  req: Request,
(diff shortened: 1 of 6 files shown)


bartoszherba (approved):

packages/middleware/src/errors/defaultErrorHandler.ts
```diff
@@ -1,32 +1,21 @@
...
- * @param res
  */
-export const defaultErrorHandler = (
-  error: ClientSideError,
-  req: Request,
-  res: AlokaiResponse
+export const defaultErrorHandler: Integration["errorHandler"] = (
+  error,
+  _req,
+  res
```
rohrig, line 10: If I try removing the types now in the customer's project, there are still errors because they are implicit type any. 

image

jagoral: even when you add types for the integration? (`satisfies`)
CleanShot 2024-12-12 at 18 05 26@2x

rohrig: that does it, missed that part. Thanks

PR #7374, opened 2025-01-28
poc: processHeaders handler

WojtekTheWebDev:
a PoC to address issue with set-cookie headers in CDN implementation - added handler modifies all request that comes through the middleware app and replace `x-alokai-locale` header with `vsf-locale` cookie.


WojtekTheWebDev:
Closing the PR as it's only a PoC

PR #7401, opened 2025-02-20
docs(TF-410): multibrand ci/cd

jagoral:


docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,407 @@ navigation:
 
 # Deployment - CI/CD
 
+Managing deployments for multiple stores can be challenging. This guide explains how Continuous Integration (CI) and Continuous Deployment (CD) work in a multistore setup, helping you create an efficient and reliable deployment process.
```
grixu, line 10: A bit of storytelling here would be nice - like: deployment working on local machine is fine, but let's be honest, the one we want it's automatically happening when you push your commits to your git provider.

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,407 @@ navigation:
...
+## Continuous Integration (CI)
+
+Let's look at how continuous integration works in a multistore setup, using GitHub Actions as an example while explaining how to adapt each step for other CI platforms.
+
+### 1. Setting Up Dependencies
+
+First, we need to set up the environment. Our setup action handles several crucial steps:
+
+::tip Why a Separate Setup Action?
+We use a separate setup action to follow the DRY (Don't Repeat Yourself) principle, as these setup steps are used in multiple workflows - for instance `continuous-integration.yml` and `continuous-deployment.yml`. However, you can include these steps directly in your workflow if you prefer - there's no technical requirement to have them as a separate action.
```
grixu, line 44: Great! 👍

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,407 @@ navigation:
...
+::
+
+Here's how our setup action is defined:
+
+```yaml
```
grixu, line 49: How could we have this in sync with the original file? To avoid having an outdated docs 🤔

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,407 @@ navigation:
...
+      npm_email: ${{ vars.NPM_EMAIL }}
+```
+
+::tip Adapting for Other CI Platforms
+Your CI platform needs to support: full git history access for change detection, Node.js environment setup, Yarn cache configuration, Turbo build cache setup for faster builds, and access to Alokai Enterprise Registry.
+::
+
+### 2. Change Detection
+
+The change detection step identifies which stores need deployment. Here's a simplified version of our approach:
```
grixu, line 130: "which stores need deployment" - but later we also talking about tests. So maybe just name it as "which stores are changed"?

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,407 @@ navigation:
...
+    inputs:
+      store_ids:
+        description: 'Space separated list of store IDs to deploy'
+        required: false  # When not provided, detect changed stores
+```
+
+You can use the automatic deployment for staging and manual deployment for production environment.
+
+::tip Deployment Protection
+For Github Actions, you can use deployment protection rules to require manual approval before deployments. Check out the Github [docs](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments) for more details.
```
grixu, line 245: Very nice that you put it here 👍

grixu (approved):

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
jagoral: sure 👍 https://github.com/vuestorefront/vue-storefront/pull/7401/commits/66846511245942cdd7f51773f16be41437ff7efd

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
jagoral: that's a valid question.. As this is a general problem in our docs, and we don't have a solution for it, I've decided to describe the setup action on a high level https://github.com/vuestorefront/vue-storefront/pull/7401/commits/6f44a364fafe8e06df9ccc9cacada78975af1637

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,382 @@ navigation:
...
+
+In a multistore setup, CI/CD is designed to:
+1. Automatically detect which stores have changed
+2. Run tests only for affected stores
+3. Build and deploy only the changed stores
+
+This targeted approach ensures:
+- Faster deployments (only changed stores are processed)
+- Efficient resource usage
+- Reduced redundancy (unchanged stores are skipped in CI/CD pipeline)
```
mateuszo, line 33: _Reduced redundancy_ doesn't sound right.

```suggestion
- Less redundancy (unchanged stores are skipped in CI/CD pipeline)
```

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,382 @@ navigation:
...
+
+## Continuous Integration (CI)
```
mateuszo, line 35: Having all the steps listed here and explained briefly would be nice. An image visualizing this would be superb ;)

mateuszo (commented):
This is great. However, I'm not sure whether this should be in the public documentation. On one hand, I'm not keen on hiding any information, but on the other knowledge presented here might give an incentive to our customers to build their own pipelines and deploy to their own infra, what is AFAIK not what we want. Or, even they might even be deploying to our cloud but with a custom pipeline and I don't think we want to support that - IMHO it creates a new bug category. Lastly, I find this as our secret sauce ;) we give it for free here and make our service look less like a service. 

WDYT @filrak  ?

jagoral:
@mateuszo I think this docs might be helpful for customers who don't use Github Actions. If we don't provide CI/CD for other providers (like Gitlab, Bitbucket, Jenkins), they can use this page to build it on their own

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
```diff
@@ -7,14 +7,382 @@ navigation:
...
+### 3. Running Tests
+
+```yaml
+steps:
+  - name: Run tests for affected stores
+    run: |
+      yarn store test --store-id $STORE_IDS
+```
+
+## Continuous Deployment (CD)
```
filrak, line 186: We decided with @mateuszo that the points above this one are too detailed and we should keep them internally for our team. It could indeed make it much easier to leave Alokai.

filrak (changes requested):

docs/content/guides/6.multistore/2.tooling-and-concepts/4.deployment/8.ci-cd.md
jagoral: so maybe we should then make a CI workflow a blackbox, because right now all the code will be delivered to the customer anyway

grixu: We want to ensure that you understand that the client will still receive this code, so if necessary it can be migrated when they understand it. Unless it’s us who don’t understand the case you’re referring to, in which case that’s fine. In any event, we just want confirmation that we are all on the same page and understand the consequences both of removing this from the docs and that the client still receives this code.

filrak: I guess we can black-box it later on, for now this seems to be an easy fix for what @mateuszo raised. Let's keep it on confluence tho

jagoral: ok, I removed the CI part and moved it to the [Confluence](https://alokai.atlassian.net/wiki/spaces/TFMB/pages/1149304834/How+CI+pipeline+works)

filrak (approved):

PR #7406, opened 2025-02-25
Docs: Coding with AI best practices

mateuszo:


docs/content/guides/6.best-practices/3.ai-contexts.md
```diff
@@ -0,0 +1,69 @@
...
+//TODO: explain what are AI contexts
+
+## Expectations Management
+
+//TODO: elaborate on this points
+
+- AI is not a replacement for your knowledge and experience.
+- AI is a tool to help you code faster and more efficiently.
+- AI is not a replacement for your ability to reason and think.
+- AI is just predicts what would be the most likely outcome for the given input.
```
michaelKurowski, line 24: You could just remove this line and the meaning would be the same.
I feel like it's intuitive to know that AI responds to input. What may be non-obvious is context that is being attached and effective communication. The "empathy" element that I was talking about. Just like you'd need to explain some things to another dev, you have to transfer this knowledge to the AI

docs/content/guides/6.best-practices/3.ai-contexts.md
```diff
@@ -0,0 +1,69 @@
...
+There's also some randomness added to the process, so the results might be inconcistent.
+- AI can make silly mistakes. But it can do amazing things at the same time. Don't get disappointed too fast.
+- Don’t extrapolate your vision of intelligence on AI. AI is a different form of intelligence than human intelligence.
```
michaelKurowski, line 27: AI doesn't work the same way like human intelligence does. Some things that are easy to humans, are hard for AI, while some things that are hard for humans, are easy for AI.

I think we could also give examples of AI failing to do arithmetics or count "r" in strawberry.

docs/content/guides/6.best-practices/3.ai-contexts.md
```diff
@@ -0,0 +1,69 @@
...
+- AI cannot see the results of the code yet.
+
+## Recommended tools
+
+### IDEs
+
+- [Cursor](https://www.cursor.com/)
+- [Windsurf](https://windsurf.dev/)
+
+GitHub Copilot is not recommended. Agent mode is coming soon but we have not tested it yet.
```
michaelKurowski, line 39: GitHub Copilot is not recommended, but we do provide a context for it attached ootb.

docs/content/guides/6.best-practices/3.ai-contexts.md
```diff
@@ -0,0 +1,69 @@
...
+
+Alternatives to consider:
+
+- [Cline](https://cline.bot/)
+
+### LLM models
+
+- Claude
```
michaelKurowski, line 47: I mean, we don't support specific LLM model, it's the job of IDE to provide models. Just because we most extensively tested claude, doesn't mean we target claude exclusively. I'm not even sure you can switch context depending on the model. 
I'd say we recommend claude and that's about it, but this may get old fast if IDEs start to switch models automatically depending on the task

docs/content/guides/6.best-practices/3.ai-contexts.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Workflows
```
michaelKurowski, line 49: Styles of work? Workflow feels... a bit automated? Like CI/CD

docs/content/guides/6.best-practices/3.ai-contexts.md
```diff
@@ -0,0 +1,69 @@
...
+
+## General recommendations
+
+//TODO: elaborate on this points
+
+- review the code to know what's going on
+- keep files small
+- edit the context when needed
+- when something goes wrong retry or revert
+- ask AI to check the codebase for examples
```
michaelKurowski, line 64: when the AI goes against conventions

docs/content/guides/6.best-practices/3.ai-contexts.md
```diff
@@ -0,0 +1,69 @@
...
+AI field is evolving rapidly. New models, new tools, new workflows are emerging every day.
+While we do our best to keep this guide updated, we might miss some things.
+Revisit this guide regularly and don't follow it blindly.
+
+## AI Contexts
+
+Alokai provides a set of AI contexts that can be used to code with AI.
+//TODO: explain what are AI contexts
+
+## Expectations Management
```
michaelKurowski, line 17: Maybe I'd add that AI is made to work with it iteratively. I know people who think that it's one-off tool. You give it a task and if it failed, then they think AI is incapable of performing the task.

docs/content/guides/6.best-practices/3.ai-contexts.md
michaelKurowski: Also, we should explain why agent mode is a big deal

PR #7408, opened 2025-03-05
feat: add list command

rohrig:
Document: Add alokai base to commands
Document: Add yarn alokai list command


docs/content/guides/7.multistore/2.tooling-and-concepts/3.deployment/8.cd.md
```diff
@@ -94,7 +94,7 @@ jobs:
     steps:
       - name: Deploy store
         run: |
-          yarn store deploy \
+          yarn alokai store deploy \
```
rohrig, line 97: The pipeline is the only part I'm not sure about. I assume it's using yarn so it would work the same as the dev env. Please confirm this.

